PR #46, opened 2022-01-10
Start mapping model ops

FlorentinD:
Namely model.store.

Co-authored-by: Adam Schill Collberg <adam.schill.collberg@protonmail.com>


gdsclient/tests/integration/conftest.py
```diff
@@ -17,6 +17,25 @@
...
+def pytest_addoption(parser):
+    parser.addoption(
+        "--include-enterprise", action="store_true", help="include enterprise tests"
+    )
+
+
+@pytest.fixture
+def skip_if_not_enterprise(request) -> None:
+    if not request.config.getoption("--include-enterprise"):
+        pytest.skip()
```
adamnsch, line 29: I wonder if we should name it `_skip_if_not_enterprise` to declare that it should be used directly. Or better yet, maybe we could define it inside of `enterprise()` below :)

TESTING.md
```diff
@@ -41,6 +41,8 @@ The command for running the integration tests is the following:
 pytest gdsclient/tests/integration
 ```
 
+To include tests that require the Enterprise Edition of the Neo4j Graph Data Science library, you can specify the option `--include-enterprise`.
```
Mats-SX, line 44:
```suggestion
To include tests that require the Enterprise Edition of the Neo4j Graph Data Science library, you can specify the option `--include-enterprise`.
Naturally, this requires access to a valid Neo4j GDS license key, which can be acquired via https://neo4j.com/product/graph-data-science/
```

Mats-SX (approved):
LGTM -- just a comment on enabling the GDS EE

PR #54, opened 2022-01-13
Improve error message for incorrect method names

adamnsch:


FlorentinD (approved):
Looks good to me to have some earlier feedback.

In the future we might be able to auto-generate the procedure call structure from gds.list :thinking: 

An alternative solution would be to retrieve `gds.list()` once and index them. Then on `__call__` check if the procedure/function exists

PR #71, opened 2022-02-14
Increase pipeline train test graph size

adamnsch:
Assure that there are at least one relationship in each validation
fold.


graphdatascience/tests/integration/test_training_pipeline.py
```diff
@@ -19,12 +19,28 @@ def G(runner: Neo4jQueryRunner, gds: GraphDataScience) -> Generator[Graph, None,
...
+        (b)-[:REL]->(e),
+        (b)-[:REL]->(f),
+        (b)-[:REL]->(g),
         (c)-[:REL]->(a),
-        (c)-[:REL]->(b)
+        (c)-[:REL]->(b),
+        (c)-[:REL]->(d),
+        (c)-[:REL]->(e),
+        (c)-[:REL]->(f),
+        (c)-[:REL]->(g)
```
FlorentinD, line 43: NIT: you could have also changed the splitConfig

FlorentinD (approved):

graphdatascience/tests/integration/test_training_pipeline.py
adamnsch: I tried experimenting with that, but couldn't get it to work. Always some of the datasets not big enough. This solution seems better to me

PR #73, opened 2022-02-22
Methods returning graphs also return metadata

adamnsch:
Based on #72


FlorentinD (approved):
LGTM more info to the user :)

PR #74, opened 2022-02-22
Set explicit user agent in default constructor

adamnsch:


graphdatascience/graph_data_science.py
```diff
@@ -18,20 +18,21 @@ def __init__(
         self, endpoint: Union[str, QueryRunner], auth: Any = None, aura_ds: bool = False
     ):
         if isinstance(endpoint, str):
-            self._config = {}
+            self._config: Dict[str, Any] = {"user_agent": "neo4j-gds-client"}
```
Mats-SX, line 21: I think we should use the official name in this place, which would be like this
```suggestion
            self._config: Dict[str, Any] = {"user_agent": "neo4j-graphdatascience"}
```

I also wonder whether it would be cool to also include a version part, but not sure how to feed it here from the `setup.py`. Then the `AuraDS` data would include information about how up to date our users are.

graphdatascience/tests/integration/test_database_ops.py
```diff
@@ -34,6 +34,7 @@ def test_from_neo4j_driver(neo4j_driver: Driver) -> None:
 def test_from_neo4j_credentials() -> None:
     gds = GraphDataScience(URI, auth=AUTH)
     assert len(gds.list()) > 10
+    assert gds.driver_config()["user_agent"] == "neo4j-gds-client"
```
Mats-SX, line 37: if you agree, then let's change also here

graphdatascience/tests/integration/test_database_ops.py
```diff
@@ -66,6 +67,7 @@ def test_aurads_accepts_neo4j_s() -> None:
     assert gds.driver_config()["keep_alive"]
     assert gds.driver_config()["max_connection_lifetime"] == 60 * 8
     assert gds.driver_config()["max_connection_pool_size"] == 50
+    assert gds.driver_config()["user_agent"] == "neo4j-gds-client"
```
Mats-SX, line 70: and here

Mats-SX (changes requested):
I think we should use the official name of the client library. And if there's a neat way to get the version in there as well, like `neo4j-graphdatascience-v0.1.0` or something

Otherwise looks clean and nice.

graphdatascience/graph_data_science.py
adamnsch: Done

Mats-SX (approved):

Mats-SX:
Oh we should changelog this!

PR #75, opened 2022-02-23
Pipeline methods return procedures' metadata

adamnsch:
Based on #73


FlorentinD (approved):
LGTM details to the user :rocket:

PR #77, opened 2022-03-02
Add support for the GDS pipeline catalog

adamnsch:


FlorentinD:
this PR failes as it depends on changes in https://github.com/neo-technology/graph-analytics/pull/4796

README.md
```diff
@@ -212,6 +212,23 @@ Further, there's a new call named `gds.graph.get` (`graphdatascience` only) whic
...
+
+```python
+res = gds.beta.pipeline.list()
+assert len(res) == 1  # Exactly one pipeline is in the catalog
+
+res = gds.beta.pipeline.drop(pipe)
+assert res["pipelineName"] == pipe.name()
+```
+
+Further, there's a new call named `gds.pipeline.get` (`graphdatascience` only) which takes a pipeline name as input and returns a training pipeline object if a pipeline of that name exists in the user's pipeline catalog.
```
jjaderberg, line 228: This is the readme, which doesn't have a contextual dimension of time. Maybe we should avoid saying "new" here?

jjaderberg, line 228: Also suggest breaking this sentence into two, for readability

```suggestion
Further, there's a call named `gds.pipeline.get` (`graphdatascience` only).
It takes a pipeline name as input and returns a training pipeline object, if a pipeline of that name exists in the user's pipeline catalog.
```

graphdatascience/model/model.py
```diff
@@ -55,3 +56,24 @@ def drop(self) -> None:
...
+    def _query_prefix(self) -> str:
+        pass
+
+    def metrics(self) -> Dict[str, Any]:
+        return self._list_info()["modelInfo"]["metrics"]  # type: ignore
+
+    def predict_stream(self, G: Graph, **config: Any) -> QueryResult:
+        query = f"{self._query_prefix()}stream($graph_name, $config)"
+        config["modelName"] = self.name()
+        params = {"graph_name": G.name(), "config": config}
```
jjaderberg, line 70: it's a bit unfortunate to have to mix snake case and camel case conventions. is the user required to do that also, or are we making it possible for them to stay "pythonic" all the way?

jjaderberg (approved):
Looks good!
Except for the failing tests :-)

README.md
adamnsch: Good point :)

adamnsch: Done

PR #82, opened 2022-03-08
ANN is gone

lassewesth:


adamnsch:
Looks great :) Just need to run `script/makestyle` before committing and CI will be happy (https://github.com/neo4j/graph-data-science-client/blob/main/TESTING.md#style-guide for more on this)

adamnsch (approved):

lassewesth:
Sooo... do we think `$ black .` does the same? #feelinghipster 🤦

PR #83, opened 2022-03-16
Remove graph create endpoints

soerenreichardt:

PR #84, opened 2022-03-16
Implement add trainer procs for ml pipelines

soerenreichardt:
Based on #83

### DIFF
--- a/graphdatascience/pipeline/lp_training_pipeline.py
+++ b/graphdatascience/pipeline/lp_training_pipeline.py
@@ -21,10 +21,7 @@ class LPTrainingPipeline(TrainingPipeline):
 
     def addRandomForest(self, **config: Any) -> Row:
         query = f"{self._query_prefix()}addRandomForest($pipeline_name, $config)"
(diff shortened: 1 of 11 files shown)


graphdatascience/pipeline/lp_training_pipeline.py
```diff
@@ -19,6 +19,11 @@ def addFeature(self, feature_type: str, **config: Any) -> Row:
     def feature_steps(self) -> List[Dict[str, Any]]:
         return self._list_info()["pipelineInfo"]["featurePipeline"]["featureSteps"]  # type: ignore
 
+    def addRandomForest(self, **config: Any) -> Row:
```
adamnsch, line 22: This one should also be on `training_pipeline.py` now

adamnsch (changes requested):
Very nice :) I think we also need to:

- remove `configureParams`
- update README.md
- add integration tests for `addRandomForest` and `addLogisticRegression`

adamnsch (approved):
Great job we!!

PR #87, opened 2022-03-22
Close internally instantiated Neo4j driver when all references to it are dropped

adamnsch:


adamnsch:
I found no way of testing this as there seem to be no way to gauge if a driver has been closed or not. I even tried to look at the internal fields of the driver for such info, but without success. I also don't get any warnings if I don't close the driver locally :man_shrugging: But I think we can live without a test for this now.

graphdatascience/graph_data_science.py
```diff
@@ -32,7 +32,7 @@ def __init__(self, endpoint: Union[str, QueryRunner], auth: Any = None, aura_ds:
 
             driver = GraphDatabase.driver(endpoint, auth=auth, **self._config)
 
-            self._query_runner = self.create_neo4j_query_runner(driver)
+            self._query_runner = self.create_neo4j_query_runner(driver, auto_close=True)
```
s1ck, line 35:
```suggestion
            self._query_runner = self.create_neo4j_query_runner(driver, auto_close = True)
```

graphdatascience/graph_data_science.py
```diff
@@ -55,5 +55,5 @@ def from_neo4j_driver(cls: Type[GDS], driver: Driver) -> "GraphDataScience":
         return cls(cls.create_neo4j_query_runner(driver))
 
     @staticmethod
-    def create_neo4j_query_runner(driver: Driver) -> Neo4jQueryRunner:
-        return Neo4jQueryRunner(driver)
+    def create_neo4j_query_runner(driver: Driver, auto_close: bool = False) -> Neo4jQueryRunner:
+        return Neo4jQueryRunner(driver, auto_close=auto_close)
```
s1ck, line 59:
```suggestion
        return Neo4jQueryRunner(driver, auto_close = auto_close)
```

s1ck: not sure what your overall code style is wrt default parameters, but it should be just one :)

graphdatascience/query_runner/neo4j_query_runner.py
```diff
@@ -17,3 +18,7 @@ def run_query(self, query: str, params: Dict[str, Any] = {}) -> QueryResult:
 
     def set_database(self, db: str) -> None:
         self._db = db
+
+    def __del__(self) -> None:
+        if self._auto_close:
+            self._driver.close()
```
s1ck, line 24: According to the docs, closing the driver also closes all open connections/sessions. I assume the only way of checking if it's safe to close the driver is to know if there are still open sessions. 

There seems to be a field on the `Driver` https://github.com/neo4j/neo4j-python-driver/blob/3389c4ad0f51583ebd6ceaf903e7555f45d9a609/neo4j/_sync/driver.py#L296-L300 which we could use in a test? OTOH, I don't see any tests in the driver impl for testing close functionality, so I guess this is fine.

s1ck (approved):
Looks good to me. I skimmed over the driver impl and could find https://github.com/neo4j/neo4j-python-driver/blob/3389c4ad0f51583ebd6ceaf903e7555f45d9a609/neo4j/_sync/driver.py#L296-L300

but maybe this is not the right code to look at? I don't think we need to test this behaviour in particular.

graphdatascience/graph_data_science.py
adamnsch: These are two different things: _defining_ keyword parameters, and _using_ keyword parameters. For each there are very strict rules which are enforced by our most opinionated formatter https://black.readthedocs.io/en/stable/. There's no conflict of styles here.

graphdatascience/query_runner/neo4j_query_runner.py
adamnsch: This is very interesting. I'm using the latest released version of the driver (4.4.2) and there is no `Driver._close` field in my installation (see https://github.com/neo4j/neo4j-python-driver/blob/a0263b05048dbb379f8f4989842d0785cbde7c86/neo4j/__init__.py#L268).
I guess we can test this more easily in the 5.0 driver. In fact, then we can look at whether we get warnings or not (https://github.com/neo4j/neo4j-python-driver/blob/3389c4ad0f51583ebd6ceaf903e7555f45d9a609/neo4j/_sync/driver.py#L263) and we'd not have to rely on private internal fields in the driver.

adamnsch:
> Looks good to me. I skimmed over the driver impl and could find https://github.com/neo4j/neo4j-python-driver/blob/3389c4ad0f51583ebd6ceaf903e7555f45d9a609/neo4j/_sync/driver.py#L296-L300
> 
> but maybe this is not the right code to look at? I don't think we need to test this behaviour in particular.

Thanks for the review :heart: It's the right code to look at, just on the wrong branch. But it will indeed be easy to test this in the future :)

graphdatascience/graph_data_science.py
s1ck: Thanks for the explanation. My Python is Rust-y *hue hue*

PR #92, opened 2022-04-22
Add infrastructure for multi GDS library version support

adamnsch:
In summary:
* Cache which server version we're running against in all classes that makes calls to the query runner
* Use the `multimethod` library for method overloading
* Add decorator that declares which GDS library versions a client function is compatible with
* Use new pytest marker for skipping integration tests intended for other GDS library versions
* Use pytest `parameterize` to specify "mock" GDS library version for tests that require it
* Add back support for GDS library 2.0 `gds.graph.removeNodeProperties` signature with more testing


graphdatascience/tests/integration/conftest.py
```diff
@@ -34,3 +34,36 @@ def runner(neo4j_driver: Driver) -> Neo4jQueryRunner:
...
+        skip_enterprise = pytest.mark.skip(reason="need --include-enterprise option to run")
+        for item in items:
+            if "enterprise" in item.keywords:
+                item.add_marker(skip_enterprise)
+
+    if not config.getoption("--include-model-store-location"):
+        skip_stored_models = pytest.mark.skip(reason="need --include-model-store-location option to run")
+        for item in items:
+            if "model_store_location" in item.keywords:
+                item.add_marker(skip_stored_models)
```
adamnsch, line 50: This part was just moved here from `graphdatascience/tests/conftest.py`

adamnsch:
I think I need to update the docker image for the testing to go through due to the new library

graphdatascience/tests/unit/test_error_handling.py
```diff
@@ -120,3 +128,18 @@ def test_nonexisting_util_endpoint(gds: GraphDataScience) -> None:
 def test_wrong_client_only_prefix(gds: GraphDataScience) -> None:
     with pytest.raises(SyntaxError, match="There is no 'gds.beta.model.get' to call"):
         gds.beta.model.get("model")
+
+
+def test_incompatible_server_version(runner: QueryRunner) -> None:
+    G = Graph("dummy", runner, ServerVersion(0, 0, 0))  # This version object is not relevant for the test
+    gds = GraphDataScience(runner)
+    gds._server_version = ServerVersion(2, 2, 0)  # Something later than 2.1.0
```
breakanalysis, line 136: need to use pytest.mark.parametrize to set the version

graphdatascience/graph_data_science.py
```diff
@@ -37,10 +42,18 @@ def __init__(self, endpoint: Union[str, QueryRunner], auth: Any = None, aura_ds:
         else:
             self._query_runner = endpoint
 
-        super().__init__(self._query_runner, "gds")
+        try:
+            server_version_string = self._query_runner.run_query("RETURN gds.version()").squeeze()
+        except Exception as e:
+            raise UnableToConnectError(e)
+
+        server_version_numbers = map(int, server_version_string[:5].split("."))
```
breakanalysis, line 50:
```suggestion
        foo = server_version_string.split(".")[:3]
        server_version_numbers = map(lambda s: int(re.sub("-alpha[\d]+", "", s)), foo)
```

also import re, and check that it works

graphdatascience/tests/integration/test_database_ops.py
```diff
@@ -63,14 +63,19 @@ def test_aurads_rejects_neo4j_ssc() -> None:
...
 
 
 def test_run_cypher(gds: GraphDataScience) -> None:
     result = gds.run_cypher("CALL gds.list()")
     assert len(result) > 10
+
+
+def test_server_version(gds: GraphDataScience) -> None:
+    cached_server_version = gds._server_version
+    server_version = gds.version()[:5].split(".")
```
breakanalysis, line 77: regex also here

requirements/base.txt
```diff
@@ -1,2 +1,3 @@
-neo4j>=4.4.2, <5.0
-pandas>=1.0, <2.0
+multimethod >= 1.0, < 2.0
```
breakanalysis, line 1: kinda wonder if its not safer and sufficient to require a specific version. or would that risk versions being deprecated and then fail to be fetched ?

graphdatascience/graph/graph_export_runner.py
```diff
@@ -2,16 +2,12 @@
...
-from ..query_runner.query_runner import QueryRunner
 from .graph_object import Graph
 
 
-class GraphExportRunner(IllegalAttrChecker):
-    def __init__(self, query_runner: QueryRunner, namespace: str):
-        self._query_runner = query_runner
-        self._namespace = namespace
-
+class GraphExportRunner(CallerBase, IllegalAttrChecker):
```
breakanalysis, line 10: doesnt this need a server version too ?

graphdatascience/algo/algo_proc_runner.py
```diff
@@ -25,7 +21,7 @@ def _run_procedure(self, G: Graph, config: Dict[str, Any]) -> DataFrame:
         return self._query_runner.run_query(query, params)
 
     def estimate(self, G: Graph, **config: Any) -> Series:
-        self._proc_name += "." + "estimate"
+        self._namespace += "." + "estimate"
```
breakanalysis, line 24: how does one call estimate ? gds.nodeClassification.train().estimate(config) ?
im confused because I'd guess train should take the config and run immediately

.flake8
```diff
@@ -1,3 +1,3 @@
 [flake8]
 max-line-length = 120
-exclude = build, venv
+exclude = build, venv, .tox
```
breakanalysis, line 3: what is .tox ?

breakanalysis (changes requested):
Looks really solid.
Only need to fix the version regexing as we saw.

.flake8
adamnsch: It's a cache for the tool `tox`. tox something use to test our lib against a matrix of dependencies and Python versions

requirements/base.txt
adamnsch: Because of how people install things with pip, the convention is to specify an as generous interval as possible here. Several packages a user have installed in their Python environment may have the same `multimethod` dependency. If they all had a different version of `multimethod` fixed, only one package could be installed unless they were the same. One solution is of course to use something like virtual environments, but not all people do. And having one VE for each package installation is rather steep I think.

Since we're just within one major version here, it should at least in theory not be an issue for us. If we were to support multiple major version for some dependency, we'd have to make to to add every such major version with tox.

graphdatascience/graph/graph_export_runner.py
adamnsch: Yes, that's why iit inherits `CallerBase`

graphdatascience/algo/algo_proc_runner.py
adamnsch: For the pipelines you call `pipe.train_estimate(G, concurrency=4, metrics=[...],...)` to estimate

graphdatascience/tests/unit/test_error_handling.py
adamnsch: done

graphdatascience/tests/integration/test_database_ops.py
adamnsch: done

adamnsch:
> Looks really solid. Only need to fix the version regexing as we saw.

Thanks for the review! I think all comments have been addressed

breakanalysis (approved):
LGTM!

PR #93, opened 2022-04-28
Initial arrow integration for streaming properties

DarthMax:
- Initial support for gds flight server for streamNodeProperty
- Add test for running streamNodeProperties with and without arrow
- Improve initialization of ArrowQueryRunner

### DIFF
--- a/graphdatascience/graph_data_science.py
+++ b/graphdatascience/graph_data_science.py
@@ -11,6 +11,7 @@ from .direct_endpoints import DirectEndpoints
 from .error.uncallable_namespace import UncallableNamespace
 from .query_runner.arrow_query_runner import ArrowQueryRunner
 from .query_runner.neo4j_query_runner import Neo4jQueryRunner
+from .query_runner.query_runner import QueryRunner
 from .server_version.server_version import ServerVersion
 from .version import __version__
 
@@ -30,7 +31,7 @@ class GraphDataScience(DirectEndpoints, UncallableNamespace):
 
     def __init__(
         self,
-        endpoint: Union[str, Driver],
+        endpoint: Union[str, Driver, QueryRunner],
         auth: Optional[Tuple[str, str]] = None,
         aura_ds: bool = False,
         arrow: bool = True,
@@ -52,6 +53,12 @@ class GraphDataScience(DirectEndpoints, UncallableNamespace):
             driver = GraphDatabase.driver(endpoint, auth=auth, **self._config)
 
             self._query_runner = Neo4jQueryRunner(driver, auto_close=True)
+
+        elif isinstance(endpoint, QueryRunner):
+            if arrow:
+                raise ValueError("Arrow cannot be used if the QueryRunner is provided directly")
+
+            self._query_runner = endpoint
 
         else:
             driver = endpoint
--- a/graphdatascience/graph_data_science.py
+++ b/graphdatascience/graph_data_science.py
@@ -1,5 +1,6 @@
 import re
-from typing import Any, Dict, Type, TypeVar, Union
+import warnings
+from typing import Any, Dict, Optional, Tuple, Type, TypeVar, Union
 
 from neo4j import Driver, GraphDatabase
 from pandas import Series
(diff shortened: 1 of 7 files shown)


requirements/base.txt
```diff
@@ -1,3 +1,4 @@
 multimethod >= 1.0, < 2.0
 neo4j >= 4.4.2, < 5.0
 pandas >= 1.0, < 2.0
+pyarrow
```
adamnsch, line 4: We should specify an explicit interval here

graphdatascience/tests/integration/conftest.py
```diff
@@ -12,7 +12,7 @@
 AUTH = None
 if os.environ.get("NEO4J_USER") is not None:
     AUTH = (
-        os.environ.get("NEO4J_USER"),
+        os.environ.get("NEO4J_USER", "DUMMY"),
```
adamnsch, line 15: I think we can revert this change

graphdatascience/graph_data_science.py
```diff
@@ -57,6 +74,21 @@ def __init__(self, endpoint: Union[str, QueryRunner], auth: Any = None, aura_ds:
             raise InvalidServerVersionError(f"{server_version_string} is not a valid GDS library version")
         self._server_version = ServerVersion(*map(int, server_version_match.groups()))
 
+        if arrow and self._server_version >= ServerVersion(2, 1, 0):
+            try:
+                arrow_info: Series = self._query_runner.run_query("CALL gds.debug.arrow()").squeeze()
+                if arrow_info["running"]:
+                    disable_server_verification = driver._pool.pool_config.trust == "TRUST_ALL_CERTIFICATES"
```
adamnsch, line 81: As @robsdedude warned us on slack we  cannot really rely on these internals and need to find another way of extracting this info

graphdatascience/tests/integration/conftest.py
adamnsch: nevermind

graphdatascience/graph_data_science.py
adamnsch: We will always set this to true for now and hopefully there will be a way to poll in the driver API in the future

requirements/base.txt
adamnsch: done

knutwalker (approved):
Nice stuff, looking forward to the next one

graphdatascience/graph_data_science.py
```diff
@@ -72,9 +99,7 @@ def driver_config(self) -> Dict[str, Any]:
...
     @classmethod
-    def from_neo4j_driver(cls: Type[GDS], driver: Driver) -> "GraphDataScience":
-        return cls(cls.create_neo4j_query_runner(driver))
-
-    @staticmethod
-    def create_neo4j_query_runner(driver: Driver, auto_close: bool = False) -> Neo4jQueryRunner:
-        return Neo4jQueryRunner(driver, auto_close=auto_close)
+    def from_neo4j_driver(
+        cls: Type[GDS], driver: Driver, auth: Optional[Tuple[str, str]] = None, arrow: bool = True
+    ) -> "GraphDataScience":
```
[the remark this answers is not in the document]
adamnsch: We cannot reference `GraphDataScience` directly since it's not fully defined at this point, so we have to use a string instead - that's Python for you :man_shrugging:

requirements/base.txt
```diff
@@ -1,3 +1,4 @@
 multimethod >= 1.0, < 2.0
 neo4j >= 4.4.2, < 5.0
 pandas >= 1.0, < 2.0
+pyarrow >= 7.0, < 8.0
```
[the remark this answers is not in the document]
adamnsch: Thanks for the info :) I think the tests are failing since Python 3.6 is not compat with pyarrow 7, so we definitely need to lower it anyway. pyarrow 3 sounds good to me, what do you think @DarthMax ? We'll need to add pyarrow major version to our tox test matrix though

graphdatascience/graph_data_science.py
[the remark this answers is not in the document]
knutwalker: Redefining stringly-typed, I see 🙈

graphdatascience/query_runner/arrow_query_runner.py
```diff
@@ -0,0 +1,64 @@
...
+            if encrypted
+            else flight.Location.for_grpc_tcp(host, int(port_string))
+        )
+
+        self._flight_client = flight.FlightClient(location, disable_server_verification=disable_server_verification)
+        self._flight_options = flight.FlightCallOptions()
+
+        if auth:
+            username, password = auth
+            (header, token) = self._flight_client.authenticate_basic_token(username, password)
```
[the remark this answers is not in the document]
adamnsch: done

requirements/base.txt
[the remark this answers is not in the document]
adamnsch: Set to >= 3.0 now. Test matrix changes in https://github.com/neo4j/graph-data-science-client/pull/97

PR #106, opened 2022-05-17
Arrow export multiple properties

DarthMax:
- Add support for exporting node properties via arrow
- Add support for exporting relationship properties via arrow

### DIFF
--- a/graphdatascience/tests/integration/test_graph_ops.py
+++ b/graphdatascience/tests/integration/test_graph_ops.py
@@ -178,7 +178,7 @@ def test_graph_streamNodeProperties(gds: GraphDataScience) -> None:
 def test_graph_streamNodeProperties_merge_property_columns(gds: GraphDataScience) -> None:
     G, _ = gds.graph.project(GRAPH_NAME, {"Node": {"properties": ["x", "y"]}}, "*")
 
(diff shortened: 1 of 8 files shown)


graphdatascience/tests/integration/test_graph_ops.py
```diff
@@ -185,10 +217,43 @@ def test_graph_streamRelationshipProperty_without_arrow(gds_without_arrow: Graph
...
+    assert {e for e in x_values["propertyValue"]} == {4, 5, 6}
+    y_values = result[result.relationshipProperty == "relY"]
+    assert {e for e in y_values["propertyValue"]} == {5, 6, 7}
+
+
+def test_graph_streamRelationshipProperties_without_arrow(gds_without_arrow: GraphDataScience) -> None:
+    G, _ = gds_without_arrow.graph.project(GRAPH_NAME, "*", {"REL": {"properties": ["relX", "relY"]}})
+
+    result = gds_without_arrow.graph.streamRelationshipProperties(G, ["relX", "relY"], concurrency=2)
+    print(result.to_markdown())
```
adamnsch, line 243: Let's remove this print and the `tabulate` dependency?

adamnsch (changes requested):
Looks great, mainly concerned about breaking the API and the new dependency. I think we said we will change what format is default when we decide to release 2.0 of the client?

PR #107, opened 2022-05-18
Fix client authentication

DarthMax:


graphdatascience/query_runner/arrow_query_runner.py
```diff
@@ -28,16 +31,15 @@ def __init__(
...
-            username, password = auth
-            header, token = self._flight_client.authenticate_basic_token(username, password)
-            if header:
-                self._flight_options = flight.FlightCallOptions(headers=[(header, token)])
+            client_options["middleware"] = [AuthFactory(auth)]
+
+        self._flight_client = flight.FlightClient(location, **client_options)
 
-    def run_query(self, query: str, params: Dict[str, Any] = {}) -> DataFrame:
+    def run_query(self, query: str, params: Optional[Dict[str, Any]] = None) -> DataFrame:
```
adamnsch, line 40: I don't understand why we changed this interface. Now we have these if statements everywhere. Would it be possible to change that back?

graphdatascience/query_runner/arrow_query_runner.py
```diff
@@ -79,9 +83,61 @@ def _run_arrow_property_get(self, graph_name: str, procedure_name: str, configur
...
+    def sending_headers(self) -> Dict[str, str]:
+        token = self._factory.token()
+        if token is None:
+            username, password = self._factory.auth
+            auth_token = "{0}:{1}".format(username, password)
+            auth_token = "Basic " + base64.b64encode(auth_token.encode("utf-8")).decode("ASCII")
+            # There seems to be a bug, `authorization` must be lower key
+            return {"authorization": auth_token}
+        else:
+            # # There seems to be a bug, `authorization` must be lower key
```
adamnsch, line 142: (nitpick) double #

knutwalker:
[[Arrow] Fix session expiration in client](https://trello.com/c/sn0xw3yu/6211-arrow-fix-session-expiration-in-client)

graphdatascience/query_runner/arrow_query_runner.py
```diff
@@ -79,9 +83,61 @@ def _run_arrow_property_get(self, graph_name: str, procedure_name: str, configur
...
+    def __init__(self, factory: AuthFactory, *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+        self._factory = factory
+
+    def received_headers(self, headers: Dict[str, Any]) -> None:
+        auth_header: str = headers.get("Authorization", None)
+        if auth_header is not None:
+            [auth_type, token] = auth_header.split(" ", 1)
+            if auth_type == "Bearer":
+                self._factory.set_token(token)
```
adamnsch, line 131:
```suggestion
        if not auth_header:
            return
        [auth_type, token] = auth_header.split(" ", 1)
        if auth_type == "Bearer":
            self._factory.set_token(token)
```
(nitpick) prefer this style to avoid a lot of indentation

graphdatascience/query_runner/arrow_query_runner.py
```diff
@@ -79,9 +83,61 @@ def _run_arrow_property_get(self, graph_name: str, procedure_name: str, configur
...
+
+    def sending_headers(self) -> Dict[str, str]:
+        token = self._factory.token()
+        if token is None:
```
adamnsch, line 135: (nitpick) More pythonic to just write `if not token:`

graphdatascience/query_runner/arrow_query_runner.py
```diff
@@ -79,9 +83,61 @@ def _run_arrow_property_get(self, graph_name: str, procedure_name: str, configur
...
+            username, password = self._factory.auth
+            auth_token = "{0}:{1}".format(username, password)
```
adamnsch, line 137: (nitpick) Since Python 3.6 people typically use f-strings for this kind of thing (we also do that everywhere else in the repo - maybe I should add a style check for it)

adamnsch (changes requested):
Looks good, just a bit confused about the change to the query runner interface

graphdatascience/query_runner/arrow_query_runner.py
adamnsch: https://realpython.com/python-f-strings/#f-strings-a-new-and-improved-way-to-format-strings-in-python

graphdatascience/query_runner/arrow_query_runner.py
DarthMax: @knutwalker (as well as Intellij) remarked on this a a potential bug. The default dict that is created is a shared instance through the lifetime of the class instance/interpreter (?). If it is changed at some point these changes would be seen by any invocation of the method.

adamnsch: Ah, good catch!

adamnsch (approved):
LGTM :) Just some minor nitpicks that I'd personally fix before merge, but I'll leave that for you to decide

PR #112, opened 2022-05-27
Add more graph convenience methods

adamnsch:
Based on #111

### DIFF
--- a/graphdatascience/graph/graph_object.py
+++ b/graphdatascience/graph/graph_object.py
@@ -1,4 +1,4 @@
-from typing import List
+from typing import Any, List
 
 import pandas
 from pandas.core.series import Series
@@ -86,3 +86,6 @@ class Graph:
         )
 
         return result.squeeze()  # type: ignore
+
+    def creation_time(self) -> Any:  # neo4j.time.DateTime not exported
+        return self._graph_info()["creationTime"]
--- a/graphdatascience/graph/graph_object.py
+++ b/graphdatascience/graph/graph_object.py
@@ -73,8 +73,8 @@ class Graph:
         )
         return result.squeeze()["exists"]  # type: ignore
 
-    def drop(self) -> None:
+    def drop(self, failIfMissing: bool = False) -> None:
         self._query_runner.run_query(
-            "CALL gds.graph.drop($graph_name, false)",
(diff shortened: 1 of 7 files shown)


graphdatascience/graph/graph_object.py
```diff
@@ -28,6 +28,12 @@ def _graph_info(self, yields: List[str] = []) -> Series:
 
         return info.squeeze()  # type: ignore
 
+    def database(self) -> str:
+        return self._graph_info(["database"])  # type: ignore
```
Mats-SX, line 32: it would be really neat if we had the feature to control `YIELD` given this kind of API. At the moment we `YIELD *` and then pick on the client side what to actually use, right? Which means that we end up computing the expensive `degreeDistribution` for things like this. (Fortunately it is cached, so we only pay once, but still.)

graphdatascience/tests/integration/test_graph_object.py
```diff
@@ -37,6 +37,14 @@ def G(gds: GraphDataScience) -> Generator[Graph, None, None]:
     gds.graph.drop(G, False)
 
 
+def test_graph_database(gds: GraphDataScience, G: Graph) -> None:
+    assert G.database() == gds.database()
+
+
+def test_graph_configuration(G: Graph) -> None:
+    assert "Node" in G.configuration()["nodeProjection"].keys()
```
Mats-SX, line 45: very superficial test, but I think this is enough. we have plenty of tests on the server side

graphdatascience/tests/integration/test_graph_object.py
```diff
@@ -100,3 +108,11 @@ def test_graph_drop(G: Graph) -> None:
 
     with pytest.raises(Exception):
         G.drop(failIfMissing=True)
+
+
+def test_graph_creation_time(G: Graph) -> None:
+    assert G.creation_time().year > 2000
```
Mats-SX, line 114: it's the y2k bug all over

Mats-SX (approved):

graphdatascience/graph/graph_object.py
Mats-SX: Oh we have that feature already? That's great.

PR #116, opened 2022-06-01
Improve performance of `Graph.{creation,modification}_time`

adamnsch:


adamnsch:
@Mats-SX Now we're actually utilizing the yield optimization :see_no_evil:

graphdatascience/graph/graph_object.py
```diff
@@ -88,7 +88,7 @@ def drop(self, failIfMissing: bool = False) -> Series:
         return result.squeeze()  # type: ignore
 
     def creation_time(self) -> Any:  # neo4j.time.DateTime not exported
-        return self._graph_info()["creationTime"]
+        return self._graph_info(["creationTime"])
 
     def modification_time(self) -> Any:  # neo4j.time.DateTime not exported
-        return self._graph_info()["modificationTime"]
+        return self._graph_info(["modificationTime"])
```
Mats-SX, line 94: Subtle!!

Mats-SX (approved):

PR #120, opened 2022-06-03
Make `run_cypher` argument `params` default to `None`

adamnsch:
Since the previous default, `{}`, is only initiated at method definition
and not for every call of the method. This might prevent future bugs, in
the case where the query runner populates the default params with
something every call.


knutwalker (approved):
good catch 👍

PR #123, opened 2022-06-23
Fix pipeline parameter space test

FlorentinD:


FlorentinD:
TODO: for master the size is 3 and for 2.0 its 2

PR #133, opened 2022-07-20
Drop support for Python 3.6

adamnsch:
Since it has been end-of-life for a while now and no longer receiving
security updates.


FlorentinD (approved):
:snake: :wave:

PR #136, opened 2022-07-22
Fix integration tests for filtered LP in GDS 2.2.0

brs96:
Co-authored-by: Florentin Dörre <florentin.dorre@neotechnology.com>


knutwalker (approved):
👍

PR #138, opened 2022-07-25
Stream topology endpoint

soerenreichardt:


soerenreichardt:
[[Topology] Add stream topology arrow endpoint](https://trello.com/c/ILci9J7q/6417-topology-add-stream-topology-arrow-endpoint)

knutwalker (changes requested):
Overall, looks good.
I have one remark about the default type, Github says there is a conflict, and TC says

```
graphdatascience/query_runner/arrow_query_runner.py:123: error: Argument 3 to "_run_arrow_property_get" of "ArrowQueryRunner" has incompatible type "Set[Any]"; expected "Dict[str, Any]"
```

graphdatascience/graph/graph_entity_ops_runner.py
```diff
@@ -150,3 +150,11 @@ def drop(
         }
 
         return self._query_runner.run_query(query, params).squeeze()  # type: ignore
+
+    @compatible_with("stream", min_inclusive=ServerVersion(2, 2, 0))
+    def stream(self, G: Graph, relationship_types: Strings = ["*"], **config: Any) -> DataFrame:
```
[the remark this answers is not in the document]
soerenreichardt: Good point, didn't know that. I think that pattern is used across the codebase. Maybe we should create a card to fix all locations

knutwalker (approved):
Approving, waiting for whatever CI complains about :)

graphdatascience/tests/integration/test_graph_ops.py
```diff
@@ -532,6 +532,28 @@ def test_graph_relationshipProperties_stream_without_arrow_separate_property_col
...
+@pytest.mark.compatible_with(min_inclusive=ServerVersion(2, 2, 0))
+def test_graph_relationships_stream_without_arrow(gds_without_arrow: GraphDataScience) -> None:
+    G, _ = gds_without_arrow.graph.project(GRAPH_NAME, "*", "REL")
+
+    result = gds_without_arrow.beta.graph.relationships.stream(G, ["REL"])
+
+    assert list(result.keys()) == ["sourceNodeId", "targetNodeId", "relationshipType"]
+    assert {e for e in result["sourceNodeId"]} == {0, 1}
+    assert {e for e in result["targetNodeId"]} == {1, 2}
+
```
FlorentinD, line 544: we did not depend on the neo4j ids before. As we are testing multiple version combinations on the same neo4j db, these fail here. 

@soerenreichardt Instead we should get the neo-ids via a Cypher query and uses these for the expected ids

FlorentinD: fixing a bug https://github.com/neo-technology/graph-analytics/pull/5952

FlorentinD: waiting for fix on server-side to be merged

soerenreichardt: Yes that should be the reason for the flaky tests here

PR #139, opened 2022-08-03
Expose tls_root_certs in Client constructor

s1ck:
TODOs:
* [x]  update integration test documentation


graphdatascience/tests/conftest.py
```diff
@@ -3,6 +3,9 @@
 
 def pytest_addoption(parser: Any) -> None:
     parser.addoption("--include-enterprise", action="store_true", help="include tests requring GDS enterprise")
+    parser.addoption(
```
zen-dog, line 6: nit about the flag name: at least in Aura, all *DS databases uses encryption so, it's not really an "enterprise" feature.

graphdatascience/graph_data_science.py
```diff
@@ -32,7 +32,31 @@ def __init__(
...
+            A flag that indicates that that the client is used to connect
+            to a Neo4j Aura instance.
+        arrow : bool, default True
+            A flag that indicates that the client should use Apache Arrow
+            for data streaming if it is available on the server.
+        arrow_disable_server_verification : bool, default True
+            A flag that indicates that, if the flight client is connecting with
+            TLS, that it skips server verification. If this is enabled, all
+            other TLS settings are overridden.
+        arrow_tls_root_certs : Optional[bytes], default None
```
zen-dog, line 55: A better UX would be falling back to the OS root CA bundle (I used python certifi to find it in my tests). But that's up to you, of course.

zen-dog (commented):
A few nits only. I really think that falling back to OS root CA makes sense for all Aura use cases.

graphdatascience/tests/conftest.py
s1ck: makes sense, we can call it `--encrypted`

graphdatascience/graph_data_science.py
s1ck: Arrow / grpc already has a mechanism for setting the default value: https://grpc.github.io/grpc/cpp/structgrpc_1_1_ssl_credentials_options.html#a0ce1730020e18d04b6af48c88e069869

If we always set the value in the GDS client, changing that env var won't have an effect, i.e. users that know about that will be surprised. Maybe it's also an option for Aura to not handle the default in the client but setting the env variable instead?

s1ck:
@zen-dog Thanks for the review! We changed the config param, but left the `tls_root_certs` behavior for now. Maybe setting the existing env variable for grpc is sufficient for Aura.

graphdatascience/graph_data_science.py
zen-dog: I mean, wouldn't a typical user be overwhelmed with finding where the root CA bundle lives on their machine? And neo4j driver picks it up automatically for bolt encrypted connection. But at the end, these are your users so 🤷

graphdatascience/graph_data_science.py
```diff
@@ -32,7 +32,31 @@ def __init__(
...
+        aura_ds : bool, default False
+            A flag that indicates that that the client is used to connect
+            to a Neo4j Aura instance.
+        arrow : bool, default True
+            A flag that indicates that the client should use Apache Arrow
+            for data streaming if it is available on the server.
+        arrow_disable_server_verification : bool, default True
+            A flag that indicates that, if the flight client is connecting with
+            TLS, that it skips server verification. If this is enabled, all
+            other TLS settings are overridden.
```
Mats-SX, line 54: Is this the feature? I'm not sure what "server verification" means exactly, but it does sound odd to me to give users the ability to bypass something that sounds like security.

graphdatascience/tests/conftest.py
Mats-SX: this is just a test flag right?

graphdatascience/tests/integration/test_graph_ops.py
```diff
@@ -389,6 +389,23 @@ def test_roundtrip_with_arrow_22(gds: GraphDataScience) -> None:
         G_2.drop()
 
 
+@pytest.mark.encrypted
+@pytest.mark.compatible_with(min_inclusive=ServerVersion(2, 1, 0))
+def test_roundtrip_with_arrow_encrypted(gds_with_tls: GraphDataScience) -> None:
+    G, _ = gds_with_tls.graph.project(GRAPH_NAME, {"Node": {"properties": ["x", "y"]}}, {"REL": {"properties": "relX"}})
+
+    rel_df = gds_with_tls.graph.streamRelationshipProperty(G, "relX")
+    node_df = gds_with_tls.graph.streamNodeProperty(G, "x")
```
Mats-SX, line 398: Shouldn't we use the new, renamed endpoints here? Or maybe if we do then it won't work on 2.1 so probably not

Mats-SX (approved):
Looks good to me!

graphdatascience/graph_data_science.py
s1ck: It's just exposing a flag of the arrow flight client. It basically means that you can connect with tls, but the server is only accepting unencrypted, it'll still connect. It's not disabling the server-side encryption.

TESTING.md
```diff
@@ -52,6 +52,30 @@ To include tests that exercise storing and loading models, you must specify the
...
+# Allow bolt connections either encrypted or unencrypted
+dbms.connector.bolt.tls_level=OPTIONAL
+dbms.ssl.policy.bolt.enabled=true
+dbms.ssl.policy.bolt.base_directory=<absolute-path-to-graph-datascience-client>/graphdatascience/tests/integration/resources
+dbms.ssl.policy.bolt.public_certificate=arrow-flight-gds-test.crt
+dbms.ssl.policy.bolt.private_key=arrow-flight-gds-test.key
+dbms.ssl.policy.bolt.client_auth=NONE
+```
+
+To run only integration tests that are marked as `enterprise_ssl`, call:
```
adamnsch, line 72: Why do we want this "only" behavior?

adamnsch: Also it's called `encrypted` right? Not `enterprise_ssl`?

graphdatascience/graph_data_science.py
```diff
@@ -32,7 +32,31 @@ def __init__(
         auth: Optional[Tuple[str, str]] = None,
         aura_ds: bool = False,
         arrow: bool = True,
+        arrow_disable_server_verification: bool = True,
```
adamnsch, line 35: This constructor is starting to get out of hand. The idea of using the same constructor for arrow was to hide it from the user, but that's no longer happening anyway. I could imagine having something like a static `GraphDataScience.with_tls` constructor. But I don't know

TESTING.md
adamnsch: Also I don't see any testing of this in CI. Should we not have that?

TESTING.md
```diff
@@ -52,6 +52,30 @@ To include tests that exercise storing and loading models, you must specify the
...
+
+```bash
+pytest graphdatascience/tests/integration --include-enterprise
```
adamnsch, line 75: `--encrypted`?

TESTING.md
```diff
@@ -52,6 +52,30 @@ To include tests that exercise storing and loading models, you must specify the
 Note however that this also requires you to have specified a valid path for the `gds.model.store_location` configuration key of your database.
 
 
+### Running tests that require encrypted connections
```
adamnsch, line 55: Nice docs :)

graphdatascience/tests/conftest.py
```diff
@@ -3,6 +3,9 @@
...
+        "--encrypted", action="store_true", help="runs only tests requiring an encrypted connection to Neo4j"
```
adamnsch, line 7: If we really want the "only" behavior I think `--only-encrypted` would be clearer

graphdatascience/tests/integration/conftest.py
```diff
@@ -44,6 +46,22 @@ def gds() -> GraphDataScience:
     return _gds
 
 
+@pytest.fixture(scope="package")
+def gds_with_tls(request: FixtureRequest) -> GraphDataScience:
```
adamnsch, line 50: I guess this is ok, but probably cleaner to just use `pathlib` or something to get the path

adamnsch (commented):
Nice work :) Have some questions, and I notice a changelog entry is missing

TESTING.md
s1ck: Good point .. we renamed if after another comment and forgot the docs.

graphdatascience/graph_data_science.py
s1ck: I agree. I mean .. the `disable_server_verification` doesn't necessarily be exposed, but some builder or factory methods would be nice. We need to make some more changes wrt Arrow config, I can try to sketch something on that PR.

TESTING.md
s1ck: Are we doing integration testing on CI? We'd need a separate DB with encryption enabled. I can check the CI config.

s1ck: it's the `only` behaviour because you can't connect to a db which is set up for encryption with an unencrypted client connection.

adamnsch: Yes, we are doing integration test in CI (see `GdsClientTestBase`)

adamnsch: We could add another CI step to the pipeline which just runs `--encrypted` and has a DB set up for it

graphdatascience/tests/conftest.py
s1ck: `encrypted-only` :)

brs96:
[[PF] Find out flaky python client](https://trello.com/c/A4nJM7qD/6614-pf-find-out-flaky-python-client)

PR #141, opened 2022-08-08
Add tox section to run encrypted-only tests

s1ck:


tox.ini
```diff
@@ -12,3 +12,6 @@ deps =
 	pyarrow7: pyarrow >= 7.0, < 8.0
 	pyarrow8: pyarrow >= 8.0, < 9.0
 commands = pytest graphdatascience/tests --include-enterprise --include-model-store-location -Werror
+
+[testenv:encrypted]
+commands = pytest graphdatascience/tests --encrypted -Werror
```
s1ck, line 17: I am running this via `tox -e encrypted` which executed the environment using my local Python installation (3.8). I'd like to run this on the same combinations as the non-encrypted tests. Is there an elegant way to configure that?

adamnsch: I've been looking into this for a bit now, and honestly I'm a little disappointed in tox. envlist is only used if -e is not passed to the tox cli. And I think the easiest way for us to proceed is:
```[tox]
envlist =
	py{37,38,39}-pyarrow{4,5,6,7,8}-{standard,encrypted}
	py310-pyarrow{7,8}-{standard,encryped}

[testenv]
deps =
	pytest
	pyarrow4: pyarrow >= 4.0, < 5.0
	pyarrow5: pyarrow >= 5.0, < 6.0
	pyarrow6: pyarrow >= 6.0, < 7.0
	pyarrow7: pyarrow >= 7.0, < 8.0
	pyarrow8: pyarrow >= 8.0, < 9.0

commands =
	standard: pytest graphdatascience/tests --include-enterprise --include-model-store-location -Werror
	encrypted: pytest graphdatascience/tests --encrypted -Werror
```
And then calling the commands:
`tox -e $(tox -l | grep standard | paste -sd "," -)` and 
`tox -e $(tox -l | grep encrypted | paste -sd "," -)`
from the command line.

The alternative would be an extra tox.ini file for encrytpted, but that's slightly worse i think

adamnsch (changes requested):
Unfortunately this was harder than I thought

tox.ini
s1ck: It would be nice if `-e` would match on substrings, but yeah ...

s1ck: > Unfortunately this was harder than I thought

Thanks!

s1ck:
@adamnsch :ping_pong:

adamnsch (approved):

s1ck:
I think I need to merge this in order to make CI happy.

PR #146, opened 2022-08-10
Remove Python 3.10 support

s1ck:
We can not support Python 3.10 since the Neo4j Bolt driver does not support 3.10 and we heavily rely on it. The problem surfaced when using encrypted connections to the server where we ran into https://peps.python.org/pep-0644/

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,7 +1,5 @@
 [tox]
-envlist =
-	py{37,38,39}-pyarrow{4,5,6,7,8}-{standard,encrypted},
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -1,7 +1,5 @@
 [tox]
-envlist =
```
adamnsch, line 2: Faster testing :tada:

adamnsch (approved):
:+1:

brs96:
[[PF] Find out flaky python client](https://trello.com/c/A4nJM7qD/6614-pf-find-out-flaky-python-client)

PR #147, opened 2022-08-10
Add support for `gds.debug.arrow` endpoint

adamnsch:


adamnsch:
Decided along with @s1ck to scrap this PR as we don't want this endpoint directly exposed in the client

PR #148, opened 2022-08-10
Use pathlib to resolve test resources instead

adamnsch:


adamnsch:
There seem to be a issue here where some `mypy` versions (newer ones) warn about this. I use the latest version locally, and that one warns, but the version in CI hasn't been updated in a while I guess

graphdatascience/tests/integration/conftest.py
```diff
@@ -48,7 +48,7 @@ def gds() -> GraphDataScience:
 
 @pytest.fixture(scope="package")
 def gds_with_tls(request: FixtureRequest) -> GraphDataScience:
-    test_dir = os.path.dirname(request.path)
+    test_dir = os.path.dirname(request.path)  # type: ignore
```
s1ck, line 51: Maybe we can also use the approach that you suggested on the other PR. But it's also fine to ignore the warning.

s1ck (approved):

graphdatascience/tests/integration/conftest.py
adamnsch: I guess that would be easier. We should probably update the mypy in the docker image though

PR #153, opened 2022-08-16
DO NOT MERGE: Add script to reproduce getting a huge config from gds.graph.list

adamnsch:
Thanks to Nicola of the docs team for the code.


FlorentinD:
@nvitucci we merged an improvement and it will be in the next `alpha` release :)

PR #155, opened 2022-08-17
Fix doc version name

nvitucci:


recrwplay (approved):
Looks good.

I added a change to the version number for the sitemap extension in preview.yml. Strictly, since we don't use preview.yml in this repo to build docs for publishing we could remove the extension from that playbook, but I think it's good practice to always include it.

nvitucci:
> Looks good.
> 
> I added a change to the version number for the sitemap extension in preview.yml. Strictly, since we don't use preview.yml in this repo to build docs for publishing we could remove the extension from that playbook, but I think it's good practice to always include it.

Thanks, it makes sense - I forgot about it.

PR #162, opened 2022-08-25
Add support for testing targeting an AuraDS instance

adamnsch:


graphdatascience/tests/integration/conftest.py
```diff
@@ -55,7 +55,11 @@ def gds_with_tls(request: FixtureRequest) -> GraphDataScience:
         root_ca = f.read()
 
     _gds = GraphDataScience(
-        URI_TLS, auth=AUTH, arrow=True, arrow_disable_server_verification=True, arrow_tls_root_certs=root_ca
+        URI_TLS,
+        auth=AUTH,
+        arrow=True,
+        arrow_disable_server_verification=True,
+        arrow_tls_root_certs=root_ca,
```
s1ck, line 62: Weird that it didn't format that before. Is it because it's in test scope?

graphdatascience/tests/integration/conftest.py
```diff
@@ -71,6 +75,12 @@ def gds_without_arrow() -> GraphDataScience:
 
 
 def pytest_collection_modifyitems(config: Any, items: Any) -> None:
+    if config.getoption("--target-aura"):
```
s1ck, line 78: maybe .. since we call the other one `--include-enterprise` we can do `--exclude-aura`?

s1ck: it's a nit-pick so feel free to ignore. The other options are also kind of randomly named.

tox.ini
```diff
@@ -18,4 +18,5 @@ deps =
 commands =
 	standard: pytest graphdatascience/tests --include-enterprise --include-model-store-location -Werror
 	encrypted: pytest graphdatascience/tests --encrypted-only -Werror
+	aura: pytest graphdatascience/tests --include-enterprise --target-aura -Werror
```
s1ck, line 21: nice

s1ck (approved):
Looks good. Maybe the config can be named differently, but it's also fine as-is.

graphdatascience/tests/integration/conftest.py
adamnsch: The previous formatting was fine according to the rules of `black` I think. I just changed it for personal preference. It still obeys the rule we've set up

graphdatascience/tests/integration/conftest.py
adamnsch: Not sure I think the "exclude-aura" naming is great if it's an option for actually testing against aura. Maybe something like "exclude-self-managed-only" would be more logical, but too complicated I think. I will keep as is for now. We can rename later if we want

PR #164, opened 2022-08-30
Fix training NR pipelines

FlorentinD:


FlorentinD:
[Fix using NodeRegression pipelines in the client](https://trello.com/c/SHZOWHUn/6882-fix-using-noderegression-pipelines-in-the-client)

changelog.md
```diff
@@ -9,6 +9,8 @@
 
 ## Bug fixes
 
+* Fix training Node regession pipelines created via `gds.alpha.pipeline.nodeRegression.create`.
```
breakanalysis, line 12: I dont understand what bug this is fixing. Looks more like the whole feature was not exposed ?

breakanalysis, line 12:
```suggestion
* Fix training Node regression pipelines created via `gds.alpha.pipeline.nodeRegression.create`.
```

breakanalysis (changes requested):
Changes look good, however I think the changelog should specify what its fixing.

changelog.md
FlorentinD: It is a bug, as the feature was documented and therefore exposed

FlorentinD:
Thank you for the review I extended the change log entry, but kept it as a bug-fix as the feature was exposed to the user via Documentation as well as the python classes were present

PR #163, opened 2022-08-29
Add doc testing

adamnsch:

PR #166, opened 2022-08-31
Add graph construct notebook

nvitucci:
Proposal to add Jupyter notebooks + related docs to the GDS repo + docs

### DIFF
--- a/doc/modules/ROOT/pages/end-to-end-examples/load-data-via-graph-construction.adoc
+++ b/doc/modules/ROOT/pages/end-to-end-examples/load-data-via-graph-construction.adoc
@@ -1,0 +1,167 @@
+= Load data to a projected graph via graph construction
+
+[TIP]
+====
+Jupyter Notebook available on https://github.com/neo4j/graph-data-science-client/blob/main/examples/load-data-via-graph-construction.ipynb[Github^]
+====
+
+This notebook shows the usage of the `gds.alpha.graph.construct` method
+(available only in GDS 2.1+) to build a graph directly in memory.
+
+*NOTE:* If you are using AuraDS, it is currently not possible to write
+the projected graph back to Neo4j.
+
+== Setup
+
+We need an environment where Neo4j and GDS are available, for example
+AuraDS (which comes with GDS preinstalled) or Neo4j Desktop.
+
+Once the credentials to this environment are available, we can install
+the `graphdatascience` package and create the `gds` object.
+
+[source, python]
+----
+!pip install graphdatascience==1.3.0
+----
+
+[source, python]
+----
+# Import the client
+from graphdatascience import GraphDataScience
+
+# Replace with the actual credentials
+AURA_CONNECTION_URI = "neo4j+s://xxxxxxxx.databases.neo4j.io"
+AURA_USERNAME = "neo4j"
+AURA_PASSWORD = ""
+
+# Configure the client with AuraDS-recommended settings if using AuraDS
+gds = GraphDataScience(
+    AURA_CONNECTION_URI,
+    auth=(AURA_USERNAME, AURA_PASSWORD),
+    aura_ds=True
+)
+----
+
+We also import `pandas` to create a Pandas DataFrame from the original
+data source.
+
+[source, python]
+----
+import pandas as pd
+----
+
+== Load the Cora dataset
+
+[source, python]
+----
+# TODO: use URLs within the client repo when the notebook is added there
+CORA_CONTENT = "https://raw.githubusercontent.com/neo4j/graph-data-science/master/test-utils/src/main/resources/cora.content"
+CORA_CITES = "https://raw.githubusercontent.com/neo4j/graph-data-science/master/test-utils/src/main/resources/cora.cites"
+----
+
+We need to perform an additional preprocessing step to convert the
+`subject` field (which is a string in the dataset) into an integer,
+because node properties have to be numerical in order to be projected
+into a graph.
+
+[source, python]
+----
+# Using non-consecutive values to show class ordering later on
+
+SUBJECT_TO_ID = {
+    "Neural_Networks": 0,
+    "Rule_Learning": 1,
+    "Reinforcement_Learning": 2,
+    "Probabilistic_Methods": 3,
+    "Theory": 4,
+    "Genetic_Algorithms": 5,
+    "Case_Based": 6
+}
+----
+
+[source, python]
+----
+# Read each CSV locally as a Pandas DataFrame
+content = pd.read_csv(CORA_CONTENT, header=None)
+cites = pd.read_csv(CORA_CITES, header=None)
+----
+
+[source, python]
+----
+# Create a new DataFrame with a `nodeId` field, a list of node labels,
+# and the additional node properties `subject` and `features`
+nodes = pd.DataFrame().assign(
+    nodeId=content[0], 
+    labels="Paper", 
+    subject=content[1].replace(SUBJECT_TO_ID), 
+    features=content.iloc[:, 2:].apply(list, axis=1)
+)
+----
+
+[source, python]
+----
+# Show the first 5 rows of the new DataFrame
+nodes.head()
+----
+
+[source, python]
+----
+# Create a new DataFrame containing the relationships between the nodes.
+# To create the equivalent of an undirected graph, we need to add direct
+# and inverse relationships.
+dir_relationships = pd.DataFrame().assign(
+    sourceNodeId=cites[0],
+    targetNodeId=cites[1],
+    relationshipType="CITES"
+)
+
+inv_relationships = pd.DataFrame().assign(
+    sourceNodeId=cites[1],
+    targetNodeId=cites[0],
+    relationshipType="CITES"
+)
+
+relationships = pd.concat([dir_relationships, inv_relationships]).drop_duplicates()
+----
+
+[source, python]
+----
+# Show the first 20 rows of the new DataFrame
+relationships.head(5)
+----
+
+[source, python]
+----
+# Finally, create the in-memory graph
+G = gds.alpha.graph.construct(
+    "cora-graph",
+    nodes,
+    relationships
+)
+----
+
+[source, python]
+----
+# Check that the new graph has been created
+gds.graph.list()
+----
+
+[source, python]
+----
+# Count the nodes in the graph
+G.node_count()
+----
+
+[source, python]
+----
+# Get the value of the "subject" node property for
+# each node in the graph, printing only the first 5
+gds.graph.streamNodeProperties(G, ["subject"]).head()
+----
+
+== Cleanup
+
+[source, python]
+----
+G.drop()
+----
--- a/examples/load-data-via-graph-construction.ipynb
+++ b/examples/load-data-via-graph-construction.ipynb
@@ -1,0 +1,328 @@
+{
+  "cells": [
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "view-in-github",
+        "colab_type": "text"
+      },
+      "source": [
+        "<a href=\"https://colab.research.google.com/github/nvitucci/graph-data-science-client/blob/add-construct-notebook/examples/load-data-via-graph-construction.ipynb\" target=\"_parent\"><img src=\"https://colab.research.google.com/assets/colab-badge.svg\" alt=\"Open In Colab\"/></a>"
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "KhH0v_bwUrXp"
+      },
+      "source": [
+        "# Load data to a projected graph via graph construction\n",
+        "\n",
+        "This notebook shows the usage of the `gds.alpha.graph.construct` method (available only in GDS 2.1+) to build a graph directly in memory.\n",
+        "\n",
+        "**NOTE:** If you are using AuraDS, it is currently not possible to write the projected graph back to Neo4j."
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "X5ttqidnWPTC"
+      },
+      "source": [
+        "## Setup\n",
+        "\n",
+        "We need an environment where Neo4j and GDS are available, for example AuraDS (which comes with GDS preinstalled) or Neo4j Desktop. \n",
+        "\n",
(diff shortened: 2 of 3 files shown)


adamnsch:
The test that was causing CI failure is fixed on `main` now :)

adamnsch:
We're adding document testing in this PR https://github.com/neo4j/graph-data-science-client/pull/163. So you could either make the asciidoc testable or give the python source code blocks the role `no-test` to avoid the test script picking them up :) let us know if you have any questions on this

nvitucci:
> We're adding document testing in this PR #163. So you could either make the asciidoc testable or give the python source code blocks the role `no-test` to avoid the test script picking them up :) let us know if you have any questions on this

If making them testable is stable enough I can give it a try, otherwise I can mark them as `no-test` for now and do this later.

This also made me think of something else: I've treated the login as Aura-based by default, do you think I should add a local-based example cell too?

Mats-SX:
I think the plan was to add the docs page version of the notebook to the GDS Manual (server), not the client. There is already a chapter End-to-end examples in the GDS Manual: https://neo4j.com/docs/graph-data-science/current/end-to-end-examples/end-to-end-examples/

nvitucci:
> I think the plan was to add the docs page version of the notebook to the GDS Manual (server), not the client. There is already a chapter End-to-end examples in the GDS Manual: https://neo4j.com/docs/graph-data-science/current/end-to-end-examples/end-to-end-examples/
> 
> So I think we should split this PR in two: one that adds the actual notebook and one that adds the docs page, but as a PR to https://github.com/neo-technology/graph-analytics

It would be fine for me, but the main difference would be that this notebook is completely client-focused, while the examples we have now in the GDS manual are Cypher-focused. Would this still be ok?

Mats-SX:
Additionally, we may want to move the Python Client placeholder chapter to before End-to-end examples, so that we introduce it before we have several pages using/referencing it.

nvitucci:
> > > I think the plan was to add the docs page version of the notebook to the GDS Manual (server), not the client. There is already a chapter End-to-end examples in the GDS Manual: https://neo4j.com/docs/graph-data-science/current/end-to-end-examples/end-to-end-examples/
> > > So I think we should split this PR in two: one that adds the actual notebook and one that adds the docs page, but as a PR to https://github.com/neo-technology/graph-analytics
> > 
> > 
> > It would be fine for me, but the main difference would be that this notebook is completely client-focused, while the examples we have now in the GDS manual are Cypher-focused. Would this still be ok?
> 
> Good point, I hadn't thought of that. My vision is that we will eventually have Python Client code examples all over the server manual (new norm). Currently we have none, after having extracted the client manual. I am torn on the idea of having end-to-end examples in the client manual. I feel like the client manual is best if kept minimal and to the point of reference documentation, rather than use-case or tutorial-style. The server manual already escaped that with the addition of the End-to-end examples chapter.

In my view, the end-to-end examples in the client manual is a good location for this kind of content because it shows not only how to use the client itself, but a whole Python workflow (for instance all the Pandas stuff, which is sort of less related to GDS specifically). Anyway, I'll leave this decision to you.

> This has me reaching for these conclusions:
> 
>     * we add this Python-centric example to the end-to-end chapter
> 
>     * we create another card for writing out also a Cypher-tabbed version of the same example
> 
>     * we create a card to add also a tabbed version of the current fastrp+knn example with Python Client syntax
> 
>     * repeat for the next notebook you wrote
> 
> 
> What do you think?

We can certainly do that (which would be in line with the examples we have for AuraDS), although that will mean duplicating the code and the required maintenance, especially if the end goal is to only use client code.

What if we create subsections of the "end-to-end examples" section instead - would it be more clear or more confusing? I'm thinking something on the lines of:

```
End-to-end examples
    Cypher examples
        FastRP + kNN
        ...
    Python examples
        Graph construction
        ...
```

Mats-SX:
@nvitucci Okay, I think you have a good perspective. I think I'll retract my comment about having to split the work in this PR. Let's continue the way you've implemented it (pending the content review), and then we can discuss later the end game for the content in the server manual vs client. 

Honestly I'm open to any suggestion from you as this does transcend the level of the documentation that I'm confident that the GDS team can own and develop effectively. I'm glad you're leading the charge here.

nvitucci:
> @nvitucci Okay, I think you have a good perspective. I think I'll retract my comment about having to split the work in this PR. Let's continue the way you've implemented it (pending the content review), and then we can discuss later the end game for the content in the server manual vs client.
>

Sure, I agree. So, to summarise:

* we keep both the docs (in the `End-to-end examples` section) and the notebook (`examples` directory) in this repo;
* additionally, I'd suggest to add the source data as well (the Cora data files, in this case) to this repo, for instance in a `data` subdirectory of `examples`;
* we do not add or change anything in the GDS server docs for the time being (we will review this later on, as you mentioned).

> Honestly I'm open to any suggestion from you as this does transcend the level of the documentation that I'm confident that the GDS team can own and develop effectively. I'm glad you're leading the charge here.

No worries, I think we are still testing the waters here and will improve according to what proves to be most useful to our users.

adamnsch:
> * additionally, I'd suggest to add the source data as well (the Cora data files, in this case) to this repo, for instance in a `data` subdirectory of `examples`;

We have a PR for this now: https://github.com/neo4j/graph-data-science-client/pull/173
We're just figure out licensing concerns, though we might want to just merge it for now

doc/modules/ROOT/pages/end-to-end-examples/load-data-via-graph-construction.adoc
```diff
@@ -0,0 +1,167 @@
+= Load data to a projected graph via graph construction
+
+[TIP]
+====
+Jupyter Notebook available on https://github.com/neo4j/graph-data-science-client/blob/main/examples/load-data-via-graph-construction.ipynb[Github^]
```
adamnsch, line 5: This link should probably point to the branch that corresponds to the pip installed version below

doc/modules/ROOT/pages/end-to-end-examples/load-data-via-graph-construction.adoc
```diff
@@ -0,0 +1,167 @@
...
+
+We need an environment where Neo4j and GDS are available, for example
+AuraDS (which comes with GDS preinstalled) or Neo4j Desktop.
+
+Once the credentials to this environment are available, we can install
+the `graphdatascience` package and create the `gds` object.
+
+[source, python]
+----
+!pip install graphdatascience==1.3.0
```
adamnsch, line 24: Since we are currently working on `1.4` we should have:
```suggestion
!pip install graphdatascience==1.4
```

doc/modules/ROOT/pages/end-to-end-examples/load-data-via-graph-construction.adoc
```diff
@@ -0,0 +1,167 @@
...
+Jupyter Notebook available on https://github.com/neo4j/graph-data-science-client/blob/main/examples/load-data-via-graph-construction.ipynb[Github^]
+====
+
+This notebook shows the usage of the `gds.alpha.graph.construct` method
+(available only in GDS 2.1+) to build a graph directly in memory.
+
+*NOTE:* If you are using AuraDS, it is currently not possible to write
+the projected graph back to Neo4j.
+
+== Setup
```
adamnsch, line 14: General nitpick: Our asciidoc code style is to have to blank lines before a new header (https://github.com/neo4j/graph-data-science-client/tree/main/doc#authoring)

doc/modules/ROOT/pages/end-to-end-examples/load-data-via-graph-construction.adoc
```diff
@@ -0,0 +1,167 @@
...
+
+We need an environment where Neo4j and GDS are available, for example
+AuraDS (which comes with GDS preinstalled) or Neo4j Desktop.
+
+Once the credentials to this environment are available, we can install
+the `graphdatascience` package and create the `gds` object.
+
+[source, python]
```
adamnsch, line 22: General comment: If we don't make it testable (indeed this may be too much work atm) we need to add `role=no-test` to the python source blocks

doc/modules/ROOT/pages/end-to-end-examples/load-data-via-graph-construction.adoc
```diff
@@ -0,0 +1,167 @@
+= Load data to a projected graph via graph construction
+
+[TIP]
+====
+Jupyter Notebook available on https://github.com/neo4j/graph-data-science-client/blob/main/examples/load-data-via-graph-construction.ipynb[Github^]
+====
+
+This notebook shows the usage of the `gds.alpha.graph.construct` method
```
adamnsch, line 8: Nice intro :)

examples/load-data-via-graph-construction.ipynb
```diff
@@ -0,0 +1,326 @@
+{
+  "cells": [
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "colab_type": "text",
+        "id": "view-in-github"
+      },
+      "source": [
+        "<a href=\"https://colab.research.google.com/github/nvitucci/graph-data-science-client/blob/add-construct-notebook/examples/load-data-via-graph-construction.ipynb\" target=\"_parent\"><img src=\"https://colab.research.google.com/assets/colab-badge.svg\" alt=\"Open In Colab\"/></a>"
```
adamnsch, line 10: Is this colab source relevant?

doc/modules/ROOT/pages/end-to-end-examples/load-data-via-graph-construction.adoc
adamnsch: I guess we can use `{docs-version}` like we do in https://github.com/neo4j/graph-data-science-client/blob/main/doc/modules/ROOT/pages/index.adoc for this :) (below as well)

adamnsch (changes requested):
Looks great! Just have some minor thoughts :)

The CI failure is just some styling thing. If you just `scripts/makestyle` it should be fine (https://github.com/neo4j/graph-data-science-client/blob/main/TESTING.md#style-guide for more info)

adamnsch:
I think we also want to list this notebook along side the other one in the README: https://github.com/neo4j/graph-data-science-client#usage

nvitucci:
> > * additionally, I'd suggest to add the source data as well (the Cora data files, in this case) to this repo, for instance in a `data` subdirectory of `examples`;
> 
> We have a PR for this now: #173 We're just figure out licensing concerns, though we might want to just merge it for now

Nice. When this is done, I can replace the links to download the data.

doc/modules/ROOT/pages/end-to-end-examples/load-data-via-graph-construction.adoc
nvitucci: Yes, makes sense. Fixed.

doc/modules/ROOT/pages/end-to-end-examples/load-data-via-graph-construction.adoc
nvitucci: Sure thing.

examples/load-data-via-graph-construction.ipynb
nvitucci: You are right, it should point to this repo instead (it points to mine because we are using a fork). Let me try and fix it.

doc/modules/ROOT/pages/end-to-end-examples/load-data-via-graph-construction.adoc
```diff
@@ -0,0 +1,170 @@
...
+
+We need an environment where Neo4j and GDS are available, for example
+AuraDS (which comes with GDS preinstalled) or Neo4j Desktop.
+
+Once the credentials to this environment are available, we can install
+the `graphdatascience` package and create the `gds` object.
+
+[source, python, role=no-test]
+----
+!pip install graphdatascience==1.4
```
adamnsch, line 25: I meant that we could use `{docs-version}` here as well :)

nvitucci: Ah yes, good idea. Done.

adamnsch (approved):
Looks good, but seems like you have to rebase :)

nvitucci:
A few things I've discussed with @adamnsch before the PR is completed and merged:

- If we plan to add examples that are not really "end to end" (such as the graph construct example), we may want to rename the section from "End-to-end examples" to just "Examples" or "Tutorials".
- The current example makes Aura the prominent connection example, but it does not have to be the case. We can either add a link to alternative connections methods or have additional cells (potentially commented) with them.
- Before a new version of the client is published, a notebook cannot make use of that version explicitly. For example, if the dev version is 1.4, the notebook cannot yet use 1.4 but has to use at least the previous release. We decided 1) to keep the latest version explicit for clarity (i.e. instead of `pip install graphdatascience` we use `pip install graphdatascience==x.y-1`) and 2) to bump the version after the release. Possible alternatives can be to either skip the version or to skip the library installation altogether (as in the new PyG example notebook).
- We'll be keeping some output cells, so that interesting and relevant results can be shown and described more easily.

Any feedback is welcome. We can still revisit later.

nvitucci:
> I think something went wrong in your rebase, so there's lots of extra commits on your branch

I've just realised the same. I'm investigating this.

nvitucci:
> > I think something went wrong in your rebase, so there's lots of extra commits on your branch
> 
> I've just realised the same. I'm investigating this.

It seems it was just a Github artifact. It shows properly for me now.

examples/load-data-via-graph-construction.ipynb
```diff
@@ -0,0 +1,717 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 26,
+   "metadata": {
+    "id": "tOZW-I2FBLm_"
+   },
+   "outputs": [],
+   "source": [
+    "# TODO: use URLs within the client repo when the notebook is added there\n",
```
adamnsch, line 138: FYI `gds.graph.load_cora` is merged now. But I don't think that should affect this notebook. However, I don't think we will necessarily have the content files raw in this repo so this TODO probably isn't relevant anymore.

adamnsch: Yeah I think it's fine, but maybe we still should have a TODO actually regarding putting the data in a more suitable/permanent place. Either is fine I think :)

PR #180, opened 2022-09-16
Add github templates

FlorentinD:
Helping users to write good issues

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,33 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: BUG
+assignees: ''
+
+---
+
+**Describe the bug**
+<!-- A clear and concise description of what the bug is. If you think its directly related to the GDS plugin, please open the issue at https://github.com/neo4j/graph-data-science/issues -->
+
+**To Reproduce**
+<!-- Running `python_code` helps to provide vital details -->
+
+graphdatascience library version: X.Y
+GDS plugin version: X.Y.Z
+Python version: X.Y.Z
+Neo4j version: X.Y.Z
+Operating system: (for example Windows 95/Ubuntu 16.04)
+
+Steps to reproduce the behavior:
+- [ ] Graph/dataset to reproduce the bug.  Alternatively describe as much as possible the data, e.g., graph size and other schema/structure details.
+- [ ] Python code
+- [ ] Other queries not run through the python client
+
+**Expected behavior**
+<!-- A clear and concise description of what you expected to happen. -->
+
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,32 @@
...
+
+---
+
+**Describe the bug**
+<!-- A clear and concise description of what the bug is. -->
+
+**To Reproduce**
+<!-- Running `python_code` helps to provide vital details -->
+
+Python Client version: X.Y.Z
```
adamnsch, line 16: The client only has major and minor now `X.Y`

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,32 @@
...
+GDS version: X.Y.Z
+Neo4j version: X.Y.Z
```
adamnsch, line 18: We might be interested in Python version too?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,10 @@
+Thank you for your contribution to the Graph Data Science Client project.
+
+<!-- Please include a summary of the change, such as, which issue was fixed or feature was added. 
+If relevant, link to the corresponding issue.
+Please also include relevant motivation and context. List any dependencies that are required for this change. -->
+
+Before submitting this PR, please make sure:
+- [ ] You signed the [Neo4j CLA](https://neo4j.com/developer/cla/#sign-cla) (Contributor License Agreement) so that we are allowed to ship your code in our library
```
adamnsch, line 8: All the details are outlined in `CONTRIBUTING.md` so we might want to refer to that?

adamnsch (approved):
Nice :) Me and @Mats-SX deliberately did not add these templates when setting up the repo. The reason was that the threshold for opening something should be as low as possible. I think that's still valid. But I definitely also see the point of these templates :)

.github/ISSUE_TEMPLATE/bug_report.md
adamnsch: Also we might want to say "`graphdatascience` library version" to be clearer?

.github/ISSUE_TEMPLATE/bug_report.md
FlorentinD: good point :+1:

.github/PULL_REQUEST_TEMPLATE.md
FlorentinD: added a reference :+1:

FlorentinD:
Good point, I added a sentence in the issue template that links to our graph-data-science repository

PR #184, opened 2022-09-21
Make neo4j 5.X driver compatible

adamnsch:
Co-Authored-By: Florentin Dörre <florentin.dorre@neotechnology.com>
Co-Authored-By: Brian Shi <brian.shi@neotechnology.com>

Make sure:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,6 +20,7 @@ classifiers = [
     "Programming Language :: Python :: 3.7",
     "Programming Language :: Python :: 3.8",
     "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
     "Topic :: Database",
     "Topic :: Scientific/Engineering",
     "Topic :: Software Development",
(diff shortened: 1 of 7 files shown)


setup.py
```diff
@@ -20,6 +20,7 @@
     "Programming Language :: Python :: 3.7",
     "Programming Language :: Python :: 3.8",
     "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
```
knutwalker, line 23: Do we need to note something in the docs or the readme, that 3.10 is only supported when the neo4j 5.x driver is used? Or are we supporting 3.10 for all versions?

knutwalker (approved):
👍

setup.py
adamnsch: Good points. We need to add a new badge to the readme, and to the compat matrix in the docs. I think pip will handle the fact that 5.X should only be used with py3.10

adamnsch:
> +1

@knutwalker @FlorentinD  To make the tests pass, which of the following would you prefer?

1. Catch the deprecation warning for now
2. Run the encrypted tests without `-Werror`

FlorentinD:
I would opt for caching it for now

setup.py
adamnsch: done

Issue #179, opened 2022-09-16
streamNodeProperty() doesn't work with gds.run_cypher() as I guess


MOSSupport:
graphdatascience 1.3

I tried a query like this:

```
query = f'''
   call gds.graph.streamNodeProperty(
      'xxx',
      'xxxx',
      ['xxxxx']
   )
yield nodeId as id, propertyValue as degree
return id, degree limit 100
...
result = gds.run_cypher(query)
```
=> KeyError: 'graph_name'

I figured out to make it work like this:

```
query = f'''
   ...
'''
params = {
   'graph_name': 'xxx',
   'properties': 'xxxx',
   'entities'" ['xxxxx'],
   'config': ''
}
result = gds.run_cypher(query, params)
```

=> No error, but it returned all rows(not limited to 100) as nodeId and propertyValue(not renamed as id and degree)

Other cypher queries works with gds.run_cypher(query) as expected.

MOSSupport:
Hi,
I just tried to reduce the number of rows in the result for test purpose.

Thanks.
Dongho.

FlorentinD:
For further help, we need the GDS and Neo4j version you were running the queries against.

When I tried it in my test environment, I could get the expected result.

MOSSupport:
Neo4j 4.4.8 Enterprise image(Debian 11, openjdk 11.0.15) from docker hub, I'm running it on RHEL 7.9.
GDS 2.1.9

FlorentinD:
Hello,
I used the same versions as you, but I still see the rename of the property as well as the limit to only 100 rows.
Could you try running a similar query with our example notebook (https://github.com/neo4j/graph-data-science-client/blob/main/examples/load-data-via-graph-construction.ipynb)?

The query I used based on your description was
`gds.run_cypher("CALL gds.graph.streamNodeProperty($graph_name, $property, $nodeLabels) YIELD nodeId AS id, propertyValue AS degree RETURN id, degree LIMIT 10", 
{"graph_name": G.name(), "property": "subject", "nodeLabels": ["Paper"]})`

MOSSupport:
I tried the code you typed above. But still the same "KeyError" returned as I explained above.
And I found one more thing:
[...]

PR #185, opened 2022-09-21
Investigate filtered stream properties with arrow problem

FlorentinD:
Trying to highlight another bug reported in #179


FlorentinD:
[Arrow streamNodeProperties does not return correct nodeIds](https://trello.com/c/b7EBVIT8/7000-arrow-streamnodeproperties-does-not-return-correct-nodeids)

PR #192, opened 2022-09-30
Test only one environment targeting AuraDS

adamnsch:

Make sure:


FlorentinD (approved):
Very reasonable change.

For PRs we decided we want faster build times.
Now with Aura we saw flaky behaviour which should be reduced with only testing one combination. Especially as we dont expect specific deviation on Aura based on dependency versions but rather the Aura environment in general.

As a side-effect we are also saving some cloud time. So environment-friendly even :)

PR #196, opened 2022-10-10
Make `gds.graph.get` resolve other users' graphs for admin users

adamnsch:

Make sure:

### DIFF
--- a/graphdatascience/graph/graph_proc_runner.py
+++ b/graphdatascience/graph/graph_proc_runner.py
@@ -106,7 +106,8 @@ class GraphProcRunner(CallerBase, UncallableNamespace, IllegalAttrChecker):
 
     @client_only_endpoint("gds.graph")
     def get(self, graph_name: str) -> Graph:
-        if not self.exists(graph_name)["exists"]:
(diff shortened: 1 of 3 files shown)


graphdatascience/graph/graph_proc_runner.py
```diff
@@ -106,7 +106,8 @@ def list(self, G: Optional[Graph] = None) -> DataFrame:
 
     @client_only_endpoint("gds.graph")
     def get(self, graph_name: str) -> Graph:
-        if not self.exists(graph_name)["exists"]:
+        result = self._query_runner.run_query(f"CALL gds.graph.list('{graph_name}') YIELD graphName")
+        if len(result["graphName"]) == 0:
```
Mats-SX, line 110: We don't have a test for this using an `admin` user from what I can see. I think we should add one.

Mats-SX (changes requested):
I think we should also add a test

graphdatascience/graph/graph_proc_runner.py
adamnsch: I don't think so. That would be like testing that pagerank generates the correct result in a client test in my opinion. I.e. I think it's a server side concern that should be (and is) tested there

Mats-SX (approved):

PR #202, opened 2022-10-21
Support setting and list defaults

FlorentinD:

Make sure:

At the moment, this fails as a `dict` is not hash-able for `params = {"config": config}`. 
I did not understand why it works in other places though such as for the typical algorithms yet.

### DIFF
--- a/graphdatascience/system/system_endpoints.py
+++ b/graphdatascience/system/system_endpoints.py
@@ -1,4 +1,4 @@
-from typing import Any, Optional
+from typing import Any, Dict, Optional
 
 from pandas import DataFrame, Series
 
@@ -64,3 +64,33 @@ class IndirectSystemEndpoints(CallerBase):
         query = f"CALL {self._namespace}()"
 
         return self._query_runner.run_query(query).squeeze()  # type: ignore
+
+    def set(self, key: str, value: Any, username: Optional[str] = None) -> None:
+        self._namespace += ".set"
+
+        params = {"key": key, "value": value}
+
+        # Checking for explicit None as '' is a valid user
+        if username is not None:
+            query = f"CALL {self._namespace}($key, $value, $username)"
+            params["username"] = username
+        else:
+            query = f"CALL {self._namespace}($key, $value)"
+
+        self._query_runner.run_query(query, params)
+
+    def list(self, key: Optional[str] = None, username: Optional[str] = None) -> DataFrame:
+        self._namespace += ".list"
+
+        config: Dict[str, Any] = {}
+
+        if key:
+            config["key"] = key
+        # Checking for explicit None as '' is a valid user
+        if username is not None:
+            config["username"] = username
+
+        query = f"CALL {self._namespace}($config)"
+        params = {"config": config}
+
+        return self._query_runner.run_query(query, params)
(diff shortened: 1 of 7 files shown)


FlorentinD:
[Several server endpoints not supported in Python client](https://trello.com/c/EdLW5RNy/7082-several-server-endpoints-not-supported-in-python-client)

graphdatascience/system/system_endpoints.py
```diff
@@ -39,6 +81,10 @@ def is_licensed(self) -> bool:
     def debug(self) -> DebugProcRunner:
         return DebugProcRunner(self._query_runner, f"{self._namespace}.debug", self._server_version)
 
+    @property
+    def config(self) -> DebugProcRunner:
+        return DebugProcRunner(self._query_runner, f"{self._namespace}.config", self._server_version)
```
adamnsch, line 86:
```suggestion
        return DefaultsAndLimitsEndPoints(self._query_runner, f"{self._namespace}.config", self._server_version)
```
right?

graphdatascience/system/system_endpoints.py
```diff
@@ -16,6 +16,48 @@ def sysInfo(self) -> "Series[Any]":
...
+        if username is not None:
+            config["username"] = username
+
+        query = f"CALL {self._namespace}($config)"
+        params = {"config": config}
+
+        return self._query_runner.run_query(query, params)
+
+
+class DefaultsAndLimitsEndPoints(CallerBase):
```
adamnsch, line 51: This also needs to inherit from `UncallableNamespace` and `IllegalAttrChecker` for nicer error handling

adamnsch (changes requested):
Looks good! Just some minor remarks

graphdatascience/system/system_endpoints.py
FlorentinD: After config, you can only use `list` or `set` so the `DebugProcRunner` should be correct. checkstyle also complained there

adamnsch: Actually I think this should be removed. `config` is not a direct endpoint as it's prefixed by `alpha`. You actually never use it

graphdatascience/system/system_endpoints.py
adamnsch, line 51: I think I would make a `config_endpoints.py` module and put them there

graphdatascience/system/system_endpoints.py
FlorentinD: My thinking was, that it gives a better auto-completion.

But then as we also dont have endpoints for individual algorithms, this probably falls into a similar category.
Will remove it :+1:

adamnsch: It doesn't give any autocompletion as you never use it. You would need to write `gds.config.defaults....` since it's a _direct endpoint_, but that's not gonna work since the `DebugProcRunner` doesn't have a `defaults` attribute. Moreover, since `config` is not preceded by `alpha` the endpoint is not correct. Actually it does suggest `gds.config...` for autocompletion but that's misleading

FlorentinD: thanks for the clarification!

PR #211, opened 2022-11-01
Skip defaults tests when targeting AuraDS

adamnsch:
Since the "neo4j" user lacks auth there.

Make sure:


FlorentinD:
setting it to neo4j might not work for the other non-aura versions :thinking:

adamnsch:
> setting it to neo4j might not work for the other non-aura versions thinking

good point

FlorentinD:
just verified, that locally i would get ` User '' not authorized to set default for user 'neo4j'` otherwise

FlorentinD (approved):

PR #218, opened 2022-11-11
Fix broken link

nvitucci:


adamnsch (approved):
Nice! @nvitucci I think we need to port it to the 1.5 (current) branch too right?

PR #219, opened 2022-11-17
Add Node Regression notebook and dataset

Mats-SX:
[Direct link to notebook](https://github.com/neo4j/graph-data-science-client/blob/2fcb088eeaf09322a3f9d60dfab4022c5d0afb16/examples/node-regression-with-subgraph-and-graph-sample.ipynb)

~~Keeping this as draft until we have confirmation of ability to host the dataset.~~


examples/node-regression-with-subgraph-and-graph-sample.ipynb
```diff
@@ -0,0 +1,713 @@
...
+    "from graphdatascience import GraphDataScience\n",
+    "\n",
+    "# Replace with the actual connection URI and credentials\n",
+    "NEO4J_CONNECTION_URI = \"neo4j+s://<aurads-instance>.databases.neo4j.io\"\n",
+    "NEO4J_USERNAME = \"neo4j\"\n",
+    "NEO4J_PASSWORD = \"\"\n",
+    "\n",
+    "\n",
+    "# Connect to Neo4j and create the GraphDataScience object\n",
+    "gds = GraphDataScience(NEO4J_CONNECTION_URI, auth=(NEO4J_USERNAME, NEO4J_PASSWORD), aura_ds=True)\n",
```
breakanalysis, line 81: perhaps mention that aura_ds should be false if local db ?

examples/node-regression-with-subgraph-and-graph-sample.ipynb
```diff
@@ -0,0 +1,713 @@
...
+   }
+  },
+  {
+   "cell_type": "markdown",
+   "source": [
+    "# Preparing the dataset for the pipeline\n",
+    "\n",
+    "In order to use the dataset, we must make sure that the feature dimension is consistent over the entire node sets.\n",
+    "In their raw form, the feature vectors are of different lengths.\n",
+    "To overcome this, we will use a one-hot encoding.\n",
```
breakanalysis, line 291: i would say the main reason to use one-hot encoding is that we dont support categorical id's or sets of those as features. i dont think other libraries do either because its not something a typical model understands and you could perhaps say they are not even features. 

in a way you could say a side effect of the one-hot encoding is that you get same length feature vectors.

Maybe you could write that the data is not in an ml-friendly format because most or all ml-models cannot work with features whose values are id's. using one-hot encoding transforms the data to ml-friendly vectors (?)

examples/node-regression-with-subgraph-and-graph-sample.ipynb
```diff
@@ -0,0 +1,713 @@
...
+      "Name: 0, dtype: object\n"
+     ]
+    }
+   ],
+   "source": [
+    "# Now, we can use the subgraph projection to separate out only one of the components\n",
+    "G_chameleon, subgraph_result = gds.beta.graph.project.subgraph(\n",
+    "    \"chameleon_component\",\n",
+    "    G_animals,\n",
+    "    \"n.wcc_component = 0\",  # Component with id 0 corresponds to the Chameleon part of the graph\n",
```
breakanalysis, line 472: how do they know that ? perhaps write a cyperquery that shows the max and min component number of the chameleon nodes to be `0` ?

examples/node-regression-with-subgraph-and-graph-sample.ipynb
```diff
@@ -0,0 +1,713 @@
...
+      "File \u001B[0;32m/opt/homebrew/Caskroom/miniconda/base/lib/python3.9/site-packages/neo4j/io/_common.py:244\u001B[0m, in \u001B[0;36mResponse.on_failure\u001B[0;34m(self, metadata)\u001B[0m\n\u001B[1;32m    242\u001B[0m \u001B[38;5;28;01mif\u001B[39;00m callable(handler):\n\u001B[1;32m    243\u001B[0m     handler()\n\u001B[0;32m--> 244\u001B[0m \u001B[38;5;28;01mraise\u001B[39;00m Neo4jError\u001B[38;5;241m.\u001B[39mhydrate(\u001B[38;5;241m*\u001B[39m\u001B[38;5;241m*\u001B[39mmetadata)\n",
+      "\u001B[0;31mClientError\u001B[0m: {code: Neo.ClientError.Procedure.ProcedureCallFailed} {message: Failed to invoke procedure `gds.alpha.pipeline.nodeRegression.addNodeProperty`: Caused by: java.lang.IllegalArgumentException: Could not find a procedure called gds.fastrp.mutate}"
+     ]
+    }
+   ],
+   "source": [
+    "# Now, let's construct a training pipeline!\n",
```
breakanalysis, line 545: I think this merits a cell with a larger heading

breakanalysis (approved):
Looks good! I have a few comments.
Also would be good to clear the cell output from the notebook and save it again as it shows errors.

examples/node-regression-with-subgraph-and-graph-sample.ipynb
Mats-SX: You mean a markdown cell with only that line as a title?

Mats-SX: I added a cell to introduce it.

examples/node-regression-with-subgraph-and-graph-sample.ipynb
Mats-SX: Due to tie-breaking rules, there will always be a component with id 0 (lower id wins). But we don't know what labels that indicate without Cyphering. I will add something.

Mats-SX: Ach, I don't want to use Cypher in the notebook tho -- I said above it was pure Python.

examples/node-regression-with-subgraph-and-graph-sample.ipynb
Mats-SX: Solved together!

examples/node-regression-with-subgraph-and-graph-sample.ipynb
Mats-SX: OK I ended up expanding this part quite a bit.

Mats-SX:
@breakanalysis Thanks for the great comments. I've made a larger update in the last commit -- could you have another look and see what you think?

breakanalysis (commented):
I like the use of WCC and looking at labels of each component :)

PR #238, opened 2023-01-05
broken link fix after flagged on semrush report

lidiazuin:
Wrong link was leading to broken link page. This fix should be cherry-picked to all versions of the docs.


adamnsch (approved):
Great!

PR #246, opened 2023-01-11
Method for getting DB node properties

brs96:

Make sure:

### DIFF
--- a/graphdatascience/graph/graph_entity_ops_runner.py
+++ b/graphdatascience/graph/graph_entity_ops_runner.py
@@ -94,7 +94,9 @@ class GraphNodePropertiesRunner(GraphEntityOpsBaseRunner):
             db_nodes = self._util_proc_runner.asNodes(nodeIds)
             properties = []
             for node in db_nodes:
-                properties.append(list([node.id] + list(map(lambda p: node._properties[p], db_node_properties))))
+                properties.append(
+                    list([node.id] + list(map(lambda p: node._properties[p], db_node_properties)))  # type: ignore
+                )
             db_properties_df = pd.DataFrame(properties, columns=["nodeId"] + db_node_properties).drop_duplicates(
                 subset=["nodeId"]
             )
--- a/graphdatascience/graph/graph_entity_ops_runner.py
+++ b/graphdatascience/graph/graph_entity_ops_runner.py
@@ -1,3 +1,4 @@
+from functools import reduce
 from typing import Any, Dict, List, Type, Union
 
 import pandas as pd
@@ -90,15 +91,9 @@ class GraphNodePropertiesRunner(GraphEntityOpsBaseRunner):
             )
 
         if db_node_properties:
-            nodeIds = result["nodeId"].tolist()
-            db_nodes = self._util_proc_runner.asNodes(nodeIds)
-            properties = []
-            for node in db_nodes:
-                properties.append(
-                    list([node.id] + list(map(lambda p: node.get(p), db_node_properties)))  # type: ignore
-                )
-            db_properties_df = pd.DataFrame(properties, columns=["nodeId"] + db_node_properties).drop_duplicates(
-                subset=["nodeId"]
+            unique_node_ids = result["nodeId"].drop_duplicates().tolist()
+            db_properties_df = self._query_runner.run_query(
+                self._build_query(db_node_properties), {"ids": unique_node_ids}
             )
             if "propertyValue" not in result.keys():
                 result = result.join(db_properties_df.set_index("nodeId"), on="nodeId")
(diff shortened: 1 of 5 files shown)


brs96:
[Convenience to resolve nodeIds in result df](https://trello.com/c/jXruxIma/7551-convenience-to-resolve-nodeids-in-result-df)

graphdatascience/graph/graph_entity_ops_runner.py
```diff
@@ -1,7 +1,9 @@
 from typing import Any, Dict, List, Type, Union
 
+import pandas as pd
 from pandas import DataFrame, Series
```
FlorentinD, line 4: I think we should do either way of these imports. 
Leaning towards the `from` version :)

graphdatascience/graph/graph_entity_ops_runner.py
```diff
@@ -85,6 +89,25 @@ def stream(
                 columns={"variable": "nodeProperty", "value": "propertyValue"}
             )
 
+        if db_node_properties:
+            nodeIds = result["nodeId"].tolist()
+            db_nodes = self._util_proc_runner.asNodes(nodeIds)
```
FlorentinD, line 94: this is not very efficient as we get the whole node objects via bolt.

Would be interesting how large the performance impact is, but I would rather use a specific YIELD clause in the cypher query to only return the desired properties instead of returning the whole nodes.

graphdatascience/graph/graph_entity_ops_runner.py
```diff
@@ -85,6 +89,25 @@ def stream(
...
+            properties = []
+            for node in db_nodes:
+                properties.append(
+                    list([node.id] + list(map(lambda p: node.get(p), db_node_properties)))  # type: ignore
+                )
+            db_properties_df = pd.DataFrame(properties, columns=["nodeId"] + db_node_properties).drop_duplicates(
+                subset=["nodeId"]
+            )
```
FlorentinD, line 102: We should drop the duplicates before passing the node list to cypher.

graphdatascience/graph/graph_entity_ops_runner.py
```diff
@@ -85,6 +89,25 @@ def stream(
...
+            if "propertyValue" not in result.keys():
+                result = result.join(db_properties_df.set_index("nodeId"), on="nodeId")
+            else:
+                db_properties_df = db_properties_df.melt(id_vars=["nodeId"]).rename(
```
FlorentinD, line 106: we only need melt as you return the whole node objects. So just returning the desired properties + nodeId would make this part also easier :)

graphdatascience/query_runner/neo4j_query_runner.py
```diff
@@ -45,6 +45,10 @@ def run_query(
                     "ignore",
                     message=r"^ssl.OP_NO_SSL\*/ssl.OP_NO_TLS\* options are deprecated$",
                 )
+                warnings.filterwarnings(
+                    "ignore",
+                    message=r"^`id` is deprecated, use `element_id` instead$",
+                )
```
FlorentinD, line 51: thats annoying but nothing we can do though :shrug:

graphdatascience/tests/integration/test_graph_ops.py
```diff
@@ -303,6 +303,14 @@ def test_graph_nodeProperties_stream_with_arrow(gds: GraphDataScience) -> None:
     y_values = result[result.nodeProperty == "y"]
     assert {e for e in y_values["propertyValue"]} == {2, 3, 4}
 
+    z_values = result[result.nodeProperty == "z"]
+    assert len(z_values) == 3
+    for e in z_values["propertyValue"]:
+        assert e in [[9], [42], [1337]]
```
FlorentinD, line 309: there is no easier assert for python? 

A bit spoiled by assertJ I guess.
But I would hope there is a test util for pandas series

graphdatascience/tests/integration/test_graph_ops.py
```diff
@@ -351,6 +367,14 @@ def test_graph_nodeProperties_stream_without_arrow(gds_without_arrow: GraphDataS
...
```
FlorentinD, line 373: could be worth to write our own test-util function for this if we use it this often

FlorentinD (changes requested):
Very nice feature, I think people will like it alot!

graphdatascience/graph/graph_entity_ops_runner.py
Mats-SX: A custom query could be something like

```
MATCH (n) WHERE id(n) IN $nodeIds
RETURN n.p1 AS p1, n.p2 AS p2, n.p3 AS p3
```

where `pi` are the names given by parameter

graphdatascience/tests/integration/test_graph_ops.py
```diff
@@ -317,10 +325,16 @@ def test_graph_streamNodeProperties_with_arrow_separate_property_columns(gds: Gr
 def test_graph_nodeProperties_stream_with_arrow_separate_property_columns(gds: GraphDataScience) -> None:
     G, _ = gds.graph.project(GRAPH_NAME, {"Node": {"properties": ["x", "y"]}}, "*")
 
-    result = gds.graph.nodeProperties.stream(G, ["x", "y"], separate_property_columns=True, concurrency=2)
-    assert list(result.keys()) == ["nodeId", "x", "y"]
+    result = gds.graph.nodeProperties.stream(
+        G, ["x", "y"], db_node_properties=["z", "name"], separate_property_columns=True, concurrency=2
```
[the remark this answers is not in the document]
brs96: Good point, that will lead to conflicting keys.
Added check to throw error early with a test, don't see any value for user to give duplicate keys at the moment...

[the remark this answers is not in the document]
FlorentinD: Agree, I also think they should use other names in this case :+1:

brs96:
Might leave the test cleanups to later work and add more helpers altogether, wdyt

FlorentinD (approved):

PR #257, opened 2023-01-23
Add generated doc example

nvitucci:
This PR adds:

- a script and supporting templates (based on `nbconvert` + Pandoc) to generate documents from notebooks;
- one generated doc (`load-data-via-graph-construction.adoc`) which replaces the existing one;
- generated docs for all the other notebooks.


nvitucci:
@mats @FlorentinD @adamnsch @brs96

I think we can review this focusing on the process and on the output rather than on the writing style (which I started drafting out, but it's easier to work on once the process is established).

I've added the instructions in the TESTING file as they seem to fit there for now, but we may decide to move them somewhere else.

TESTING.md
```diff
@@ -140,4 +140,24 @@ mypy .
 
 from the root. See `mypy.ini` for our custom mypy settings.
 
+## Generate documentation from Jupyter notebooks
 
+The documents in the `tutorials` section are automatically generated from the Jupyter notebooks.
+
+The `convert.sh` script in the `scripts/nb2doc/` directory can be used:
```
FlorentinD, line 147:
```suggestion
The script `scripts/nb2doc/convert.sh` can be used:
```

TESTING.md
```diff
@@ -140,4 +140,24 @@ mypy .
...
+
+* to generate documents from new notebooks;
+* to ensure that any changes to the existing notebooks are reflected in the existing documents (for instance in a CI setting).
+
+The script must be run from the project root directory and requires [Pandoc](https://pandoc.org/) to be already installed.
+
+```bash
+$ scripts/nb2doc/convert.sh
```
FlorentinD, line 155: I get errors for `Unknown option --markdown-headings.
Try pandoc --help for more information.`.

pandoc version 2.9.2.1 
What are you using? Maybe we should specify which version we expect this to work with

FlorentinD (commented):
I just tried out the script and had some problem. Maybe you have an idea?

TESTING.md
```diff
@@ -140,4 +140,24 @@ mypy .
...
 from the root. See `mypy.ini` for our custom mypy settings.
 
+## Generate documentation from Jupyter notebooks
 
+The documents in the `tutorials` section are automatically generated from the Jupyter notebooks.
+
+The `convert.sh` script in the `scripts/nb2doc/` directory can be used:
+
+* to generate documents from new notebooks;
+* to ensure that any changes to the existing notebooks are reflected in the existing documents (for instance in a CI setting).
```
FlorentinD, line 150: Would it be possible to run the script before building the documentation?
As in, we dont need to checkin the generated `adoc` files but only generate them on the fly

TESTING.md
FlorentinD: Ok, installing the latest version `3.0.1` works and produces the same output.

However there is some warnings - [NbConvertApp] Converting notebook examples/node-regression-with-subgraph-and-graph-sample.ipynb to asciidoc
``` MissingIDFieldWarning: Code cell is missing an id field, this will become a hard error in future nbformat versions. You may want to use `normalize()` on your notebooks before validations (available since nbformat 5.1.4). Previous versions of nbformat are fixing this issue transparently, and will stop doing so in the future.
  validate(nb)```

Comment (nvitucci) [reply TESTING.md:155]: I am using version 2.19.2 so I was thinking to add a note with `Pandoc >= 2.19`, but if you've tried Pandoc 3 I can take a look and make it conformant (so we can say `Pandoc >= 3.0`).

File: TESTING.md (hunk 1 above)
Comment (nvitucci) [reply TESTING.md:147]: I originally worded it like this and then changed my mind, but if this reads better I'll revert :)

File: TESTING.md (hunk 3 above)
Comment (FlorentinD) [reply TESTING.md:150]: Good point. Failing on a diff seems indeed be better then

File: scripts/nb2doc/asciidoc-template/index.adoc.j2 [hunk 4]
```diff
@@ -0,0 +1,94 @@
+// DO NOT EDIT - AsciiDoc file generated automatically
```
FlorentinD, line 1: Is this based on a template from somewhere?

Wonder if f.i. we use Latex even

FlorentinD (approved):
Great work, I like the generation!

The next step would be the CI check.
This can be done in a follow up PR though :) 
My naive thought would be to call the convert script and check git diff if its different

scripts/nb2doc/asciidoc-template/index.adoc.j2
nvitucci: Yes, this is based on the base Asciidoc template included in `nbconvert`. I've kept everything and only added to it or made changes were I saw fit, but we can customise it as we want.

nvitucci:
> The next step would be the CI check. This can be done in a follow up PR though :) My naive thought would be to call the convert script and check git diff if its different

That's what I'd have in mind too. We could try that first.

TESTING.md
nvitucci: Actually it seems that Pandoc 3+ is not yet supported by `nbconvert`:

`Your version must be at least (1.12.1) but less than (3.0.0).`

Running it with 3.0.1 anyway still works and does not produce any changes to the generated docs. We can still stick to the latest pre-3.x version (2.19.2) for the time being.

nvitucci: Indeed. For our purposes, when running in CI we'd still be able to use the Pandoc version we need and, in case, update it later.

PR #263, opened 2023-02-04
Adding Convenience Method to Drop Graph by Name

zach-blumenfeld:
When using the python client, there are times when you may not have a reference object for an existing graph that you previously created in a past run (this happens for me in experimentation, but I imagine it can happen anytime you are re-running an analysis/program). If you attempt to re-project a graph with the same name you will get an error and it can be non-trivial for new users to figure out how to get a reference to the previous graph and drop it.  Ultimately to  avoid the error you can do something like
```
if gds.graph.exists(PROJ_NAME)['exists']:
    gds.graph.get(PROJ_NAME).drop()
g, _ = gds.graph.project(PROJ_NAME, ....)
```
All this PR does is add a convenience method `gds.graph.drop_by_name(PROJ_NAME,...)` so you can do 
```
gds.graph.drop_by_name(PROJ_NAME,...)
g, _ = gds.graph.project(PROJ_NAME, ....)
```
Obviously, the convenience method could also be used any other time you want to drop a graph that you do not have a reference to. 

Ideally, there would be an argument like `override=True` in the project procedure.  Looking at how the client works though I could not find an elegant way to do that and realized that a better solution, if you wanted to go that route, would probably be to add the argument on the server-side.

### DIFF
--- a/graphdatascience/tests/integration/test_graph_ops.py
+++ b/graphdatascience/tests/integration/test_graph_ops.py
@@ -134,6 +134,16 @@ def test_graph_drop(gds: GraphDataScience) -> None:
         gds.graph.drop(G, True)
 
 
+def test_graph_drop_by_name(gds: GraphDataScience) -> None:
+    gds.graph.project(GRAPH_NAME, "*", "*")
+
+    result = gds.graph.drop_by_name(GRAPH_NAME, True)
+    assert result is not None
+    assert result["graphName"] == GRAPH_NAME
+    with pytest.raises(Exception):
+        gds.graph.drop_by_name(GRAPH_NAME, True)
+
+
 def test_graph_type_check(gds: GraphDataScience) -> None:
     G, _ = gds.graph.project(GRAPH_NAME, "*", "*")
 
--- a/graphdatascience/tests/integration/test_graph_ops.py
+++ b/graphdatascience/tests/integration/test_graph_ops.py
@@ -134,14 +134,15 @@ def test_graph_drop(gds: GraphDataScience) -> None:
         gds.graph.drop(G, True)
 
 
-def test_graph_drop_by_name(gds: GraphDataScience) -> None:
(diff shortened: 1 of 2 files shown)


Mats-SX:
You're correct that your code block to get and drop the graph, and then re-project it, works. But it seems unnecessary; you could just get it, not drop it and not re-project it:

```
if gds.graph.exists(PROJ_NAME)['exists']:
    g = gds.graph.get(PROJ_NAME).drop()
```

adamnsch:
Thank you for this contribution, the PR looks great. Would love to get more PRs from you in the future :)

I do agree with Mats however. It's true that it makes your use-case slightly more convenient perhaps, but we don't want people passing strings back and forth in their code I think.
We'd also have to document this function as a special case, and motivate why this particular function has the `name` option whereas others don't. Or add `gds.pageRank.stream_by_name(...)` as well? But I don't think that's a route we want to go down.

graphdatascience/tests/integration/test_graph_ops.py
```diff
@@ -134,6 +134,16 @@ def test_graph_drop(gds: GraphDataScience) -> None:
         gds.graph.drop(G, True)
 
 
+def test_graph_drop_by_name(gds: GraphDataScience) -> None:
+    gds.graph.project(GRAPH_NAME, "*", "*")
+
+    result = gds.graph.drop_by_name(GRAPH_NAME, True)
```
Mats-SX, line 140:
```suggestion
    result = gds.graph.get(GRAPH_NAME).drop(True)
```

I guess I don't see why the `drop_by_name` is better than `get().drop()`
If the problem is that `get()` will report an error if it does not find anything, I would be open to adding a boolean `failIfMissing` also to `get()`.

zach-blumenfeld: @Mats-SX yes, `get()` returning an error when not found would be the issue. I agree adding a `failIfMissing` parameter is a more elegant solution to this problem.

PR #264, opened 2023-02-07
Adding FailIfMissing Parameter for gds.graph.get()

zach-blumenfeld:
Adding `FailIfMissing` parameter to `get()` so you can drop graphs created in previous runs that you do not have a reference for. i.e.: `gds.graph.get(GRAPH_NAME, False).drop()` 

Based on discussion from #263 .

### DIFF
--- a/graphdatascience/graph/graph_proc_runner.py
+++ b/graphdatascience/graph/graph_proc_runner.py
@@ -153,9 +153,12 @@ class GraphProcRunner(CallerBase, UncallableNamespace, IllegalAttrChecker):
         return self._query_runner.run_query(query, params)
 
     @client_only_endpoint("gds.graph")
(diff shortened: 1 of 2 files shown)


graphdatascience/graph/graph_proc_runner.py
```diff
@@ -153,9 +153,12 @@ def list(self, G: Optional[Graph] = None) -> DataFrame:
         return self._query_runner.run_query(query, params)
 
     @client_only_endpoint("gds.graph")
-    def get(self, graph_name: str) -> Graph:
+    def get(self,
+            graph_name: str,
+            failIfMissing: bool = True,
```
Mats-SX, line 158: If this is `False` but there is no graph, we cannot return a Graph object. Instead, we should return `None`, I think.

PR #271, opened 2023-02-15
Update notebooks to get Neo4j DB credentials and URI from env

adamnsch:

Make sure:

### DIFF
--- a/examples/fastrp-and-knn.ipynb
+++ b/examples/fastrp-and-knn.ipynb
@@ -56,11 +56,18 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "import os\n",
     "from graphdatascience import GraphDataScience\n",
     "\n",
-    "# Override NEO4J_URI and NEO4J_AUTH here according to your setup\n",
-    "NEO4J_URI = \"bolt://localhost:7687\"\n",
+    "# Get Neo4j DB URI and credentials from environment if applicable\n",
+    "NEO4J_URI = os.environ.get(\"NEO4J_URI\", \"bolt://localhost:7687\")\n",
     "NEO4J_AUTH = None\n",
+    "if os.environ.get(\"NEO4J_USER\") and os.environ.get(\"NEO4J_PASSWORD\"):\n",
+    "    NEO4J_AUTH = (\n",
+    "        os.environ.get(\"NEO4J_USER\"),\n",
+    "        os.environ.get(\"NEO4J_PASSWORD\"),\n",
+    "    )\n",
+    "\n",
     "gds = GraphDataScience(NEO4J_URI, auth=NEO4J_AUTH)"
    ]
   },
@@ -379,27 +386,8 @@
   }
  ],
  "metadata": {
-  "kernelspec": {
-   "display_name": "Python 3 (ipykernel)",
-   "language": "python",
-   "name": "python3"
-  },
   "language_info": {
-   "codemirror_mode": {
-    "name": "ipython",
-    "version": 3
-   },
-   "file_extension": ".py",
-   "mimetype": "text/x-python",
-   "name": "python",
-   "nbconvert_exporter": "python",
-   "pygments_lexer": "ipython3",
-   "version": "3.8.10"
-  },
-  "vscode": {
-   "interpreter": {
-    "hash": "916dbcbb3f70747c44a77c7bcd40155683ae19c65e1c03b4aa3499c5328201f1"
-   }
+   "name": "python"
   }
  },
  "nbformat": 4,
--- a/examples/load-data-via-graph-construction.ipynb
+++ b/examples/load-data-via-graph-construction.ipynb
@@ -1,7 +1,6 @@
 {
  "cells": [
   {
-   "attachments": {},
    "cell_type": "markdown",
    "metadata": {
     "id": "KhH0v_bwUrXp"
@@ -21,7 +20,6 @@
    ]
   },
   {
-   "attachments": {},
    "cell_type": "markdown",
    "metadata": {},
    "source": [
@@ -31,7 +29,6 @@
    ]
   },
   {
-   "attachments": {},
    "cell_type": "markdown",
    "metadata": {
     "id": "X5ttqidnWPTC"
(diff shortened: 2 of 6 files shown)


examples/fastrp-and-knn.ipynb
```diff
@@ -379,27 +386,8 @@
...
-   "mimetype": "text/x-python",
-   "name": "python",
-   "nbconvert_exporter": "python",
-   "pygments_lexer": "ipython3",
-   "version": "3.8.10"
-  },
-  "vscode": {
-   "interpreter": {
-    "hash": "916dbcbb3f70747c44a77c7bcd40155683ae19c65e1c03b4aa3499c5328201f1"
-   }
```
FlorentinD, line 402: nice you could get rid of these. Is there a way to avoid this in the future.
I think vs code added these automatically - e.g. what did you run to clean it?

examples/load-data-via-graph-construction.ipynb
```diff
@@ -47,20 +44,19 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {
-    "id": "izJD_HlcAdQ0"
```
FlorentinD, line 51: the id is something that annoyed me recently as well. I think @nvitucci  said it can be useful if you want to reference a specifc cell though

examples/load-data-via-graph-construction.ipynb
```diff
@@ -235,92 +220,9 @@
...
-       "</div>"
-      ],
-      "text/plain": [
-       "    nodeId labels  subject                                           features\n",
-       "0    31336  Paper        0  [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, ...\n",
-       "1  1061127  Paper        1  [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, ...\n",
-       "2  1106406  Paper        2  [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, ...\n",
-       "3    13195  Paper        2  [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, ...\n",
-       "4    37879  Paper        3  [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, ..."
-      ]
```
FlorentinD, line 317: these outputs were considered useful in the past to give viewers a hint on how it will look.
Not clearly opposed to removing it, but was it needed for the testing?

FlorentinD (commented):
Nice we could remove some arbitrary metadata.

Mostly wonder about the removed outputs and the id field.
Also how we can make sure the notebooks stay clean in the future / what needs to be run to get them clean.

Further I would like to let @nvitucci the chance to give his opinion, although that can also happen async after merging this PR

examples/fastrp-and-knn.ipynb
adamnsch: There's a tool called `jupyter nbconvert` that one can use. Could be nice to add this to `makestyle`/`checkstyle` actually

examples/load-data-via-graph-construction.ipynb
adamnsch: The testing doesn't do anything with the output atm. And even if it did, it would probably use output it produces when running the notebookes, instead of this cached output?

examples/fastrp-and-knn.ipynb
FlorentinD: Thats also what nicola uses to convert them to docs :+1:

examples/load-data-via-graph-construction.ipynb
FlorentinD: I would personally lean towards no outputs as it makes our lives easier (see our GDS docs where we always fight with outputs).

Just on a note, we might want to add more asserts in the notebooks to make sure they dont change between versions

FlorentinD (approved):
I think we can merge this as it is and discuss the general style we strive for in notebooks a dedicated meeting/chat.

examples/fastrp-and-knn.ipynb
adamnsch: ah ok, cool :)

adamnsch: I can make a PR for that as well then

examples/load-data-via-graph-construction.ipynb
adamnsch: > I would personally lean towards no outputs as it makes our lives easier (see our GDS docs where we always fight with outputs).

:+1: 

> Just on a note, we might want to add more asserts in the notebooks to make sure they dont change between versions

Yeah, I think we should have a work item for improving the notebook testing

adamnsch:
> I think we can merge this as it is and discuss the general style we strive for in notebooks a dedicated meeting/chat.

Sounds good :+1:

examples/fastrp-and-knn.ipynb
nvitucci: Yes, this depends on the tools we use to created/edit notebooks. I don't know if they could be useful for something, but at the moment I am not against scrubbing them via styling if possible. If a tool reintroduces them, we should look into how to configure the tool as well (for instance there should be such an option in VSCode).

examples/load-data-via-graph-construction.ipynb
nvitucci: As you know I'm keen to keep the results from a docs reading perspective _and_ a (yet not completely defined) testing perspective, but if there is a strong opinion against keeping them, we can discard them.

I see two alternatives that we can try out:

- introduce `assert`s in the notebooks and use them for both testing and docs purposes
- use metadata to mark the output cells that we want to keep, so that the notebook is not polluted by unnecessary results but still contains output in "strategic" (documentation-oriented and/or test-oriented) places

I'll give it a try to both and raise a separate discussion.

examples/load-data-via-graph-construction.ipynb
nvitucci: The cell IDs have been introduced in [nbformat 4.5](https://nbformat.readthedocs.io/en/latest/format_description.html#cell-ids). We could get rid of them, but this might create issues in the future; it did create an error in the conversion process that I cannot find again ATM.

I'm wondering if it would make sense to just have them added at once to all notebooks and update the `nbformat` metadata.

PR #272, opened 2023-02-16
Add tooling for checking and formatting notebooks

adamnsch:
Thank you for your contribution to the Graph Data Science Client project.

Before submitting this PR, please read [Contributing to the Neo4j Ecosystem](https://github.com/neo4j/graph-data-science-client/blob/main/CONTRIBUTING.md). 

Make sure:
- [x] You signed the [Neo4j CLA](https://neo4j.com/developer/cla/#sign-cla) (Contributor License Agreement) so that we are allowed to ship your code in our library
- [x] Your contribution is covered by tests

PR #274, opened 2023-02-16
Update generated docs + Add tutorials page link to docs

FlorentinD:
- Improve title for FastRP + KNN example
- Update generated docs
- Link to generated notebook docs

Make sure:

### DIFF
--- a/examples/fastrp-and-knn.ipynb
+++ b/examples/fastrp-and-knn.ipynb
@@ -5,7 +5,7 @@
    "id": "5cb64c55",
    "metadata": {},
    "source": [
-    "# FastRP and kNN end-to-end example\n",
+    "# Product recommendations with kNN based on FastRP embeddings\n",
     "\n",
     "This notebook exemplifies how to use the `graphdatascience` Python library to operate Neo4j GDS.\n",
     "It shows an adapted version of the FastRP and kNN end-to-end example from the GDS Manual, found [here](https://neo4j.com/docs/graph-data-science/current/end-to-end-examples/fastrp-knn-example).\n",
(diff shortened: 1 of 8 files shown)


examples/fastrp-and-knn.ipynb
```diff
@@ -386,8 +386,19 @@
...
+   "name": "python3"
+  },
   "language_info": {
-   "name": "python"
+   "name": "python",
+   "version": "3.10.6"
+  },
+  "vscode": {
+   "interpreter": {
+    "hash": "916dbcbb3f70747c44a77c7bcd40155683ae19c65e1c03b4aa3499c5328201f1"
```
FlorentinD, line 400: these got added automatically :/

adamnsch (approved):
Great :) I'll not forget about this next time!

Btw, we added links to the notebooks in our GDS server manual recently, eg: https://neo4j.com/docs/graph-data-science/current/machine-learning/node-embeddings/fastrp/

Should we update those titles as well? And should they point to the client docs now instead of the notebooks?

FlorentinD:
We should have tooling to check these are in sync. When talking with Nicola, I think he wanted to give it a try as well.
The tooling is likely quite similar to what you added in #272

I would lean towards linking to the client docs there as well :thinking:

adamnsch:
> We should have tooling to check these are in sync. When talking with Nicola, I think he wanted to give it a try as well. The tooling is likely quite similar to what you added in #272
> 
> I would lean towards linking to the client docs there as well thinking But we first need to publish the client docs, maybe after the 1.6 cherry pick is live.

Ah, that's a good idea. Should be easy to set up

PR #287, opened 2023-02-23
Retry Neo4j queries that fail with retryable errors

adamnsch:

Make sure:

### DIFF
--- a/graphdatascience/query_runner/neo4j_query_runner.py
+++ b/graphdatascience/query_runner/neo4j_query_runner.py
@@ -1,4 +1,5 @@
 import re
+import time
 import warnings
 from concurrent.futures import Future, ThreadPoolExecutor, wait
 from typing import Any, Dict, List, Optional
@@ -18,6 +19,7 @@ from .query_runner import QueryRunner
 class Neo4jQueryRunner(QueryRunner):
     _LOG_POLLING_INTERVAL = 0.5
     _NEO4J_DRIVER_VERSION = ServerVersion.from_string(neo4j.__version__)
+    _MAX_NUM_RETRYS = 10
 
     def __init__(
         self, driver: neo4j.Driver, database: Optional[str] = neo4j.DEFAULT_DATABASE, auto_close: bool = False
@@ -56,13 +58,22 @@ class Neo4jQueryRunner(QueryRunner):
                     message=r"^`id` is deprecated, use `element_id` instead$",
                 )
 
-            try:
-                result = session.run(query, params)
-            except Exception as e:
-                if custom_error:
-                    self.handle_driver_exception(session, e)
-                else:
-                    raise e
+            trys = 1
+            while trys < self._MAX_NUM_RETRYS:
+                try:
+                    result = session.run(query, params)
+                    break
+                except neo4j.exceptions.Neo4jError as e:
+                    trys += 1
+
+                    if self._error_retryable(e):
+                        time.sleep(0.5)
+                        continue
+
+                    if custom_error:
+                        self.handle_driver_exception(session, e)
+                    else:
+                        raise e
 
             # Though pandas support may be experimental in the `neo4j` package, it should always
             # be supported in the `graphdatascience` package.
@@ -72,6 +83,12 @@ class Neo4jQueryRunner(QueryRunner):
             )
 
             return result.to_df()
+
+    def _error_retryable(self, e: neo4j.exceptions.Neo4jError) -> bool:
+        if self._NEO4J_DRIVER_VERSION.major > 4:
+            return e.is_retryable()
+        else:
+            return e.is_retriable()
 
     def run_query_with_logging(
         self, query: str, params: Optional[Dict[str, Any]] = None, database: Optional[str] = None
--- a/changelog.md
+++ b/changelog.md
@@ -16,6 +16,7 @@
 
 * When an almost correct method is called, raise an error with a message that suggests the most probable correct method name that was intended.
 * Improved IDE auto-completion support to give significantly fewer false positive suggestions.
+* Retry underlying Cypher queries that fail with retryable errors.
 
 
 ## Other changes


graphdatascience/query_runner/neo4j_query_runner.py
```diff
@@ -73,6 +84,12 @@ def run_query(
 
             return result.to_df()
 
+    def _error_retryable(self, e: neo4j.exceptions.Neo4jError) -> bool:
+        if self._NEO4J_DRIVER_VERSION.major > 4:
+            return e.is_retryable()
+        else:
+            return e.is_retriable()
```
FlorentinD, line 91: i wonder if this does not also include some exception cases where the query ran half way.

adamnsch: https://neo4j.com/docs/api/python-driver/current/api.html#neo4j.exceptions.Neo4jError.is_retryable maybe :man_shrugging:

robsdedude: That depends on the query. A `CALL { ... } IN TRANSACTION` query for instance might have already been partially committed. Also some procedures (`CALL foobar()`) might change state on the server before the final commit.

Other than that, since you are only catching `neo4j.exceptions.Neo4jError` (and not also `neo4j.exceptions.DriverError`) you are never handling dropped connections. Which in this case is good. Because with `session.run` you get no guarantee as to when the underlying transaction, which is managed by the server, gets committed other than that it will definitely have been committed after you've successfully streamed all records. So when you loose connectivity, you can't easily know if the query has been committed yet or not.

graphdatascience/query_runner/neo4j_query_runner.py
```diff
@@ -56,13 +58,22 @@ def run_query(
...
-                result = session.run(query, params)
-            except Exception as e:
-                if custom_error:
-                    self.handle_driver_exception(session, e)
-                else:
-                    raise e
+            trys = 1
+            while trys < self._MAX_NUM_RETRYS:
+                try:
+                    result = session.run(query, params)
```
robsdedude, line 64: Line 85
```python
result.to_df()
```

should most likely also be in the try block. As the result cursor is a stream an error could still be returned later down the road. So you should make sure to fully consume the result before you can assume that the transaction has been committed successfully.

adamnsch: Didn't know that :O makes sense though!

graphdatascience/query_runner/neo4j_query_runner.py
adamnsch: Thank you for this input @robsdedude :) I have some questions on how to proceed:


> Also some procedures (CALL foobar()) might change state on the server before the final commit.

Are these some kind of special queries? Or does it include regular GDS procedures? If so, could there be problems in calling `session.run` multiple times (upon receiving retryable errors) like we do here?


After learning a little bit about `session.run`, and reading your feedback, it does seem like this patch would improve things for the user since it would retry failed queries, which `session.run` doesn't do on its own. However I also read that managed transactions does this by default, and that "Auto-commit transactions should only be used when [transaction functions](https://neo4j.com/docs/python-manual/current/transactions/#managed-transactions) do not fit the purpose, or for quick prototyping." Would you suggest we start using managed transactions instead of `session.run`?


> Other than that, since you are only catching neo4j.exceptions.Neo4jError (and not also neo4j.exceptions.DriverError) you are never handling dropped connections.

One of the things we would like to address with this work is to be able to handle changes in the routing table which triggers loss of connections afaiu (please correct me if I'm wrong @Mats-SX @FlorentinD). If this method is not addressing that, then how would we go about it? If it's even possible.


Thanks in advance :)

robsdedude: > Are these some kind of special queries? Or does it include regular GDS procedures?

I don't know what your GDS procedures do internally. The key point for when to use `session.run` instead of transaction functions is when you cannot (reliably) roll back a transaction. Transaction functions are fine for:
 * idempotent queries (e.g., read-only queries)
 * queries that can be rolled back

> One of the things we would like to address with this work is to be able to handle changes in the routing table which triggers loss of connections. If this method is not addressing that, then how would we go about it? If it's even possible.

Yes, a change in the RT can cause a connectivity error (note that routing is only ever enabled when the driver is configured with a `neo4j` scheme URI, not sure if this applies to the GDS use-case). The problem is that session.run has no explicit commit. So when you loose connectivity you don't know it the query has been persisted or not. At least not without using some extra queries or other logic to check.

adamnsch: > I don't know what your GDS procedures do internally. The key point for when to use session.run instead of transaction functions is when you cannot (reliably) roll back a transaction. Transaction functions are fine for:

> idempotent queries (e.g., read-only queries)
> queries that can be rolled back

Thanks for the information :) I guess I meant more generally: what trait(s) of a procedure would cause it to possibly change state on the server before a commit?
Apart from this consideration, it does seem like `session.run` is fine for our use-case according to your description here. We should add retries though (like in this PR).

> Yes, a change in the RT can cause a connectivity error (note that routing is only ever enabled when the driver is configured with a neo4j scheme URI, not sure if this applies to the GDS use-case).

Ok. We've had a request to specifically handle the RT connectivity issue, so I guess it is a use-case for us.

>  The problem is that session.run has no explicit commit. So when you loose connectivity you don't know it the query has been persisted or not. At least not without using some extra queries or other logic to check.

I see. So if we were to use managed transactions, we'd know if the server has received the query before any kind of error is raised on the client side?
Supposing that we know that the query failed without being persisted on the server (`tx.commit` failed would be a symptom?) due to an RT change, how do we retry it with the new RT? Do we simply do a new `commit`, a new session, or something else? I do read that "The driver receives a routing table from the server upon successful connection" in the driver docs, but I don't know exactly what that implies for our path forward here. Somewhere else it does say "sessions borrow connections from the connection pool as needed" which leads me to believe that creating a new session is not enough, if/since the connections in the pool are stale anyway

Mats-SX: >what trait(s) of a procedure would cause it to possibly change state on the server before a commit?
For example, `gds.graph.project` changes the state on the server by registering the graph in the catalog, and that can happen even if the procedure somehow breaks before it fully completes its roundtrip back to the client, like if someone terminates its transaction right between registering the graph in the catalog and returning from the procedure method.

Similarly, to @robsdedude's other point, `CALL { ... } IN TRANSACTION` is akin to our `.write` modes: if a `.write` proc is terminated, some of the state has already been committed.

Mats-SX: To elaborate on my first point: if the `gds.graph.project` proc fails _after_ having registered a graph in the graph catalog, we will _always_ fail if we retry it. So retrying is not always safe.

This is not the case for all GDS procs. Some are definitely read-only, like `.stream`, `.stats`, `.estimate` procs. They are always safe to retry.

adamnsch: Great explanations, thanks you @Mats-SX :pray:

PR #314, opened 2023-03-07
Verify connection before making a Neo4j query

adamnsch:
And retry for up to 10 minutes.

I have verified that it works when resizing an AuraDS instance, which is the use-case we're targeting.

Make sure:

### DIFF
--- a/graphdatascience/query_runner/neo4j_query_runner.py
+++ b/graphdatascience/query_runner/neo4j_query_runner.py
@@ -1,4 +1,6 @@
+import logging
 import re
+import time
 import warnings
 from concurrent.futures import Future, ThreadPoolExecutor, wait
 from typing import Any, Dict, List, Optional
@@ -9,6 +11,7 @@ from pandas import DataFrame
 from tqdm.auto import tqdm
 
 from ..error.endpoint_suggester import generate_suggestive_error_message
+from ..error.unable_to_connect import UnableToConnectError
 from ..server_version.server_version import ServerVersion
 from .cypher_graph_constructor import CypherGraphConstructor
 from .graph_constructor import GraphConstructor
@@ -25,6 +28,7 @@ class Neo4jQueryRunner(QueryRunner):
         self._driver = driver
         self._auto_close = auto_close
         self._database = database
+        self._logger = logging.getLogger()
 
     def run_query(
         self,
@@ -38,6 +42,8 @@ class Neo4jQueryRunner(QueryRunner):
 
         if database is None:
             database = self._database
+
+        self._verify_connectivity()
 
         with self._driver.session(database=database) as session:
             if self._NEO4J_DRIVER_VERSION == ServerVersion(4, 4, 6):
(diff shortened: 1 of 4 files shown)


graphdatascience/query_runner/neo4j_query_runner.py
```diff
@@ -174,3 +180,33 @@ def handle_driver_exception(session: neo4j.Session, e: Exception) -> None:
...
+                        message=r"^The configuration may change in the future.$",
+                    )
+                self._driver.verify_connectivity()
+                break
+            except neo4j.exceptions.DriverError as e:
+                exception = e
+                if retrys % WARN_INTERVAL == 0:
+                    self._logger.warn("Unable to connect to the Neo4j DBMS. Trying again...")
+
+                time.sleep(WAIT_TIME)
```
FlorentinD, line 206: waiting a second sounds like a good timeout :+1:

graphdatascience/tests/integration/test_database_ops.py
```diff
@@ -94,28 +93,27 @@ def test_aurads_rejects_bolt() -> None:
         ValueError,
         match=r"AuraDS requires using the 'neo4j\+s' protocol \('bolt' was provided\)",
     ):
-        GraphDataScience("bolt://localhost:7687", auth=AUTH, aura_ds=True)
+        GraphDataScience._configure_aura("bolt://localhost:7687", {})
```
FlorentinD, line 96: cool that we can reuse this method inside tests now :+1:

graphdatascience/query_runner/neo4j_query_runner.py
```diff
@@ -174,3 +180,33 @@ def handle_driver_exception(session: neo4j.Session, e: Exception) -> None:
         all_endpoints = list_result.to_df()["name"].tolist()
 
         raise SyntaxError(generate_suggestive_error_message(requested_endpoint, all_endpoints)) from e
+
+    def _verify_connectivity(self) -> None:
+        WAIT_TIME = 1
+        MAX_RETRYS = 10 * 60
```
FlorentinD, line 186: wonder if we want to allow users to configure this timeout

FlorentinD (approved):
Nice change!

Mainly wonder if we want to expose the timeout parameter to the user.

graphdatascience/query_runner/neo4j_query_runner.py
adamnsch: Yes, I thought about that too. My reasoning for not doing it, was to wait for feedback to see if anyone actually wants it. Don't want to complicate the surface needlessly. But it would be very easy to add of course. Wdyt?

FlorentinD: Waiting for feedback sounds good. We could also ask Dave directly what his opinion is (or the gds power users).

adamnsch: Let's wait then :+1:

PR #322, opened 2023-03-31
Add NetworkX graph loading functionality

adamnsch:

Make sure:


FlorentinD:
Do you know why the docs preview build did not see any docs changes? :thinking:

adamnsch:
> Do you know why the docs preview build did not see any docs changes? thinking

Strange. Maybe something we need to report to Neil

adamnsch:
Requires https://github.com/neo-technology/graph-analytics/pull/7375 for the testing to actually run

doc/modules/ROOT/pages/graph-object.adoc
```diff
@@ -161,6 +162,129 @@ If multiple node dataframes are used, they need to contain distinct node ids acr
...
+Either all edges of the `networkx.Graph` must have valid `relationshipType` attributes, or they can be _completely_ left out from the graph.
+That is, a `networx.Graph` with no `relationshipType` edge attributes at all is also allowed.
+In this latter case, the relationships in the projected `Graph` will all have the relationship type `R`.
+
+
+==== Relationship properties
+
+Relationship properties for the projected GDS graph are taken from attributes on `networkx.Graph` edges.
+The keys of the attributes will map to property names, and the allowed values must follow the regular guidelines for relationship property values in GDS.
+That includes the limitation of having at most one property per relationship.
```
FlorentinD, line 275: This is not a limitation in GDS, a relationships can have more than one property projected.
However for algorithms one property needs to be selected.

Maybe we want to ease up on this restrictions?

graphdatascience/graph/nx_loader.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+class NXLoader(UncallableNamespace, IllegalAttrChecker):
+    @client_only_endpoint("gds.graph.networkx")
+    @compatible_with("load", min_inclusive=ServerVersion(2, 1, 0))
+    def load(self, nx_G: nx.Graph, graph_name: str, concurrency: int = 4) -> Graph:
+        nodes, rels = self._parse(nx_G)
+
+        undirected_rel_types = []
+        if not isinstance(nx_G, nx.DiGraph) and not isinstance(nx_G, nx.MultiDiGraph):
```
FlorentinD, line 22:
```suggestion
        if nx_G.to_directed_class():
```
reading at https://networkx.org/documentation/stable/reference/classes/multigraph.html this check would allow us to handle also custom implementations of the NetworkX API. (should be None if its directed)

graphdatascience/graph/nx_loader.py
```diff
@@ -0,0 +1,124 @@
...
+        node_props_schema: Dict[Tuple[str, ...], Set[str]] = {}
+        no_node_labels = None
+
+        for node_id, attrs in nx_G.nodes(data=True):
+            labels_attr = attrs.get("labels", None)
+            if no_node_labels is None:
+                no_node_labels = labels_attr in [None, []]
+
+            if (labels_attr in [None, []]) is not no_node_labels:
+                raise ValueError("Some but not all nodes have a 'labels' attribute")
```
FlorentinD, line 42: Do we want to report the node_id we encountered this at?
```suggestion
                raise ValueError(f"Expected a 'labels' attribute for node with id {node_id} as etiher all or none can have labels")
```

graphdatascience/graph/nx_loader.py
```diff
@@ -0,0 +1,124 @@
...
+        node_dicts_by_labels: Dict[Tuple[str, ...], Dict[str, List[Any]]] = defaultdict(lambda: defaultdict(list))
+        node_props_schema: Dict[Tuple[str, ...], Set[str]] = {}
+        no_node_labels = None
+
+        for node_id, attrs in nx_G.nodes(data=True):
+            labels_attr = attrs.get("labels", None)
+            if no_node_labels is None:
+                no_node_labels = labels_attr in [None, []]
+
+            if (labels_attr in [None, []]) is not no_node_labels:
```
FlorentinD, line 41: Why do we have this limitation couldnt we handle this in the projections? 
Just checked and you can use `[] or null`

graphdatascience/graph/nx_loader.py
```diff
@@ -0,0 +1,124 @@
...
+                raise ValueError("Some but not all nodes have a 'labels' attribute")
+
+            node_labels: List[str]
+            if no_node_labels:
+                node_labels = ["N"]
```
FlorentinD, line 46: This seems biased.
Why dont we just use `None` if there are none?

graphdatascience/graph/nx_loader.py
```diff
@@ -0,0 +1,124 @@
...
+            labels_key: Tuple[str, ...] = tuple(sorted(node_labels))
+
+            node_dicts_by_labels[labels_key]["nodeId"].append(node_id)
+
+            props = {prop for prop in attrs.keys() if prop != "labels"}
+
+            if labels_key in node_props_schema:
+                if props != node_props_schema[labels_key]:
+                    raise ValueError(
+                        f"Not all nodes with labels {node_labels} have the properties: "
```
FlorentinD, line 66: I dont get why we check the property schema here. Could we let this be handled on the gds side where we use the default default values, i.e., `[]` and `Double.NaN` and so on?

FlorentinD: Ah I see, we create dfs per node label to avoid many default values for some properties.
How about going best effort than though?
I could see us warning if we want to check that each node with a label has certain node labels.
But throwing an error seems stricter than needed

graphdatascience/graph/nx_loader.py
```diff
@@ -0,0 +1,124 @@
...
+
+            for prop in props:
+                node_dicts_by_labels[labels_key][prop].append(attrs[prop])
+
+        nodes = [
+            DataFrame({"labels": [list(labels)] * len(attrs["nodeId"]), **attrs})
+            for labels, attrs in node_dicts_by_labels.items()
+        ]
+
+        rel_dicts_by_types: Dict[str, Dict[str, List[Any]]] = defaultdict(lambda: defaultdict(list))
```
FlorentinD, line 80: for the rel dfs the same comments as for node dfs apply

graphdatascience/tests/integration/test_nx_loader.py
```diff
@@ -0,0 +1,110 @@
...
+
+    assert G.relationship_count() == 6
+    assert G.relationship_types() == ["R"]
+    assert G.relationship_properties("R") == ["weight"]
+
+    G.drop()
+
+
+@pytest.mark.compatible_with(min_inclusive=ServerVersion(2, 1, 0))
+def test_undirected_nx_without_Arrow(gds_without_arrow: GraphDataScience, undirected_nx_G: nx.Graph) -> None:
```
FlorentinD, line 60: arguably we only need to test the non arrow option here as the arrow vs non-arrow construct is tested at another place

graphdatascience/tests/unit/test_nx_loader.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+def test_parse_illegal_node_labels(gds: GraphDataScience) -> None:
+    nx_G = nx.DiGraph()
+    nx_G.add_node(42, labels=1337)
+
+    with pytest.raises(
+        ValueError,
+        match=re.escape(
+            "`labels` node attributes must be of type `str` or `list[str]`. Given `labels`: 1337 for node with id 42"
```
FlorentinD, line 163: Thats a nice error message :)

FlorentinD (changes requested):
Very exciting feature.

I am mostly surprised by the restrictions we have for loading from networkx

doc/modules/ROOT/pages/graph-object.adoc
```diff
@@ -161,6 +162,129 @@ If multiple node dataframes are used, they need to contain distinct node ids acr
...
+# Verify that the projection is what we expect
+assert G.name() == "purchases"
+assert G.node_count() == 2
+assert set(G.node_labels()) == {"Person", "Product", "Item"}
+assert G.node_properties("Person") == ["age"]
+assert G.node_properties("Product") == ["cost"]
+# Count rel not being = 2 indicates the graph is indeed directed
+assert G.relationship_count() == 1
+assert G.relationship_types() == ["BUYS"]
+assert G.relationship_properties("BUYS") == ["quantity"]
```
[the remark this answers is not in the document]
adamnsch: This example is to demonstrate the how the schema translates. I don't want to stream back the graph topology, node props, rel props, that would for sure be more extensive. But we can remove some of these assert if you find them redundant?

graphdatascience/graph/nx_loader.py
adamnsch: Cool! Will check this out

graphdatascience/graph/nx_loader.py
adamnsch: We considered this of course, but it's kind of arbitrary what nodes comes first. I.e. a node with labels might come before, and thus setting the schema. I'd rather keep it like it is

graphdatascience/graph/nx_loader.py
adamnsch: We could do that, but then you end up with a projection that's very hard to use. We did consider this. We thought it better to be string upfront than having FastRP fail later because some types are NaN/null or similar. But I don't know

graphdatascience/graph/nx_loader.py
adamnsch: I get errors if I try to load with node labels `[]` or `null` using Arrow as the backend

graphdatascience/graph/nx_loader.py
adamnsch: Getting errors as mentioned. Also then we don't get the ugly node label `__ALL__` in the schema which is nice

graphdatascience/graph/nx_loader.py
FlorentinD: ah .. not sure if its expected :open_mouth:

graphdatascience/tests/integration/test_nx_loader.py
adamnsch: True, but it's so cheap :)

graphdatascience/graph/nx_loader.py
adamnsch: Done!

doc/modules/ROOT/pages/graph-object.adoc
```diff
@@ -161,6 +162,129 @@ If multiple node dataframes are used, they need to contain distinct node ids acr
...
+
+[TIP]
+====
+Combined with https://networkx.org/documentation/stable/reference/readwrite/index.html[NetworkX's functionality for reading various graph formats] one can easily load popular graph formats like edge list and GML into GDS.
+====
+
+[TIP]
+====
+Combined with https://networkx.org/documentation/stable/reference/generators.html[NetworkX's functionality for generating various kinds of graphs] one can easily load graph types popular in the literature into GDS, such as expanders, lollipop graphs, complete graphs, and more.
+====
```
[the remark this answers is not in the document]
nvitucci: Yup, great idea! And if @adamnsch is already doing work on it, I'll be happy to review anyway :)

graphdatascience/graph/nx_loader.py
FlorentinD: https://trello.com/c/gv7FxQO2/8047-allow-empty-node-labels-when-projecting-through-arrow created a card

adamnsch: thank you! :pray:

graphdatascience/graph/nx_loader.py
adamnsch: It also makes it more consistent since we need type "R" for rels to allow for undirectedness

graphdatascience/graph/nx_loader.py
adamnsch: @FlorentinD Do we want a card for this feedback as well?

graphdatascience/graph/nx_loader.py
adamnsch: Keeping it as then for now

graphdatascience/graph/nx_loader.py
adamnsch: If it's directed it will return itself, it won't be None. But yes, probably the safest check would be:
```
if not isinstance(nx_G.to_directed_class(), type(nx_G)):
```
If they haven't overloaded the method `to_directed_class` in their custom graph, it will default to returning something like `nx.DiGraph`, but that's their problem and really nothing we can do I think

doc/modules/ROOT/pages/graph-object.adoc
adamnsch: Fixed

graphdatascience/graph/nx_loader.py
adamnsch: done

graphdatascience/graph/nx_loader.py
adamnsch: Let's keep this as is for now then?

graphdatascience/graph/nx_loader.py
adamnsch: Let's keep this as is for now then?

FlorentinD (approved):

PR #341, opened 2023-04-14
Add sphinx setup and docstring

Mats-SX:


brs96:
Adds all sphinx doc and setup. No linking with asciidoc and publishing yet.

doc/sphinx/README.adoc
```diff
@@ -0,0 +1,19 @@
+= Building GDS client ref docs with sphinx
+
+Doc sources exist in code source files.
+Edit it there.
+
+To build the html content, do this:
+
+1. Set working directory to the root of this repository.
+2. Install sphinx into your venv: `pip install sphinx`
+3. Build the HTML: `(cd doc/sphinx && make html)`
```
FlorentinD, line 10: Errors
```
/home/florentin/repos/gdsclient/doc/sphinx/source/graph.rst:15: WARNING: duplicate object description of gds.alpha.graph.sample.rwr, other instance in algorithms, use :noindex: for one of them
/home/florentin/repos/gdsclient/doc/sphinx/source/misc.rst:53: ERROR: Unknown directive type "py:fuction".

.. py:fuction:: gds.server_version(self) -> ServerVersion:

    Return the installed graph data science library version.
    ```

File: doc/sphinx/Makefile [hunk 2]
```diff
@@ -0,0 +1,20 @@
+# Minimal makefile for Sphinx documentation
```
FlorentinD, line 1: why did you use a MakeFile? (Did you write it yourself/ did you follow a tutorial?

doc/sphinx/make.bat
```diff
@@ -0,0 +1,35 @@
+@ECHO OFF
```
FlorentinD, line 1: Is this from a template?

doc/sphinx/source/conf.py
```diff
@@ -0,0 +1,38 @@
...
+# For the full list of built-in configuration values, see the documentation:
+# https://www.sphinx-doc.org/en/master/usage/configuration.html
+import os
+import sys
+
+# -- Project information -----------------------------------------------------
+# https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information
+
+project = "Neo4j Graph Data Science Python Client"
+copyright = "2023, Neo4j"
```
FlorentinD, line 12: is it `Neo4j Inc` nowadays?
Although our Java copyright header has `Neo4j` or `Neo4j Sweden AB` in it :shrug:

graphdatascience/error/client_only_endpoint.py
```diff
@@ -10,6 +10,7 @@ def client_only_endpoint(expected_namespace_prefix: str) -> Callable[[F], F]:
     def decorator(func: F) -> F:
         wraps(func)
 
+        # @wraps(func)
```
adamnsch, line 13: This should not be removed I think :O Did it mess up the docs?

doc/sphinx/source/conf.py
```diff
@@ -0,0 +1,38 @@
...
+author = "Neo4j"
+release = "2023"
+
```
FlorentinD, line 15: seeing the options. How do we version this?
Likely we should have the same as the client.

If there are some newer endpoints (new server version), we should update this listing and ideally say for which version they are available :thinking: 

Thinking further - major server version will lead to a server major client release to make these docs valid.
We should talk about this aside from this PR on what we want to do here. Also regarding generating the algo endpoints (not only the docs).

doc/sphinx/source/conf.py
```diff
@@ -0,0 +1,38 @@
...
+# -- General configuration ---------------------------------------------------
+# https://www.sphinx-doc.org/en/master/usage/configuration.html#general-configuration
+
+# Add the root of the project to the path so that Sphinx can find the Python sources
+sys.path.append(os.path.join(os.path.dirname(__file__), "..", "..", ".."))
+
+extensions = [
+    "sphinx.ext.autodoc",
```
FlorentinD, line 23:
```suggestion
    "sphinx.ext.autodoc", # include docs from docstrings
```

doc/sphinx/source/conf.py
```diff
@@ -0,0 +1,38 @@
...
+    "sphinx.ext.napoleon",
```
FlorentinD, line 24:
```suggestion
    "sphinx.ext.napoleon", # Support for NumPy and Google style docstrings
```

doc/sphinx/source/conf.py
```diff
@@ -0,0 +1,38 @@
...
+    # 'sphinx.ext.autosummary',
```
FlorentinD, line 25: How did it go with the autosummary?

doc/sphinx/source/conf.py
```diff
@@ -0,0 +1,38 @@
...
+]
+
+# napoleon_use_rtype = False
```
FlorentinD, line 28: looking at the return types did you try to not just return `DataFrame` but include the package name `pandas.DataFrame` or something?

doc/sphinx/source/conf.py
```diff
@@ -0,0 +1,38 @@
...
+
+templates_path = ["_templates"]
+exclude_patterns = []  # type: ignore
+
+
+# -- Options for HTML output -------------------------------------------------
+# https://www.sphinx-doc.org/en/master/usage/configuration.html#options-for-html-output
+
+html_theme = "alabaster"
```
FlorentinD, line 37: We like its simplicity.
 
General question -would it make sense to make a neo4j theme? (mainly for @nvitucci )

doc/sphinx/source/graph.rst
```diff
@@ -0,0 +1,185 @@
+Graph procedures
```
FlorentinD, line 1: also generated?

doc/sphinx/source/misc.rst
```diff
@@ -0,0 +1,59 @@
+Miscellaneous procedures
```
FlorentinD, line 1: guess this one is also generated?

graphdatascience/graph/graph_object.py
```diff
@@ -41,24 +49,59 @@ def _graph_info(self, yields: List[str] = []) -> "Series[Any]":
         return info.squeeze()  # type: ignore
 
     def database(self) -> str:
+        """
+
+        Returns:
```
FlorentinD: NIT - empty line at the beginning yes or no?

graphdatascience/graph/graph_object.py
```diff
@@ -41,24 +49,59 @@ def _graph_info(self, yields: List[str] = []) -> "Series[Any]":
...
+            the name of the database the graph is stored in
+        """
         return self._graph_info(["database"])  # type: ignore
 
     def configuration(self) -> "Series[Any]":
+        """
+
```
adamnsch, line 61: Nitpick: There's a lot of inconsistency with the blank lines here. Would be nice to fix that

doc/sphinx/source/conf.py
FlorentinD, line 28: looking at the return types did you try to not just return `DataFrame` but include the package name `pandas.DataFrame` or something?

graphdatascience/graph/graph_proc_runner.py
```diff
@@ -254,6 +256,7 @@ def streamNodeProperties(
         separate_property_columns: bool = False,
         **config: Any,
     ) -> DataFrame:
+        warning("Deprecated in favor of `gds.nodeProperties.stream`")
```
adamnsch, line 259: This is not true for all server versions right? It's also interesting to think about how we warn about things like this, since the server also does it. If we do keep it, I'd rather have a decorator for it

doc/sphinx/source/conf.py
```diff
@@ -0,0 +1,38 @@
...
+sys.path.append(os.path.join(os.path.dirname(__file__), "..", "..", ".."))
+
+extensions = [
+    "sphinx.ext.autodoc",
+    "sphinx.ext.napoleon",
+    # 'sphinx.ext.autosummary',
+]
+
+# napoleon_use_rtype = False
+
```
FlorentinD, line 29: In the docs we also have `self` inside the parameters, is it possible to not show `self`?

graphdatascience/graph/graph_proc_runner.py
```diff
@@ -43,6 +44,7 @@ def _path(package: str, resource: str) -> ContextManager[pathlib.Path]:
 
     @client_only_endpoint("gds.graph")
     def load_cora(self, graph_name: str = "cora", undirected: bool = False) -> Graph:
```
FlorentinD, line 46: here we dont have doc string to document the args?
Is the idea to add it later?

graphdatascience/model/model_alpha_proc_runner.py
```diff
@@ -9,6 +9,16 @@
 
 class ModelAlphaProcRunner(ModelResolver):
     def store(self, model: Model, failIfUnsupportedType: bool = True) -> "Series[Any]":
+        """
+        Persist a model from the Model Catalog to disk.
+
```
FlorentinD, line 14: We could not find this one in the generated docs.

Also the args are written in a different style (duplicating the type information)

graphdatascience/model/model_beta_proc_runner.py
```diff
@@ -9,6 +9,15 @@
 
 class ModelBetaProcRunner(UncallableNamespace, IllegalAttrChecker):
     def list(self, model: Optional[Model] = None) -> DataFrame:
+        """
+        List all models or a model matching the name of the specified model.
```
FlorentinD, line 13: Same as above, these are not showing up in the sphinx docs are they?

graphdatascience/pipeline/lp_training_pipeline.py
```diff
@@ -9,6 +9,16 @@
 
 class LPTrainingPipeline(ClassificationTrainingPipeline[LPModel]):
     def addFeature(self, feature_type: str, **config: Any) -> "Series[Any]":
+        """
+        Add a link feature to the pipeline.
+
+        Args:
+            feature_type: The type of feature to add. (For example, L2 or Hadamard)
```
FlorentinD, line 16:
```suggestion
            feature_type: The type of feature to add. 
```
Would suggest to remove this to avoid getting this outdated

graphdatascience/pipeline/nr_training_pipeline.py
```diff
@@ -9,24 +9,61 @@
 
 class NRTrainingPipeline(TrainingPipeline[NRModel]):
     def addLinearRegression(self, **config: Any) -> "Series[Any]":
+        """
+        Add a linear regression model to the pipeline.
```
FlorentinD, line 13:
```suggestion
        Add a linear regression model candidate to the pipeline.
```
In the server docs we call these candidates for clarity

graphdatascience/pipeline/nr_training_pipeline.py
```diff
@@ -9,24 +9,61 @@
...
+
+        """
         query = f"{self._query_prefix()}addLinearRegression($pipeline_name, $config)"
         params = {"pipeline_name": self.name(), "config": self._expand_ranges(config)}
 
         return self._query_runner.run_query(query, params).squeeze()  # type: ignore
 
     def addRandomForest(self, **config: Any) -> "Series[Any]":
+        """
+        Add a random forest regressor to the pipeline.
```
FlorentinD, line 29: probably we want to align the language here that this is also a candidate (applies to the other pipelines as well)

graphdatascience/graph/graph_proc_runner.py
brs96: Check the server GraphStreamRelationshipPropertiesProc.java, I think it has been marked deprecated in 2.2 & 2.3? Server logs the warning in neo4j.log.
With a decorator it will be a `@deprecated` like `@client_only_endpoint`?

graphdatascience/pipeline/pipeline_endpoints.py
```diff
@@ -13,16 +13,44 @@ def pipeline(self) -> PipelineProcRunner:
         return PipelineProcRunner(self._query_runner, f"{self._namespace}.pipeline", self._server_version)
 
     def lp_pipe(self, name: str) -> LPTrainingPipeline:
```
FlorentinD, line 15: same as above these are not used in the generated docs?

FlorentinD (changes requested):
Great work! Thanks for adding all the doc strings and setting up sphinx.

It would be great to also checkin the scripts to generate the proc docs.
(Thats our main issue)

In general we seem to omit the docs for the  proc runners.
These doc strings also have a different style compared to the rest.
Wonder if they were created at the very beginning?
We should make these match with the actual rendered docs

co reviewed with @adamnsch

doc/sphinx/Makefile
brs96: Yes, the project structure and the makefiles all generated by sphinx-quickstart.

brs96:
1. Scripts for the docs.
Unfortunately there's no single script that did it... The starting point is csv produced by gds.list(). For the algorithms I can add the signatures with regex. For most other procedures, I needed to look for the corresponding method in python client and add them manually.
2. Docstrings.
Why some methods have docstrings with doc and others not. Originally we tried with all docstrings+autofunctions, which would expose all the internal classes (those proc_runners), and namespace won't be formatted cleanly. Also, none of the algorithms will be shown. So we reverted to use py::function to list all procedures in the rst files, except for those meaningful classes (`Graph`, `GraphDataScience`, `LPTrainingPipeline` etc.). The docstrings model_a/b_proc_runners are unused, I'll remove them.

doc/sphinx/source/conf.py
brs96: I think we don't need it now.

doc/sphinx/source/conf.py
brs96: No, we could try it. For the py::function in .rst that can just be hardcoded, for the autofunction docstrings it will need a bit more effort.
But I think `DataFrame` should be clear enough?

doc/sphinx/source/conf.py
brs96: Not exactly sure, there could be a parameter for autofunction/docstring that hides this. Do we want to hide it?
Will look at this later.

graphdatascience/pipeline/pipeline_endpoints.py
brs96: These ones are actually used. Because they are in PipelineEndpoints <- DirecteEndpoints <- GraphDataScience.
All docstrings if they are there should be used, proc runners should not have docstrings and I'll remove those leftover.

doc/sphinx/Makefile
FlorentinD: maybe we can add a comment on the top :)

doc/sphinx/source/conf.py
FlorentinD: for this case you are probably right. Thought it could be nice to clarify, but I also did not find another reference doc

doc/sphinx/source/conf.py
FlorentinD: Adam and I thought it would be cleaner and also align as in some places we do already hide the self

FlorentinD:
> 1. Unfortunately there's no single script that did it... The starting point is csv produced by gds.list(). For the algorithms I can add the signatures with regex. For most other procedures, I needed to look for the corresponding method in python client and add them manually.

Oh, than I would fear that they get out of date. At least a check for completness could be useful.

brs96:
> > 1. Unfortunately there's no single script that did it... The starting point is csv produced by gds.list(). For the algorithms I can add the signatures with regex. For most other procedures, I needed to look for the corresponding method in python client and add them manually.
> 
> Oh, than I would fear that they get out of date. At least a check for completness could be useful.

Will at a script to do it. However, I was thinking how it can be done:
Start from gds.list(), that's like 407 procs I think. We can remove all the ignored_server_endpoints, but we need to count in the corresponding client ones. If it's 1-1 then we still have 407. If it's somehow not, then I'm not sure if the count can ever be maintained.
Additionally we need to count in the client_only_endpoints. Maintain a separately list and add the count there?

FlorentinD:
> 

Thats indeed harder than I hoped for. Maybe it makes sense to think of generating the actual endpoint code and then use docstrings there. I think Adam had some good ideas here. How about talking after standup tomorrow?

graphdatascience/error/deprecation_warning.py
```diff
@@ -0,0 +1,21 @@
+from functools import wraps
+from logging import warning
+from typing import Any, Callable, TypeVar, cast
+
+from ..caller_base import CallerBase
+
+F = TypeVar("F", bound=Callable[..., Any])
+
+
+def deprecation_warning(new_procedure: str) -> Callable[[F], F]:
```
adamnsch, line 10: Nice! :clap:

doc/tests/check_client_only_endpoints.py
adamnsch, line 1: I would recommend using `pytest`

graphdatascience/error/deprecation_warning.py
adamnsch: I guess this function could also take arguments specifying for which server versions the warning should be displayed

brs96:
Going to merge to publish from main, any other comment I'll include in follow up PR.

PR #352, opened 2023-04-27
Arrow remote projection procedure

soerenreichardt:
This PR introduces the capcability of connecting AuraDB with AuraDS by calling the `gds.alpha.graph.remote.project` procedure. There are no tests available yet as the setup for testing is a bit complicated: We need a Neo4j database with the arrow plugin installed (AuraDB) and a database with GDS installed (AuraDS), which are reachable via the arrow and bolt ports. There is no clear plan yet on how to manage this test setup and we're happy for ideas :)

### DIFF
--- a/graphdatascience/graph_data_science.py
+++ b/graphdatascience/graph_data_science.py
@@ -8,7 +8,7 @@ from .call_builder import IndirectCallBuilder
 from .endpoints import AlphaEndpoints, BetaEndpoints, DirectEndpoints
 from .error.unable_to_connect import UnableToConnectError
 from .error.uncallable_namespace import UncallableNamespace
-from .query_runner.arrow_query_runner import ArrowQueryRunner
+from .query_runner.arrow_query_runner import ArrowQueryRunner, AuraDbConnectionInfo
 from .query_runner.neo4j_query_runner import Neo4jQueryRunner
 from .query_runner.query_runner import QueryRunner
 from .server_version.server_version import ServerVersion
@@ -119,6 +119,13 @@ class GraphDataScience(DirectEndpoints, UncallableNamespace):
         self._query_runner.set_server_version(self._server_version)
 
         if arrow and self._server_version >= ServerVersion(2, 1, 0):
+            if aura_db_connection_info and self._server_version >= ServerVersion(2, 4, 0):
+                aura_db_conneciton_info = AuraDbConnectionInfo(
+                    aura_db_connection_info[0], aura_db_connection_info[1], self._config
+                )
+            else:
+                aura_db_conneciton_info = None
+
             try:
                 arrow_info: "Series[Any]" = self._query_runner.run_query(
                     "CALL gds.debug.arrow()", custom_error=False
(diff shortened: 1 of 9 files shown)


soerenreichardt:
[[ArrowPlugin] Implement new auth scheme in gds python client](https://trello.com/c/imrbDncF/8110-arrowplugin-implement-new-auth-scheme-in-gds-python-client)

graphdatascience/graph_data_science.py
```diff
@@ -144,6 +149,12 @@ def __init__(
                 ):
                     warnings.warn(f"Could not initialize GDS Flight Server client: {e}")
 
+        if aura_db_connection_info and self._server_version >= ServerVersion(2, 4, 0):
+            aura_db_conneciton_info = AuraDbConnectionInfo(
+                aura_db_connection_info[0], aura_db_connection_info[1], self._config
```
adamnsch, line 154: Will this always be the same config as for the other Neo4j driver? Or perhaps these are disjoint cases?

graphdatascience/query_runner/aura_db_arrow_query_runner.py
```diff
@@ -0,0 +1,137 @@
...
+        self._driver = GraphDatabase.driver(aura_db_endpoint, auth=auth, **config)
+        arrow_info: "Series[Any]" = (
+            Neo4jQueryRunner(self._driver, auto_close=True)
+            .run_query("CALL gds.debug.arrow.plugin()", custom_error=False)
+            .squeeze()
+        )
+
+        if not arrow_info.get("running"):
+            raise RuntimeError("Arrow server is not running")
+        listen_address: str = arrow_info.get("advertisedListenAddress")  # type: ignore
```
adamnsch, line 35: Is it an `Optional[str]`, if it can be `None`?

adamnsch (approved):
This looks really good! Nice work :) I actually don't think it would be that hard to set this up in CI. For these tests, instead of starting one DB we could start two with one having the Arrow plugin. And we should have the infrastructure to limit those tests to certain version of GDS and so on. Let me know if you want to discuss this further and/or pair on this. From my side it's also ok to have this untested in CI for now, but we should of course have something in the future

adamnsch:
We'll also need docs, but I guess that's in a later PR. Possibly a changelog too. Wdyt?

graphdatascience/graph_data_science.py
```diff
@@ -56,6 +60,7 @@ def __init__(
         arrow: bool = True,
         arrow_disable_server_verification: bool = True,
         arrow_tls_root_certs: Optional[bytes] = None,
+        aura_db_connection_info: Optional[Tuple[str, Tuple[str, str]]] = None,
```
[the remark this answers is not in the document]
soerenreichardt: I used `AuraDbConnectionInfo` now without the config field

graphdatascience/graph_data_science.py
soerenreichardt: You are right, they are disjoint. I used the default config now, just with a connection lifetime of 1 minute

soerenreichardt:
Thanks for the review. I carded changelog and documentation for when the testing and maybe also write-back is set up

PR #365, opened 2023-05-05
Sphinx neo4j theme

recrwplay:
Updates the Sphinx HTML generation to use a theme consistent with Neo4j docs.


doc/sphinx/source/conf.py
```diff
@@ -32,8 +32,15 @@
...
-html_theme = "alabaster"
-html_static_path = []  # type: ignore
+# use neo4j theme, which extends neo4j docs css for sphinx
+
+html_theme = "neo4j"
+html_theme_path = ["themes"]
+
+# 01-nav.js is a copy of a js file of the same name that is included in the docs-ui bundle
+def setup(app):
+    app.add_js_file('js/01-nav.js', loading_method="defer")
```
Mats-SX: Is `01-nav.js` a file that changes often?

Mats-SX:
I guess we don't yet have a PR build for the sphinx content? I couldn't see the impact of these changes as part of the standard PR build. I'll check through local checkout

Mats-SX:
Screenshot 2023-05-08 at 16 17 02
Screenshot 2023-05-08 at 16 17 34
Screenshot 2023-05-08 at 16 17 42


I built locally and observed the look from the above screenshots. This is obviously a huge step forward from the default style, but I have a few points of feedback:

- there's a lot of content in bold font, which feels a bit crowded on some pages. It would be nice to break that up and use bold only for the actual method name and then maybe non-bold italics or some color for arguments and/or return types
- I wonder if we can break the line of long class signatures so that they appear with one argument+type per line to make them more readable

There's some bad content in there as well, which I've carded up as separate work.
I pushed one commit to fix a code style issue in `conf.py`.

Mats-SX (approved):
I'd be happy to merge this as a good improvement, and to tackle further things on new PRs.

brs96 (approved):

doc/sphinx/source/conf.py
recrwplay: Not often at all. Both these js files are copies of the same files that the docs ui bundles as part of its overall js. The two added here contain the only code needed by this theme, but looking atit again, we might be better off just referencing the already published js: 

```
app.add_js_file("https://neo4j.com/docs/assets/js/site.js", loading_method="defer")
```

Mats-SX: I'll try this change and just push it if it works.

PR #367, opened 2023-05-09
Add cloud envs to tox.ini

soerenreichardt:


tox.ini
```diff
@@ -13,6 +13,9 @@ envlist =
 
 	py311-main-aura,
 
+	py37-neo4j{4,5}-pyarrow{10,11,12}-main-{cloud-architecture}
+	py37-neo4j{4,5}-pyarrow{10,11,12}-pullrequest-{cloud-architecture}
```
adamnsch, line 17: On PRs we try to keep the scope small so only use the latest pyarrow dependency (see other envs listed in this file). Otherwise it takes a long time to get CI feedback on PRs

PR #368, opened 2023-05-09
Improve ref docs navigation

Mats-SX:
- Add more pages for various objects
- Omit constructors that are not meant for users
- Add more cross-references, especially for object construction

### DIFF
--- a/doc/sphinx/source/graph_object.rst
+++ b/doc/sphinx/source/graph_object.rst
@@ -3,3 +3,4 @@ Graph object
 
 .. autoclass:: graphdatascience.graph.graph_object.Graph
    :members:
+   :exclude-members: __init__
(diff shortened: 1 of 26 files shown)


doc/sphinx/source/graph_object.rst
```diff
@@ -3,3 +3,4 @@ Graph object
 
 .. autoclass:: graphdatascience.graph.graph_object.Graph
    :members:
+   :exclude-members: __init__
```
adamnsch, line 6: :sweat_smile:

adamnsch (approved):
LGMT, nice work :clap:

doc/sphinx/source/graph_object.rst
Mats-SX: Yeah it doesn't seem like a good idea to document the constructors of the objects that are supposed to be constructed using actual procedure calls. Or factories, depending on how you view it.

PR #374, opened 2023-05-17
Add test for remote projection

soerenreichardt:


tox.ini
```diff
@@ -8,6 +8,7 @@ envlist =
 	py311-pandas{1,2}-pyarrow{10,11,12}-main-{standard,encrypted},
 
 	py37-neo4j{4,5}-pullrequest-{standard,encrypted},
+	py37-pyarrow{4,5,6,7,8,9,10,11,12}-pullrequest-{standard,encrypted},
```
brs96, line 11: This will slow down the builds quite a lot, do we need all the pyarrow versions? Perhaps we can skip a few.

brs96 (approved):

PR #376, opened 2023-05-17
Arrow remote write back

soerenreichardt:


graphdatascience/query_runner/aura_db_arrow_query_runner.py
```diff
@@ -66,6 +66,17 @@ def run_query(
             params["host"] = aura_db_arrow_endpoint
             params["config"] = {"useEncryption": False}
 
+        if ".write" in query and "config" in params and "remote" in params["config"] and params["config"]["remote"]:
```
adamnsch, line 69: Is this safe? Should there be a `else return ...` in the previous if statement before this?

adamnsch (approved):
LGTM :) Just have a question

PR #400, opened 2023-06-15
More neo4j theming

recrwplay:
Updates the Sphinx docs theme following feedback:

- Adds a background to the navigation to match other neo4j.com/docs pages
- Updates a header link target
- Fixes a rendering issue with deprecated content, removing the style that was inherited from the main docs assets and adding new styles to mark content as deprecated more clearly

### DIFF
--- a/doc/sphinx/source/themes/neo4j/layout.html
+++ b/doc/sphinx/source/themes/neo4j/layout.html
@@ -7,6 +7,10 @@
 {%- block relbar1 %}{% endblock %}
 {%- block relbar2 %}{% endblock %}
 
+{% block extrahead %}
+<link href="https://fonts.googleapis.com/css2?family=Nunito+Sans:wght@300;400;600;700&display=swap" rel="stylesheet">
+{% endblock %}
+
 {%- block content %}
 <body class="article docs sphinx">
 
@@ -16,9 +20,9 @@
   <nav class="navbar">
     <div class="navbar-brand">
       <a class="navbar-item" href="/">
-
-      <img class="navbar-logo" src="https://dist.neo4j.com/wp-content/uploads/20210422142941/neo4j-logo-2020.svg" alt="Neo4j Graph Data Science Client API Reference">
-      </a><a href="/docs/" class="navbar-item no-left-padding" aria-label="Neo4j Graph Data Science Client API Reference">
+        <img class="navbar-logo" src="https://dist.neo4j.com/wp-content/uploads/20210422142941/neo4j-logo-2020.svg" alt="Neo4j Graph Data Science Client API Reference">
+      </a>
+      <a href="/docs/graph-data-science-client" class="navbar-item no-left-padding" aria-label="Neo4j Graph Data Science Client API Reference">
         Graph Data Science Client API Reference
       </a>
 
--- a/doc/sphinx/source/themes/neo4j/layout.html
+++ b/doc/sphinx/source/themes/neo4j/layout.html
@@ -22,7 +22,7 @@
       <a class="navbar-item" href="/">
         <img class="navbar-logo" src="https://dist.neo4j.com/wp-content/uploads/20210422142941/neo4j-logo-2020.svg" alt="Neo4j Graph Data Science Client API Reference">
       </a>
-      <a href="/docs/graph-data-science-client" class="navbar-item no-left-padding" aria-label="Neo4j Graph Data Science Client API Reference">
+      <a href="{{ pathto(root_doc) }}" class="navbar-item no-left-padding" aria-label="Neo4j Graph Data Science Client API Reference">
         Graph Data Science Client API Reference
       </a>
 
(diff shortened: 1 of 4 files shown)


doc/sphinx/source/themes/neo4j/static/css/neo4j.css
```diff
@@ -190,3 +194,40 @@ div.code-block-caption:hover > a.headerlink {
...
+}
+
+body.sphinx .doc dl.deprecated {
+    margin-bottom: 0;
+    padding-bottom: 0;
+}
+
+body.sphinx .doc div.deprecated {
+    padding: 1rem;
+}
```
Mats-SX, line 233: It's very in-your-face. Maybe this follows another color shade from docs tooling? If not, I would suggest maybe a little softer color.
Screenshot 2023-06-16 at 13 28 07

I'm thinking about Neo4j Browser's highlighting, which uses a yellow squiggly line:

Screenshot 2023-06-16 at 13 37 13
Screenshot 2023-06-16 at 13 37 22

It seems to me desirable that we have a specific color to indicate the concept of "deprecated". I did a search in various docs and found [this](https://neo4j.com/docs/apoc/current/overview/apoc.warmup/apoc.warmup.run/):

Screenshot 2023-06-16 at 13 40 08

My experience also tells me a yellowish color seems intuitive. WDYT?

doc/sphinx/source/themes/neo4j/layout.html
```diff
@@ -7,6 +7,10 @@
 {%- block relbar1 %}{% endblock %}
 {%- block relbar2 %}{% endblock %}
 
+{% block extrahead %}
+<link href="https://fonts.googleapis.com/css2?family=Nunito+Sans:wght@300;400;600;700&display=swap" rel="stylesheet">
```
Mats-SX, line 11: font looks good I think

doc/sphinx/source/themes/neo4j/layout.html
```diff
@@ -16,9 +20,9 @@
   <nav class="navbar">
     <div class="navbar-brand">
       <a class="navbar-item" href="/">
-
-      <img class="navbar-logo" src="https://dist.neo4j.com/wp-content/uploads/20210422142941/neo4j-logo-2020.svg" alt="Neo4j Graph Data Science Client API Reference">
-      </a><a href="/docs/" class="navbar-item no-left-padding" aria-label="Neo4j Graph Data Science Client API Reference">
+        <img class="navbar-logo" src="https://dist.neo4j.com/wp-content/uploads/20210422142941/neo4j-logo-2020.svg" alt="Neo4j Graph Data Science Client API Reference">
+      </a>
+      <a href="/docs/graph-data-science-client" class="navbar-item no-left-padding" aria-label="Neo4j Graph Data Science Client API Reference">
```
Mats-SX, line 25: Can we target the index of the API refdocs instead? We have the target back to the client manual in the top right already.
```suggestion
      <a href="/docs/graph-data-science-client/{version}/api/" class="navbar-item no-left-padding" aria-label="Neo4j Graph Data Science Client API Reference">
```

This suggestion is probably not good but I don't know how to get the `{version}` part in there.

doc/sphinx/source/themes/neo4j/static/css/neo4j.css
```diff
@@ -56,6 +56,10 @@ body.sphinx .content .search {
         height: auto;
         background: initial;
     }
+
+    body.sphinx .sphinxsidebar {
+        background: #f5f7fa;
+    }
```
Mats-SX, line 62: this looks very nice

doc/sphinx/source/themes/neo4j/static/css/neo4j.css
```diff
@@ -190,3 +194,40 @@ div.code-block-caption:hover > a.headerlink {
...
+
+body.sphinx .doc ul li{
+    margin-bottom: 0.5rem;
+}
+
+/* functions */
+
+body.sphinx .doc dl.py.function {
+    padding-top: 2rem;
+    padding: 1rem;
```
Mats-SX, line 208: This I also like a lot

Mats-SX (commented):
I think it would be nice to consider a yellow color for the deprecation, or at least think about a good normative color scheme for that concept, also considering other products and docs products.

And I would like the top-left to link back to the API refdocs index page rather than back to the manual.

doc/sphinx/source/themes/neo4j/layout.html
recrwplay: Looks like we can use the `pathto` helper here. I've updated the link.

doc/sphinx/source/themes/neo4j/static/css/neo4j.css
recrwplay: How does this look?

Screenshot 2023-06-20 at 09 29 04

We could as a future update hide the deprecated div content and display it on hover (so the behavior would be like the glossary entries).

PR #407, opened 2023-06-20
Add wrapper over run_cypher for using Cypher Projections

knutwalker:


doc/tests/test_client_only_endpoints.py
```diff
@@ -59,7 +59,7 @@ def find_covered_server_endpoints() -> List[str]:
 
     driver.close()
 
-    return [ep["name"] for ep in all_server_endpoints if not ep["name"] in IGNORED_SERVER_ENDPOINTS]
+    return [ep["name"] for ep in all_server_endpoints if ep["name"] not in IGNORED_SERVER_ENDPOINTS]
```
s1ck, line 62: :crab: :hocho: :snake:

graphdatascience/graph/graph_cypher_runner.py
```diff
@@ -0,0 +1,134 @@
...
+        result = result.squeeze()
+
+        return Graph(graph_name, self._query_runner, self._server_version), result  # type: ignore
+
+    __separators = re.compile(r"[,(.]")
+
+    def _find_return_clause_graph_name(self, query: str) -> str:
+        """
+        Returns the 'graph name' in the RETURN clause of a Cypher projection query.
+        'graph name' here are the first argument of the `gds.graph.project` function.
```
s1ck, line 85:
```suggestion
        'graph name' here is the first argument of the `gds.graph.project` function.
```

graphdatascience/graph/graph_cypher_runner.py
```diff
@@ -0,0 +1,134 @@
...
+        """
+
+        found = None
+        at_end = False
+
+        namespace_tokens = self._namespace.split(".")
+        query_tokens = iter(query.split())
+
+        for query_token in query_tokens:
+            print("query token", query_token)
```
s1ck, line 95: `poop`?

graphdatascience/graph/graph_cypher_runner.py
```diff
@@ -0,0 +1,134 @@
...
+                    f"Invalid query, the query must end with the `RETURN {self._namespace}(...)` call: {query}"
+                )
+
+            if found is not None:
+                paren_balance = 0
+
+                chars = chain(query_token, (c for tok in query_tokens for c in tok))
+
+                for c in chars:
+                    print("char", c)
```
s1ck, line 108: `poop`?

graphdatascience/graph/graph_cypher_runner.py
```diff
@@ -0,0 +1,134 @@
...
+            qr = qr.fallback_query_runner()
+
+        result = qr.run_query(query, params, database, False)
+        result = result.squeeze()
+
+        return Graph(graph_name, self._query_runner, self._server_version), result  # type: ignore
+
+    __separators = re.compile(r"[,(.]")
+
+    def _find_return_clause_graph_name(self, query: str) -> str:
```
s1ck, line 82: Should this maybe live in some util file? With separate tests .. :)

graphdatascience/graph/graph_cypher_runner.py
```diff
@@ -0,0 +1,134 @@
...
+                    elif c == ")":
+                        paren_balance -= 1
+                        if paren_balance < 0:
+                            at_end = True
+                            break
+
+            if query_token == "RETURN":
+                tokens = (tok for token in query_tokens for tok in GraphCypherRunner.__separators.split(token) if tok)
+                for token, expected in zip_longest(tokens, namespace_tokens):
+                    print("inner token", token, "expected", expected)
```
s1ck, line 120: `poop`?

s1ck (approved):
Looks good. Maybe we can separate the string parsing function?

graphdatascience/graph/graph_cypher_runner.py
knutwalker: whoops

graphdatascience/graph/graph_cypher_runner.py
knutwalker: whoopsie

graphdatascience/graph/graph_cypher_runner.py
knutwalker: whoopsie oopsie 🙈

graphdatascience/graph/graph_cypher_runner.py
knutwalker: Dunno about a util file, it is not really generic in what it does. I did add some tests though.

knutwalker:
@s1ck thanks for the review, 🏓

graphdatascience/graph/graph_cypher_runner.py
```diff
@@ -0,0 +1,135 @@
...
+    __separators = re.compile(r"[,(.]")
+
+    @staticmethod
+    def _verify_query_ends_with_return_clause(namespace: str, query: str) -> None:
+        """
+        Verifies that the query ends in a `RETURN gds.graph.project(...)` call.
+        Invalid queries will raise a ValueError.
+        """
+
+        # We iterate through the "tokens" of the query, where a token here is
```
FlorentinD, line 85: it looks like a regexp could also work here, but I also think your version could be more maintainable :)

FlorentinD (commented):
LGTM, Just some smaller comments :)

doc/modules/ROOT/pages/graph-object.adoc
```diff
@@ -78,6 +78,59 @@ In addition to those aforementioned there are five more methods that create grap
...
+[[cypher-projection]]
+== Projecting a graph using Cypher Projection
+
+The method `gds.graph.cypher.project` allows for projecting a graph using Cypher projection.
+Cypher projection is not a dediicated procedure; rather it requires writing a Cypher query that calls the `gds.graph.project` aggregation.
+
+Read more about Cypher projection in the https://neo4j.com/docs/graph-data-science/current/management-ops/projections/graph-project-cypher-projection/[GDS manual].
+
+The method `gds.graph.cypher.project` bridges the gap between <<getting-started-minimal-example, `gds.run_cypher`>> and `gds.graph.get`.
+It takes as parameters a query string `query: str`, an optional string `database: Optional[str]` to override which database to target as well as an optional Cypher parameters dictionary as keywords args `**params: Any`.
```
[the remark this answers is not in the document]
knutwalker: Ok, I was copying from what I've found about run_cypher. I can certainly add a syntax table instead

graphdatascience/graph/graph_cypher_runner.py
```diff
@@ -0,0 +1,135 @@
...
+        """
+
+        GraphCypherRunner._verify_query_ends_with_return_clause(self._namespace, query)
+
+        # See run_cypher
+        qr = self._query_runner
+
+        # The Arrow query runner should not be used to execute arbitrary Cypher
+        if isinstance(qr, ArrowQueryRunner):
+            qr = qr.fallback_query_runner()
```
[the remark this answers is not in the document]
knutwalker: ok 👍

graphdatascience/graph/graph_cypher_runner.py
```diff
@@ -0,0 +1,135 @@
...
+
+        result = qr.run_query(query, params, database, False)
+        result = result.squeeze()
+        try:
+            graph_name = str(result["graphName"])
```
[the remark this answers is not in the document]
knutwalker: in my understanding, try/catch is more pythonic and pthonistas will tell you all about ["it's easier to ask for forgiveness"](https://docs.python.org/3/glossary.html#term-EAFP) and point you to a [stackoverflow answer](https://stackoverflow.com/questions/11360858/what-is-the-eafp-principle-in-python) or [two](https://stackoverflow.com/questions/12265451/ask-forgiveness-not-permission-explain) ^^

graphdatascience/graph/graph_cypher_runner.py
```diff
@@ -0,0 +1,135 @@
...
+                    # so we assume that the key was graphName and use the value.
+                    graph_name = result
+            else:
+                raise ValueError(
+                    f"Invalid query, the query must end with the `RETURN {self._namespace}(...)` call: {query}"
+                )
+
+        return GraphCreateResult(Graph(graph_name, self._query_runner, self._server_version), result)  # type: ignore
+
+    __separators = re.compile(r"[,(.]")
```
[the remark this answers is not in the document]
knutwalker: I'm gonna sneak in some not-top-of-the-class constants in Java as well, I think it makes sense to define those as close to the usage as possible. If it's only uses within one method, I'd define it within this method (if the language permits that) or next to that method, i.e. above. This way, related code is together in the source and I don't have to jump around in the file. If a constant is used across methods, its scope is the whole class and so it can live at the top.
That being said, this is **not** a constant though, this is a class variable, which is supposed to follow the same naming conventions as function and method names.
A constant would be on the module level at the top of the file (and indeed in SCREAMING_UPPER_CASE), but again, I like to reduce the scope of it, so it's not a module level constant because we only need it in the one method.

graphdatascience/graph/graph_cypher_runner.py
knutwalker: This reminds me of https://stackoverflow.com/a/1732454 ^^

graphdatascience/tests/unit/test_graph_cypher.py
```diff
@@ -0,0 +1,146 @@
+import pytest
+from pandas import DataFrame
+
+from .conftest import CollectingQueryRunner
+from graphdatascience.graph.graph_cypher_runner import GraphCypherRunner
+from graphdatascience.graph_data_science import GraphDataScience
+from graphdatascience.server_version.server_version import ServerVersion
```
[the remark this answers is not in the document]
knutwalker: thanks :)

knutwalker:
@FlorentinD thanks for the review and 🏓

graphdatascience/graph/graph_cypher_runner.py
```diff
@@ -42,36 +41,24 @@ def project(
...
-        qr = self._query_runner
+        # 'rows' as argument for squeeze() means that we only squeeze a single row of a DataFrame
+        # into a Series. Without this, a single-column DataFrame would be squeezed into a scalar,
+        # and we cannot index that scalar with a string.
+        # This happens in our unit tests, where we mock the result of the query runner and don't
+        # want to mock additional data for squeeze() to work.
+        # However, there is a bug in pandas, where the typing definition rejects 'rows' as valid
+        # argument for squeeze(). The documentation also doesn't mention 'rows' at first, but lists
+        # an example with 'rows' further down, so we have to silence mypy ¯\_(ツ)_/¯
+        result = self._query_runner.run_query(query, params, database, False).squeeze("rows")  # type: ignore
```
FlorentinD, line 52: thorough investigation :D

knutwalker:
@FlorentinD test failures? nevah! 🙈 

Apparently the squeeze("rows") thing didn't work out :/
I though I ran the ITs locally, but apparently I did not. Anyway, back to the old squeezing like you suggested initially

FlorentinD:
Looking at the failure, we need to update the `aura_db_arrow_query_runner`as well

PR #408, opened 2023-06-22
Add new notebooks to the documentation

FlorentinD:


doc/modules/ROOT/pages/tutorials/centrality-algorithms.adoc
```diff
@@ -0,0 +1,317 @@
+// DO NOT EDIT - AsciiDoc file generated automatically
+
+= Centrality Algorithms
```
adamnsch, line 3: Missing colab link? Other page too

adamnsch (changes requested):
Looks good! But are we missing colab links?

doc/modules/ROOT/pages/tutorials/centrality-algorithms.adoc
FlorentinD: true, they exist in the notebooks but got lost during conversion :thinking:

FlorentinD: @nvitucci do you have an idea why the cell is ignored? I cant spot the difference to other notebooks

nvitucci: The way the converter works at the moment requires the `colab_type` key to be present in the Colab cell metadata (which is empty in this notebook), like this:

```
"metadata": {
    "colab_type": "text",
    "id": "view-in-github"
   }
```

FlorentinD: thanks that solved it. Should have looked at the raw files and not the VS Code view

adamnsch (approved):

PR #410, opened 2023-06-27
Improve graph creations + with expression

FlorentinD:


graphdatascience/tests/integration/test_graph_object.py
```diff
@@ -154,3 +154,21 @@ def test_with_graph(gds: GraphDataScience) -> None:
         assert TempG.exists()
 
     assert not gds.graph.exists(temp_graph_name)["exists"]
```
FlorentinD, line 156: the test above still works so this is not breaking :)

doc/sphinx/source/graph.rst
```diff
@@ -93,11 +93,11 @@ These all assume that an object of :class:`.GraphDataScience` is available as `g
 
     Streams the given node property.
 
-.. py:function:: gds.graph.project(graph_name: str, node_spec: Any, relationship_spec: Any, **config: Any) -> Tuple[Graph, Series[Any]]
+.. py:function:: gds.graph.project(graph_name: str, node_spec: Any, relationship_spec: Any, **config: Any) -> GraphResult
 
     Creates a named graph in the catalog for use by algorithms.
-
-.. py:function:: gds.graph.project.cypher(graph_name: str, node_spec: Any, relationship_spec: Any, **config: Any) -> Tuple[Graph, Series[Any]]
+# FIXME - how to document GraphResult here
```
FlorentinD, line 99: How can I document this dataclass correctly. I did not know where it fit. A seperate page seemed overkill

graphdatascience/graph/graph_result.py
```diff
@@ -0,0 +1,35 @@
...
+from types import TracebackType
+from typing import Optional, Type, TypeVar
+
+from ..graph.graph_object import Graph
+
+TGraphResult = TypeVar("TGraphResult", bound="GraphResult")
+
+
+@dataclass(frozen=True)
+class GraphResult:
```
Mats-SX, line 11: In order to not be a breaking change, this class needs to operate just like a Tuple. I would've expected it to inherit from `Tuple`, but perhaps you've managed to just implement the methods a Tuple requires?

doc/sphinx/source/graph.rst
Mats-SX: Separate page is the structure we have for it now. You can compare to `NodePropertyStep` or `LinkFeature` classes. It's been fed back that it's not great but no solution has been made yet. I would pile it up next to those and solve them all later.

graphdatascience/graph/graph_result.py
Mats-SX: The name here is interesting. I want to think back to our session on "conceptualising graph projection" at GDSCon Spring 2023, where we defined the operation of "graph creation". I think this type will always be returned when any of the "graph creation" operations is used; projections, generation, import, construct, etc. (`graph.get` is a bit of an exception IMO; it doesn't really _do_ anything).

So I would like to consider these names

```
GraphCreationResult
GraphCreateResult
GraphCreation
```

My preference is `GraphCreationResult`. WDYT?

graphdatascience/tests/integration/test_graph_object.py
Mats-SX: excellent. it's a good indicator that you mimicked the Tuple type. I'd like to make sure we thoroughly analyse that fact though, studying all aspects of the Tuple type. Chances are that someone is using it in a different way than we do in this repo.

graphdatascience/graph/graph_result.py
```diff
@@ -0,0 +1,35 @@
...
+    ) -> None:
+        self.graph.drop()
+
+    def __getitem__(self, item):
+        if item == 0:
+            return self.graph
+        if item == 1:
+            return self.result
+
+        raise KeyError("Key must be between 0 and 1")
```
Mats-SX, line 35: Maybe we should add a class comment that this class has these methods because it is replacing a Tuple and wants to be backwards compat

Mats-SX (commented):
This is exactly what I was imagining. Great stuff! I am concerned with backwards compat and naming, but I am very much in favour.

changelog.md
```diff
@@ -12,6 +12,8 @@
 
 ## Improvements
 
+* Improved endpoint returning graphs to be used in `with` clauses. The expression `with gds.graph.project(...)[0] AS G` can now be simplified to `with gds.graph.project(...) AS G`.
```
adamnsch, line 15: `AS`?

doc/sphinx/source/algorithms.rst
```diff
@@ -88,7 +88,7 @@ These all assume that an object of :class:`.GraphDataScience` is available as `g
 
 
 .. py:function:: gds.alpha.graph.sample.rwr(graph_name: str, from_G: Graph, **config: Any)
-    -> Tuple[Graph, Series[Any]]
+    -> GraphResult
```
adamnsch, line 91: If we do this, it needs to be exported as well

graphdatascience/graph/graph_result.py
```diff
@@ -0,0 +1,35 @@
...
+        return self.graph
+
+    def __exit__(
+        self,
+        exception_type: Optional[Type[BaseException]],
+        exception_value: Optional[BaseException],
+        traceback: Optional[TracebackType],
+    ) -> None:
+        self.graph.drop()
+
```
adamnsch, line 28: I think there should also be `graph` and `result` attributes so that the user doesn't have to use tuple indexing (though it should of course be there for compat)

graphdatascience/graph/graph_result.py
FlorentinD: I like the GraphCreationResult. Thanks for the naming ideas

doc/sphinx/source/algorithms.rst
FlorentinD: what do you mean with export here?

adamnsch: It should be put in the root `__init__.py` I think. Although maybe we don't need that. It's just that `graph` is an internal namespace

FlorentinD:
checks are failing but also main is red. I will take a look to get main green again

doc/sphinx/source/algorithms.rst
FlorentinD: we agreed that there are more classes to expose there. Will do a follow up PR

PR #412, opened 2023-06-29
Use ruff linter instead of flake8

knutwalker:


requirements/dev/dev.txt
```diff
@@ -7,6 +7,7 @@ nbconvert == 7.4.0
 pandas-stubs == 1.5.3.230321
 pytest == 7.3.1
 pytest-annotate == 1.0.5
+ruff == 0.0.272
```
adamnsch, line 10: The only thing that makes me hesitate is the pre-release status

adamnsch (changes requested):
We also have to update the docs (`TESTING.md` I believe). Also, should we remove the flake8 conf file? I guess it doesn't really hurt to keep it around

knutwalker:
@adamnsch thanks for the review. Since getting this green on CI needs updating our build docker image and there is a workaround in #413 for my issues, I'm gonna close this :sad-ferris:

PR #422, opened 2023-07-03
Map new gds.license.state endpoint

FlorentinD:


graphdatascience/tests/integration/test_system_ops.py
```diff
@@ -41,6 +42,11 @@ def test_is_licensed(gds: GraphDataScience) -> None:
     assert gds.is_licensed()
 
 
+@pytest.mark.compatible_with(min_inclusive=ServerVersion(2, 5, 0))
+def test_license_state(gds: GraphDataScience) -> None:
+    assert gds.license.state()["isLicensed"]
```
Mats-SX, line 47: hm this might fail for a contributor who doesn't have a license installed when running tests

Mats-SX: maybe we can just assert that the column exists without mandating that it is `true`?
```suggestion
    assert isinstance(gds.license.state()["isLicensed"], bool)
```

Mats-SX (commented):
Kinda wanna merge this to make `main` green, but I have one enhancement for unlicensed testers that could make sense

Mats-SX:
Also I don't understand why the CloudArch tests are failing

graphdatascience/tests/integration/test_system_ops.py
Mats-SX: This did not run green. 

```
[Step 1/1]     def test_license_state(gds: GraphDataScience) -> None:
[07:21:47][Step 1/1] >       assert isinstance(gds.license.state()["isLicensed"], bool)
[07:21:47][Step 1/1] E       assert False
[07:21:47][Step 1/1] E        +  where False = isinstance(True, bool)
```

PR #424, opened 2023-07-04
Try fix failing Cloud tests

FlorentinD:


graphdatascience/tests/unit/conftest.py
```diff
@@ -12,7 +12,7 @@
 from graphdatascience.server_version.server_version import ServerVersion
 
 # Should mirror the latest GDS server version under development.
-DEFAULT_SERVER_VERSION = ServerVersion(2, 4, 0)
+DEFAULT_SERVER_VERSION = ServerVersion(2, 5, 0)
```
Mats-SX, line 15: Ah, that's a good one. We should add a checklist item for updating it on our branching card.

graphdatascience/tests/unit/conftest.py
```diff
@@ -65,8 +65,11 @@ def runner(server_version: ServerVersion) -> CollectingQueryRunner:
 
 
 @pytest.fixture
-def gds(runner: CollectingQueryRunner) -> GraphDataScience:
-    return GraphDataScience(runner, arrow=False)
+def gds(runner: CollectingQueryRunner) -> Generator[GraphDataScience, None, None]:
+    gds = GraphDataScience(runner, arrow=False)
+    yield gds
+
+    gds.close()
```
Mats-SX, line 72: super odd but I do suspect this line is the fix

Mats-SX (approved):

adamnsch (commented):
Nice! :) I thought we had a `__del__` on the GDS object to take care of this, but I guess not haha. SHould we have though? `__del__` is not the best modern python practice though I think...

FlorentinD:
I cannot find the `__del__` method on the GDS object.

Without knowing about good python practices, I think this could be a good case for a `__del__` method

PR #426, opened 2023-07-06
Add support for Neo4j driver bookmarks

adamnsch:

Make sure:

Remains to be done:
- [x] Changelog entry
- [x] Manual docs
- [x] Reference docs

### DIFF
--- a/graphdatascience/graph_data_science.py
+++ b/graphdatascience/graph_data_science.py
@@ -1,7 +1,7 @@
 import warnings
 from typing import Any, Dict, Optional, Tuple, Type, TypeVar, Union
 
-from neo4j import Bookmarks, Driver, GraphDatabase
+from neo4j import Driver, GraphDatabase
 from pandas import DataFrame, Series
 
 from .call_builder import IndirectCallBuilder
@@ -40,7 +40,7 @@ class GraphDataScience(DirectEndpoints, UncallableNamespace):
         arrow_disable_server_verification: bool = True,
         arrow_tls_root_certs: Optional[bytes] = None,
         aura_db_connection_info: Optional[AuraDbConnectionInfo] = None,
-        bookmarks: Optional[Bookmarks] = None,
+        bookmarks: Optional[Any] = None,
     ):
         """
         Construct a new GraphDataScience object.
@@ -66,6 +66,8 @@ class GraphDataScience(DirectEndpoints, UncallableNamespace):
         arrow_tls_root_certs : Optional[bytes], default None
             PEM-encoded certificates that are used for the connecting to the
             Arrow Flight server.
+        bookmarks : Optional[Any], default None
+            The Neo4j bookmarks to require a certain state before the next query gets executed.
         """
 
         if isinstance(endpoint, str):
@@ -164,13 +166,13 @@ class GraphDataScience(DirectEndpoints, UncallableNamespace):
         """
         self._query_runner.set_database(database)
 
-    def set_bookmarks(self, bookmarks: Bookmarks) -> None:
+    def set_bookmarks(self, bookmarks: Any) -> None:
         """
         Set Neo4j bookmarks to require a certain state before the next query gets executed
 
         Parameters
         ----------
-        bookmarks: Bookmarks
+        bookmarks: Bookmark(s)
             The Neo4j bookmarks defining the required state
         """
         self._query_runner.set_bookmarks(bookmarks)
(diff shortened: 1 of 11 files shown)


graphdatascience/graph_data_science.py
```diff
@@ -74,7 +75,7 @@ def __init__(
 
             driver = GraphDatabase.driver(endpoint, auth=auth, **self._config)
 
-            self._query_runner = Neo4jQueryRunner(driver, auto_close=True)
+            self._query_runner = Neo4jQueryRunner(driver, auto_close=True, bookmarks=bookmarks)
```
FlorentinD, line 78: Did you know about https://neo4j.com/docs/api/python-driver/current/api.html#neo4j.api.BookmarkManager ? 

The drivers team recommended using it :)

FlorentinD: I think passing an Optional BookMarkManger could make more sense

graphdatascience/query_runner/neo4j_query_runner.py
```diff
@@ -45,7 +51,7 @@ def run_query(
 
         self._verify_connectivity()
 
-        with self._driver.session(database=database) as session:
+        with self._driver.session(database=database, bookmarks=self._bookmarks) as session:
```
FlorentinD, line 54: I still think we might need to propagate the bookmarks to the server. But we should verify if the outer read transaction is enough. 

As the inner GDS writes will also need to know about the initial cypher writes

FlorentinD: or f.i. the native projection uses a new transaction to check if there is a label index.
Or a old cypher projection also uses internal transactions

graphdatascience/graph_data_science.py
FlorentinD: also they note the potential performance impact, so keeping it optional makes sense.
(https://neo4j.com/docs/api/python-driver/current/api.html#bookmark-manager-ref)

adamnsch: I've just read https://neo4j.com/docs/python-manual/current/bookmarks/ To implement a bookmarkmanager seems a bit overkill since we have exactly the bookmark support we need already in the session API?

graphdatascience/query_runner/neo4j_query_runner.py
adamnsch: Is it not enough to stall the entire query on a "client" level for that? Is Nathan also propagating bookmarks to the server? Would love to discuss this

graphdatascience/query_runner/neo4j_query_runner.py
```diff
@@ -48,6 +54,7 @@ def run_query(
         with self._driver.session(database=database) as session:
```
FlorentinD, line 54: so here we need to add the bookmarks.

graphdatascience/graph_data_science.py
FlorentinD: i think you are right.
Also thinking further, I imagine we dont need to pass the bookmarks to the server afterall.
As any query run inside the session will be after the bookmark. So internal transactions created f.i. during projection should still be after the bookmark here.

graphdatascience/graph_data_science.py
```diff
@@ -155,9 +156,41 @@ def __getattr__(self, attr: str) -> IndirectCallBuilder:
...
+        -------
+        The (possibly None) Neo4j bookmarks defining the currently required state
+        """
+        return self._query_runner.bookmarks()
+
+    def last_bookmarks(self) -> Optional[Bookmarks]:
+        """
+        Get the Neo4j bookmarks defining the state following the most recently called query
+
+
```
adamnsch, line 187: remove one newline

graphdatascience/graph_data_science.py
```diff
@@ -1,7 +1,7 @@
 import warnings
 from typing import Any, Dict, Optional, Tuple, Type, TypeVar, Union
 
-from neo4j import Driver, GraphDatabase
+from neo4j import Bookmarks, Driver, GraphDatabase
```
FlorentinD, line 4: Bookmarks is only available in the Neo4j 5 driver.

FlorentinD: For Neo4j 4 driver we can use `Bookmark` it seems

brs96 (approved):

nvitucci (commented):
I cannot make other suggestions, probably because the PR has already been merged. I can still submit the review in progress, which I have done.

FlorentinD:
I did not know that its not possible to review after merge. I created a PR now with your first suggestions at https://github.com/neo4j/graph-data-science-client/pull/483

PR #432, opened 2023-07-18
Broken link to GDS manual

lidiazuin:


doc/modules/ROOT/pages/known-limitations.adoc
```diff
@@ -4,4 +4,4 @@ Operations known to not work with the Python client are:
 
 * https://neo4j.com/docs/graph-data-science/current/management-ops/utility-functions/#utility-functions-numeric[Numeric utility functions] (will never be supported)
 * https://neo4j.com/docs/graph-data-science/current/management-ops/create-cypher-db/[Cypher on GDS] (might be supported in the future)
-* https://neo4j.com/docs/graph-data-science/current/management-ops/projections/graph-project-cypher-aggregation/[Projecting graphs using Cypher Aggregation] (might be supported in the future)
```
FlorentinD, line 7: I think it would be even better to add a redirect to the server docs instead of fixing it here (`page-alias`). Then we other places we cant control wont break

FlorentinD (approved):
Thank you for fixing our broken links!

In this case its a url we control, so I suggest to fix it on the server side as well.
Would you be up for creating a PR there? 🙏

doc/modules/ROOT/pages/known-limitations.adoc
lidiazuin: It turns out that this is already set there with `page-aliases`, both in version 2.4 and master (see here https://github.com/neo4j/graph-data-science/commit/d6a869a5bf9cc32c3971a80f7d34cf21ce8a2e91). I thought it could be unpublished on TC, but it seems to be fine there as well. I'm not sure why this is not catching here?

FlorentinD: this is only for the 2.5-preview not for 2.4.
We need to cherry pick the typo fix to the 2.4 branch I think

PR #434, opened 2023-07-20
Update Closeness Centrality refs

vnickolov:


vnickolov:
[Update Closeness Centrality namespace](https://trello.com/c/gXCFL9aK/8485-update-closeness-centrality-namespace)

PR #451, opened 2023-08-03
updated docs after detiering graph procedures

yuvalr1neo:


FlorentinD:
once #450 this should go green

doc/sphinx/source/graph.rst
```diff
@@ -78,6 +78,14 @@ These all assume that an object of :class:`.GraphDataScience` is available as `g
 
     Exports a named graph into a new offline Neo4j database.
```
FlorentinD, line 79: Similar to what I suggested Jonatan. Could you add deprecation notes to the old ones? See scaleProperties as an example

PR #452, opened 2023-08-03
Propagate cypher warnings

FlorentinD:


graphdatascience/query_runner/neo4j_query_runner.py
```diff
@@ -77,7 +77,29 @@ def run_query(
...
+                    if severity == "WARNING":
+                        # FIXME handle the field deprecations
+                        if "query used a deprecated field from a procedure" in notification["description"]:
+                            continue
+
+                        if "deprecated" in notification["description"]:
+                            warning = DeprecationWarning(notification["description"])
+                        else:
+                            warning = RuntimeWarning(notification["description"])
+                        warnings.warn(warning)
```
FlorentinD, line 98: reading up on logging. `warnings.warn` should be used when users can fix it, while `logging.warn`should only be used if the user cannot change anything about the warning. (https://docs.python.org/3/howto/logging.html)

graphdatascience/query_runner/neo4j_query_runner.py
```diff
@@ -108,6 +115,22 @@ def run_query_with_logging(
             else:
                 return future.result()
 
+    def _forward_cypher_warnings(self, notification: Dict[str, Any]) -> None:
+        # (see https://neo4j.com/docs/status-codes/current/notifications/ for more details)
+        severity = notification["severity"]
+        if severity == "WARNING":
+            # FIXME handle the field deprecations
+            if "query used a deprecated field from a procedure" in notification["description"]:
+                return
```
FlorentinD, line 124: The problem is that uses cannot specify the yields upfront.
We could decide for the user to not return the deprecated fields by default, and they can set a flag to get them, but this could break code.

Alternatively we could have the warning by default and allow the user to trim down further

graphdatascience/error/cypher_warning_handler.py
```diff
@@ -0,0 +1,31 @@
+import warnings
+from functools import wraps
+from typing import Any, Callable, TypeVar, cast
+
+from ..caller_base import CallerBase
+
+F = TypeVar("F", bound=Callable[..., Any])
+
+
+def filter_id_func_deprecation_warning() -> Callable[[F], F]:
```
FlorentinD, line 10: As the id() function is central to GDS, we could also filter out the warning for every query

graphdatascience/tests/integration/test_graph_ops.py
```diff
@@ -672,17 +685,21 @@ def test_graph_relationshipProperties_stream_without_arrow_separate_property_col
...
+    else:
+        result = gds_without_arrow.beta.graph.relationships.stream(G, ["REL", "REL2"])
 
-    expected = gds_without_arrow.run_cypher(
-        "MATCH (n)-[r]->(m) RETURN id(n) AS src_id, id(m) AS trg_id, type(r) AS rel_type"
-    )
+    with pytest.raises(DeprecationWarning):
+        expected = gds_without_arrow.run_cypher(
+            "MATCH (n)-[r]->(m) RETURN id(n) AS src_id, id(m) AS trg_id, type(r) AS rel_type"
+        )
```
Mats-SX, line 696: why is this DeprecationWarning? I thought you filtered the id() warnings?

graphdatascience/graph/graph_proc_runner.py
```diff
@@ -336,7 +335,6 @@ def relationships(self) -> GraphRelationshipsRunner:
         self._namespace += ".relationships"
         return GraphRelationshipsRunner(self._query_runner, self._namespace, self._server_version)
 
-    @deprecation_warning("gds.graph.nodeProperties.stream", ServerVersion(2, 3, 0))
```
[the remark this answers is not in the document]
FlorentinD: yes :) 
Actually had to add the warnings for the arrow part

changelog.md
```diff
@@ -3,6 +3,7 @@
 
 ## Breaking changes
 
+* Forward previously ignored Cypher warnings as Python warnings. This includes for instance deprecation warnings.
```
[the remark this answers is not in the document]
FlorentinD: For people that run with `-Werror` its a breaking change. Arguably its not breaking in general though.

graphdatascience/query_runner/neo4j_query_runner.py
FlorentinD: By definition in python. `warnings.warn` is for warnings users can solve, whereas `log.warn` would be for warnings people cant fix.

IMO, we could then just log to let the user know.

graphdatascience/error/cypher_warning_handler.py
FlorentinD: As said in the comment above, I also dont think this is really needed to filter out completly.
Instead, I would consider to log this warning as there is nothing we can do about it.

However, arent we worrying users if they see this warning and there is no way they can fix it?

graphdatascience/tests/integration/test_graph_ops.py
FlorentinD: I only filtered it on our endpoints not for every `run_cypher`(YET)

Mats-SX (commented):
Dunno how my review from this morning got published, but here's the rest of it.

I'm not sure about un-deprecating some of the functions, but maybe I misunderstand some of the changes.

changelog.md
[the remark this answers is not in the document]
FlorentinD: Moved it

graphdatascience/query_runner/neo4j_query_runner.py
Mats-SX: Sounds good to me

graphdatascience/error/cypher_warning_handler.py
Mats-SX: I wonder who the warning is for. If I use GDS Client and `gds.run_cypher("MATCH (n) RETURN id(n)")` I ought to get a warning right?

graphdatascience/tests/unit/test_graph_ops.py
```diff
@@ -142,8 +142,7 @@ def test_graph_export_csv(runner: CollectingQueryRunner, gds: GraphDataScience)
 def test_graph_streamNodeProperty(runner: CollectingQueryRunner, gds: GraphDataScience) -> None:
     G, _ = gds.graph.project("g", "*", "*")
 
-    with pytest.warns(DeprecationWarning):
-        gds.graph.streamNodeProperty(G, "dummyProp", concurrency=2)
+    gds.graph.streamNodeProperty(G, "dummyProp", concurrency=2)
```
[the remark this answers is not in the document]
FlorentinD: Its not undeprecated, but the warning comes from the server now instead of the client.
So in a unit-test, we cannot assert on this warning anymore only in integration tests

changelog.md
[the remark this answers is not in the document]
Mats-SX: If we keep it as `breaking`, let's add that detail that it is only breaking if `-Werror` is set

graphdatascience/tests/integration/test_graph_ops.py
Mats-SX: I think that's good; let's not filter general queries.

In what endpoints do we explicitly use `id()`?

changelog.md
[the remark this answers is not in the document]
Mats-SX: Improvement is OK as well.

graphdatascience/tests/unit/test_graph_ops.py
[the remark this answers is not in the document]
Mats-SX: Ooooh unit testing, I get it. Cool!

graphdatascience/tests/integration/test_graph_ops.py
FlorentinD: we use it in:

* gds.graph.nodeProperties.stream if people request additional `db_node_properties`
* topological linkPrediction endpoints .. these endpoint work over node objects so we do a MATCH internally
* and gds.find_node_id() .. Used to forward to path algorithms f.i.

graphdatascience/query_runner/neo4j_query_runner.py
FlorentinD: The main downside is that the log becomes a bit noisy for the user.
Split between letting users know that we could handle this better or we first built the YIELD feature and then actually use warnings

graphdatascience/graph_data_science.py
```diff
@@ -103,8 +103,9 @@ def __init__(
         self._query_runner.set_server_version(self._server_version)
 
         if arrow and self._server_version >= ServerVersion(2, 1, 0):
+            yield_fields = "running, listenAddress" if self._server_version >= ServerVersion(2,2,1) else "running, advertisedListenAddress"
```
Mats-SX, line 106: catering for GDS versions before 2.2.1 😓

graphdatascience/query_runner/neo4j_query_runner.py
Mats-SX: Let's break the tie in favour of the conservative option and not make a change.

Mats-SX (approved):

PR #455, opened 2023-08-08
Improve error if GDS is not installed

FlorentinD:


graphdatascience/graph_data_science.py
```diff
@@ -92,6 +93,13 @@ def __init__(
         try:
             server_version_string = self._query_runner.run_query("RETURN gds.version()", custom_error=False).squeeze()
         except Exception as e:
+            if "Unknown function 'gds.version'" in str(e):
+                raise GdsNotFound(
```
FlorentinD, line 97: A further improvement could be to check if the license status is correct on `__init__`

FlorentinD:
[Improve error message if GDS is not installed](https://trello.com/c/FJLLqLAO/8568-improve-error-message-if-gds-is-not-installed)

Mats-SX (commented):
Actually, can we write a test for this case? It's a bit overkill to add a whole new setup in CI, but can't we do like a unit test? The CI positive cases captures that this does not cause issue on the "happy path", but it would be good to have a test that makes sure the "unhappy path" has the desired experience

FlorentinD:
> Actually, can we write a test for this case? It's a bit overkill to add a whole new setup in CI, but can't we do like a unit test? The CI positive cases captures that this does not cause issue on the "happy path", but it would be good to have a test that makes sure the "unhappy path" has the desired experience

Yes a whole new setup also seemed too much to me.
However a unit test is not very helpful, as I think the most likely case is that the exception changes and does not contain the str anymore.  A simpler unit test then would only check that the exception is the same (if we move the logic into a dedicated function)

PR #471, opened 2023-08-30
Add missing doc strings

FlorentinD:


graphdatascience/graph_data_science.py
```diff
@@ -173,9 +188,8 @@ def run_cypher(
         database: str
             the database on which to run the query
 
-        Returns
-        -------
-        The query result as a DataFrame
+        Returns:
+            The query result as a DataFrame
```
FlorentinD, line 192: NIT - this lead before to an odd rendering of `Return type: [...]`. Now it matches with the rest to just show `Returns:`

FlorentinD:
I did not understand why we have the `driver_config()` getter and why people should use it. @adamnsch you also noted that maybe we should skip docs around it.

FlorentinD:
[Missing ref docs for some methods on the `GraphDataScience` class](https://trello.com/c/t9MjEWC6/8667-missing-ref-docs-for-some-methods-on-the-graphdatascience-class)

adamnsch (approved):
LGTM 👍  Btw, I decided not to do this patch now to avoid merge conflicts with the bookmarks thing, but might not be so bad

adamnsch:
> I did not understand why we have the `driver_config()` getter and why people should use it. @adamnsch you also noted that maybe we should skip docs around it.

I think we added because it was useful for testing at some point, but would also allow more advanced users to see what config the driver used under the hood was configured with

FlorentinD:
I also think the rebase conflicts should be minor if any. I would hope it can be resolved automatically.

For the `driver_config()`- well then it does make sense to also document it

graphdatascience/graph_data_science.py
```diff
@@ -200,6 +200,12 @@ def run_cypher(
         return qr.run_query(query, params, database, False)
 
     def driver_config(self) -> Dict[str, Any]:
+        """
+        Get the configuration used to create the Neo4j driver.
```
adamnsch, line 204:
```suggestion
        Get the configuration used to create the underlying driver used to make queries to Neo4j.
```

PR #472, opened 2023-08-31
Fix pandas 2.1.0 related build issues 

FlorentinD:
- Assert code warns but not raises
- Silence pandas warning
- Filter pandas warning exposed from numpy
- Dont fix numpy
- Save datasets as parquet to avoid pandas read_pickle issue


graphdatascience/graph/graph_proc_runner.py
```diff
@@ -108,6 +111,11 @@ def construct(
     @client_only_endpoint("gds.graph")
     def load_cora(self, graph_name: str = "cora", undirected: bool = False) -> Graph:
         nodes = read_parquet(self._path("graphdatascience.resources.cora", "cora_nodes.parquet.gzip"))
+
+        if is_neo4j_4_driver:
+            # features is read as an ndarray which was not supported in neo4j 4
+            nodes["features"] = nodes["features"].apply(lambda x: x.tolist())
```
adamnsch, line 119: I guess we also need to handle this in the CE graph construct somehow?

adamnsch (approved):
Nice work :)

graphdatascience/graph/graph_proc_runner.py
FlorentinD: I think its not needed to generalize. When the user inputs a dataframe to graph.construct they can transform the data-type themselves.

So in this case I prefered to only do it for our datasets we self-host

PR #483, opened 2023-09-14
Improve bookmark docs

FlorentinD:
Co-authored-by: Nicola Vitucci <nicola.vitucci@neo4j.com>

### DIFF
--- a/doc/modules/ROOT/pages/bookmarks.adoc
+++ b/doc/modules/ROOT/pages/bookmarks.adoc
@@ -8,7 +8,8 @@ When using the GDS client, any query following a call to `gds.set_bookmarks()` i
 .Example of using bookmarks with GDS
 [source, python]
 ----
-# If you ran queries through Neo4j driver before you can pass bookmarks in the constructor
+# Use the `bookmarks` parameter to set any existing bookmarks, for example
+# after running a query through the Neo4j driver. The default value is `None`.
 gds = GraphDataScience(NEO4J_URI, auth=(NEO4J_USER, NEO4J_PASSWORD), bookmarks=None)
 
 gds.run_cypher("""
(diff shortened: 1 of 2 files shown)


doc/modules/ROOT/pages/bookmarks.adoc
```diff
@@ -3,13 +3,15 @@
 Bookmarks ensure that when executing queries in a Neo4j cluster, previous or parallel running queries are fully propagated through the cluster.
 By passing one or multiple bookmarks along with a query, the server will make sure that the query does not get executed before the represented state(s) have been established.
```
nvitucci, line 4:
```suggestion
A bookmark is a marker representing a state of the database.
Bookmarks are useful when subsequent transactions in a cluster need to be coordinated, for example when you need to read data right after writing it.

When using the GDS client, any query following a call to `gds.set_bookmarks()` is not executed until the bookmarked transactions are propagated across the cluster.
```

doc/modules/ROOT/pages/bookmarks.adoc
```diff
@@ -3,13 +3,15 @@
...
 
-.Example of using bookmarks inside GDS
+.Example of using bookmarks with GDS
 [source, python]
 ----
 # If you ran queries through Neo4j driver before you can pass bookmarks in the constructor
```
nvitucci, line 9: I am not sure I understand the explanation here. Do you need to specify the `bookmarks=None` parameter here, or is `None` the default and you only need to specify it if you have created any bookmarks with the driver that you want to use?

nvitucci (commented):
I've added some suggestions, hoping my understanding is correct.

doc/modules/ROOT/pages/bookmarks.adoc
FlorentinD: Awesome ⭐

doc/modules/ROOT/pages/bookmarks.adoc
FlorentinD: `None` is the default and yes, its if you created a driver instance before and now switch to the GDS client.

nvitucci: Ok, so what about rephrasing it as follows?

```suggestion
# Use the `bookmarks` parameter to set any existing bookmarks, for example
# after running a query through the Neo4j driver. The default value is `None`.
```

nvitucci (approved):

PR #496, opened 2023-10-02
Avoid remote flag for remote write-back

soerenreichardt:


soerenreichardt:
[[🍓] Track remote/local status of graphs in the Graph Catalog](https://trello.com/c/6Y3hQylD/8742-%F0%9F%8D%93-track-remote-local-status-of-graphs-in-the-graph-catalog)

graphdatascience/query_runner/aura_db_arrow_query_runner.py
```diff
@@ -79,6 +78,12 @@ def run_query(
 
         return self._fallback_query_runner.run_query(query, params, database, custom_error)
 
+    def is_remote_projected_graph(self, graph_name: str) -> bool:
+        database_location: str = self._fallback_query_runner.run_query(
+            f"CALL gds.graph.list('{graph_name}') YIELD databaseLocation RETURN databaseLocation"
```
Mats-SX, line 83: Oh this could be Cypher-injected! Imagine a graph name like `g' YIELD databaseLocation MATCH (n) DETACH DELETE n CALL gds.graph.list('g` 😅

Mats-SX: If we use a parameter instead of an fstring it would be safe.
```suggestion
            "CALL gds.graph.list($graph_name) YIELD databaseLocation RETURN databaseLocation",
            {"graph_name": graph_name}
```

soerenreichardt: damn, security flawed tests :D

FlorentinD (approved):
Looks good :) 

I think we only want to run these IT tests only against >= 2.6 (as 2.5 does not contain the lastest remote-flag changes

graphdatascience/query_runner/aura_db_arrow_query_runner.py
FlorentinD: @Mats-SX We have this multiple times in our client. 
Do we actually see this as an issue?
I would argue they only end up with executing arbitrary cypher-queries against **their own** instances.
on the cypher side, we check permissions.

So I dont think this injection is an actual attack vector here. (if it is, we should card it to fix in all places)

knutwalker: > Oh this could be Cypher-injected! Imagine a graph name like `g' YIELD databaseLocation MATCH (n) DETACH DELETE n CALL gds.graph.list('g` 😅

[Oh, yes. Little Droppy Nodes, we call him.](https://xkcd.com/327/)

PR #504, opened 2023-10-16
Aura sessions

FlorentinD:


FlorentinD:
[[🍓] Build GDS Client API for managing the "GDS Session"](https://trello.com/c/pgTSyvPv/8743-%F0%9F%8D%93-build-gds-client-api-for-managing-the-gds-session)

FlorentinD:
checkstyle failure indicates, we need to update the docker image to include requests

Mats-SX:
There is a commit `WIP aura connect` in the list. At the very least we should rename the commit

Mats-SX:
I also think we should add a comprehensive changelog entry. We're bringing in lots of features here.

FlorentinD:
> I also think we should add a comprehensive changelog entry. We're bringing in lots of features here. I don't think we have changelogged the remote projection endpoint previously, so should include that as well.

and potentially drop the changelogs if we dont want to publicly announce it as part of 1.9?

Mats-SX:
AFAIK, the only features we have currently planned for 1.9 are these.

graphdatascience/aura_api.py
```diff
@@ -0,0 +1,218 @@
...
+            return None
+
+        response.raise_for_status()
+
+        raw_data = response.json()["data"]
+
+        return InstanceSpecificDetails.fromJson(raw_data)
+
+    def wait_for_instance_running(self, instance_id: str, sleep_time: float = 0.2) -> bool:
+        while sleep_time < AuraApi.MAX_WAIT_TIME:
```
soerenreichardt, line 168: The `sleep_time`variable is never updated resulting in a possibly endless waiting loop.

graphdatascience/aura_sessions.py
```diff
@@ -0,0 +1,129 @@
...
+        if len(session_password) < 8:
+            raise ValueError("Password must be at least 8 characters long.")
+
+        if session_name in [session.name for session in self.list_sessions()]:
+            raise RuntimeError(f"Session with name `{session_name}` already exists.")
+
+        create_details = self._aura_api.create_instance(AuraSessions._instance_name(session_name))
+        self._aura_api.wait_for_instance_running(create_details.id)
+
+        gds_user = create_details.username
```
soerenreichardt, line 45: Good practice to prefix the usernames with where they belong too. I guess it can get very confusing with the number of credentials and usernames we are dealing with :)

graphdatascience/aura_sessions.py
```diff
@@ -0,0 +1,129 @@
...
+        url = create_details.connection_url
```
soerenreichardt, line 46: nit: could also be `gds_url`

graphdatascience/aura_sessions.py
```diff
@@ -0,0 +1,129 @@
...
+        instance_details = self._aura_api.list_instance(matched_instances[0])
+
+        if instance_details:
+            gds_url = instance_details.connection_url
+        else:
+            raise RuntimeError(
+                f"Unable to get connection information for session `{session_name}`. Does it still exist?"
+            )
+
+        # Hardcoded neo4j user as sessions are always created with this user
```
soerenreichardt, line 72: Why is that the case? Maybe it would be good if it was the gds_user and we would check immediately if the user is elegible of querying the database to throw early if it lacks permissions.

graphdatascience/aura_sessions.py
```diff
@@ -0,0 +1,129 @@
...
+            candidates = [(i.id, i.name) for i in candidate_instances]
+            raise RuntimeError(
+                f"Expected to find exactly one Aura instance with name `{instance_name}`, but found `{candidates}`."
+            )
+
+        if len(candidate_instances) == 1:
+            candidate = candidate_instances[0]
+            return self._aura_api.delete_instance(candidate.id) is not None
+
+        return False
```
soerenreichardt, line 98: Isn't this case also an error if finding more than 1 session results in an error? Otherwise I feel like the API here is mixing booleans with exceptions to convey what is going wrong or not.

graphdatascience/tests/unit/test_aura_api.py
```diff
@@ -0,0 +1,186 @@
...
+    mock_auth_token(requests_mock)
+    requests_mock.delete(
+        "https://api.neo4j.io/v1/instances/id0",
+        status_code=404,
+        reason="Not Found",
+        json={"errors": [{"message": "DB not found: id0", "reason": "db-not-found"}]},
+    )
+
+    result = api.delete_instance("id0")
+    assert result is None
```
soerenreichardt, line 50: it would be nice to show to the user what the problem was

graphdatascience/tests/unit/test_aura_api.py
```diff
@@ -0,0 +1,186 @@
...
+    mock_auth_token(requests_mock)
+
+    requests_mock.get(
+        "https://api.neo4j.io/v1/instances/id0",
+        status_code=404,
+        reason="Not Found",
+        json={"errors": [{"message": "DB not found: id0", "reason": "db-not-found"}]},
+    )
+
+    assert api.list_instance("id0") is None
```
soerenreichardt, line 179: see above, an indication for the user might be useful

soerenreichardt (approved):
Very nice work! I spotted a bug however and have a few smaller comments

soerenreichardt (changes requested):
The instance wait methods can get stuck in an endless loop

graphdatascience/aura_api.py
```diff
@@ -0,0 +1,218 @@
...
+        raw_data = response.json()["data"]
+
+        if len(raw_data) != 1:
+            raise RuntimeError(
+                f"This account has access to multiple tenants `{raw_data}`. Please specify which one to use."
+            )
+
+        return raw_data[0]["id"]  # type: ignore
+
+    def _get_request_headers(self) -> dict[str, str]:
```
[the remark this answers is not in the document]
Mats-SX: I will call it `_build_header()` because there are no other headers around.

graphdatascience/aura_sessions.py
Mats-SX: The reason is that we are creating a new DBMS instance, which always comes with one default user called `neo4j`. We are setting the password to our user's input password (for reconnecting), but I don't see why we should also allow them to specify a username. The session name serves the name-identity function now. Sure, they _can_ log in and do stuff like `CREATE USER`, but do we really need to support that when we are wanting to remove the DBMS, including system db and its RBAC?

Mats-SX: Indeed this is the `gds_user` and only used for logging into the GDS session (AuraDS instance).

graphdatascience/aura_sessions.py
Mats-SX: I agree that there is a bit of a mix here. But there is a difference: when there are more than 1 instances this is a bit of an unexpected situation. We cannot forbid our users to rename their AuraDS instances, but we can discourage them from trying to copy our naming pattern. If they still do, this can happen and then it is a situation that arises from bad usage and should result in an error because there's no way to recovery other than manual intervention (e.g. renaming).

But if we find zero sessions that match the name, the situation is a bit different. First, we have the common idempotency characteristic: deleting the same thing twice needn't be an error. If you view the semantics of deletion as "after this operation the resource should not exist" then it is successful behaviour to do nothing if the resource didn't even exist in the first place. We indicate the difference between this and _actually_ deleting in the return value.
Secondly, it is possible that, in a race situation, we successfully find the session instance we should delete, but in the time it takes for us to make the call to the Aura API, the instance was already deleted by another request. Similar to above, we don't need to fail because if it is deleted then we're successful. So we interpret the `404` we will get on the Aura API request as success with a `False`.

Personally, I prefer this design because we don't have to bother our users when they try to delete things that don't exist; compare the `failIfMissing` flag we added to `graph.drop` because users want to have automated scripts that work idempotently.

graphdatascience/tests/unit/test_aura_api.py
Mats-SX: IMO this is not a problem. They are deleting a session that's already being deleted. Things are fine.

graphdatascience/tests/unit/test_aura_api.py
Mats-SX: This is not a user-facing method. We use it from two places currently:

- `connect`, which does raise an error when the to-be-connected session is not found
- `wait_for_instance_running` which is called from `create_gds` and there we just ignore the return value. That's a bug that I will fix on this PR.

I suggest making the behaviour such that if we get `None` back then we will raise an error indicating that there was a failure in creating the session.

I don't think we need to change how `list_instance()` works though.

graphdatascience/aura_api.py
Mats-SX: nice find! there's no unit test for this method 🤦

graphdatascience/aura_sessions.py
soerenreichardt: I was actually thinking about the username we issue the arrow requests from, but I guess that one is retrieved from the token which we acquire from the actual username that is used in AuraDB. So it should be fine with the default usename here

graphdatascience/aura_api.py
Mats-SX: Added a few tests now

soerenreichardt (approved):

PR #507, opened 2023-10-17
Simplify nav menu structure

nvitucci:
Proposing a new structure for the nav menu:

- All items one level up (so all at the same level)
- Index page renamed and simplified
- ~Made "Tutorials" a clickable dropdown instead of a page with content being just the list of tutorials~
- Made "Tutorials" a partial to avoid repetition


doc/modules/ROOT/pages/index.adoc
```diff
@@ -1,4 +1,4 @@
-= The Neo4j Graph Data Science Client Manual v{docs-version}
+= Introduction
```
FlorentinD, line 1: was this maybe also for seo? / should we add keywords?

doc/modules/ROOT/pages/tutorials/tutorials.adoc
```diff
@@ -1,16 +0,0 @@
-= Tutorials
```
FlorentinD, line 1: this page was likely referenced. We should have a redirect .. but to where?

FlorentinD (commented):
I like the removed intendation, but hesitant about the removal of the tutorials page

doc/modules/ROOT/pages/tutorials/tutorials.adoc
nvitucci: I've brought back the page with a change to avoid the repetition of the list of tutorials.

doc/modules/ROOT/pages/index.adoc
nvitucci: We don't _have_ to, but I agree it's clearer to keep a reference to the product in the title. We can still use a different name for the menu item. WDYT?

FlorentinD: i like your suggestion to keep the title but simplify the menu item 👍

FlorentinD (approved):
Thanks for this improvement :)

PR #510, opened 2023-10-26
Update arrow remote db tests

FlorentinD:


FlorentinD:
merging as the remaining failure needs to be fixed on the server side

PR #512, opened 2023-10-27
Handle new license function + AuraDS

FlorentinD:


FlorentinD:
[New license functions not callable in AuraDS](https://trello.com/c/XiEQp1UY/8872-new-license-functions-not-callable-in-aurads)

graphdatascience/system/system_endpoints.py
```diff
@@ -53,10 +61,7 @@ def is_licensed(self) -> bool:
...
         except Exception as e:
             # AuraDS does not have `gds.debug.sysInfo`, but is always GDS EE.
-            if (
-                "There is no procedure with the name `gds.debug.sysInfo` "
-                "registered for this database instance." in str(e)
-            ):
+            if re.match(r"There is no procedure with the name `gds.*` registered for this database instance", str(e)):
                 isLicensed = True
             else:
                 raise e
```
adamnsch, line 67: (Unrelated to this PR) I wonder if we should actually raise better errors here. Since the user won't know or need to know that `sysInfo` is used under the hood

changelog.md
```diff
@@ -11,7 +11,7 @@
 
 ## Bug fixes
 
-* 
+* Fixed a bug which would throw on running gds.license.state against an AuraDS instance.
```
adamnsch, line 14:
```suggestion
* Fixed a bug which caused an exception to be raised when running `gds.license.state` targeting an AuraDS instance.
```

adamnsch (approved):
Nice!

adamnsch:
Needs a rebase due to the build issues on main previously

graphdatascience/system/system_endpoints.py
FlorentinD: in practice it should show that the instance is misconfigured 🤔 
(special case for AuraDS as their its configured like this on purpose)

But maybe its also in an on-prem version prevented by the admin to not run some functions

PR #513, opened 2023-10-27
Add notebook to docs check

FlorentinD:


scripts/nb2doc/check_nb_docs.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
```
nvitucci, line 1: As discussed, I'd rather run the existing script then check if any differences have been introduced using `git`, for example running `git diff --quiet doc/modules/ROOT/pages/tutorials/` and checking its exit code.

nvitucci (approved):

PR #521, opened 2023-11-07
Reword explanation of graph construct

nvitucci:
We got the following feedback on the [Graph object](https://neo4j.com/docs/graph-data-science-client/current/graph-object/) page.

> The example of creating a graph from cypher as equivalent to what gds.graph.construct() is achieving is confusing because it mislead me to believe that the method creates data in the DATABASE, instead of a GRAPH PROJECTION.
This is misleading because you wouldn't use CREATE statements to generate data directly in a graph projection.
More clarity that this actually takes a dataframe and creates a projection would have saved me a lot of time!

This PR tries to address this with a better explanation.

### DIFF
--- a/doc/modules/ROOT/pages/graph-object.adoc
+++ b/doc/modules/ROOT/pages/graph-object.adoc
@@ -145,7 +145,7 @@ assert G.node_count() == result["nodeCount"]
 [[construct]]
 == Constructing a graph from DataFrames
 
-In addition to projecting a graph from the Neo4j database, it is also possible to create in-memory graphs directly from pandas `DataFrame` objects.
+In addition to projecting a graph from the Neo4j database, it is also possible to create graphs directly from pandas `DataFrame` objects.
 
 === Syntax
 .Graph construct signature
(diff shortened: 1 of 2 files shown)


doc/modules/ROOT/pages/graph-object.adoc
```diff
@@ -145,7 +145,7 @@ assert G.node_count() == result["nodeCount"]
 [[construct]]
 == Constructing a graph from DataFrames
 
-Instead of projecting a graph from the Neo4j database it is also possible to construct new graphs using pandas `DataFrames` from the client.
+In addition to projecting a graph from the Neo4j database, it is also possible to create in-memory graphs directly from pandas `DataFrame` objects.
```
adamnsch, line 148: I think it was decided by someone at some point that we should refer to GDS graphs as "projections" rather than the less abstract "in-memory graphs". Personally I don't have an opinion on the subject

adamnsch (approved):
LGTM 👍

doc/modules/ROOT/pages/graph-object.adoc
nvitucci: I think in this case I can actually omit the "in-memory" part and just leave it as "to create graphs", if that's ok. 

More in general I thought a "projection" should be the operation of creating a graph rather than the graph itself, but I'm happy to discuss.

nvitucci:
@FlorentinD FYI I had to slightly update the notebook `check.sh` script or it wouldn't run on TeamCity.

PR #525, opened 2023-11-08
Derive cloud provider from db

FlorentinD:


FlorentinD:
[[🍓] Build GDS Client API for managing the "GDS Session"](https://trello.com/c/pgTSyvPv/8743-%F0%9F%8D%93-build-gds-client-api-for-managing-the-gds-session)

graphdatascience/aura_api.py
```diff
@@ -91,7 +92,16 @@ def __init__(self, client_id: str, client_secret: str, tenant_id: Optional[str]
         self._logger = logging.getLogger()
         self._tenant_id = tenant_id if tenant_id else self._get_tenant_id()
 
-    def create_instance(self, name: str) -> InstanceCreateDetails:
+    @classmethod
+    def extract_id(cls, uri: str) -> str:
```
Mats-SX, line 96: Since we don't use the `cls` variable, I suggest we use `@staticmethod` instead. It seems more appropriate.

graphdatascience/aura_api.py
```diff
@@ -91,7 +92,16 @@ def __init__(self, client_id: str, client_secret: str, tenant_id: Optional[str]
...
+        host = urlparse(uri).hostname
+
+        if not host:
+            raise RuntimeError(f"Could not parse the uri {uri}.")
```
Mats-SX, line 100: I like wrapping user-input values in backticks, especially if we failed to parse it. could contain odd whitespaces for instance.
```suggestion
            raise RuntimeError(f"Could not parse the uri `{uri}`.")
```

graphdatascience/aura_api.py
```diff
@@ -91,7 +92,16 @@ def __init__(self, client_id: str, client_secret: str, tenant_id: Optional[str]
...
+
+        return host.split(".")[0].split("-")[0]
```
Mats-SX, line 102: there is a risk that this parsing fails, right? array index out of bounds etc. or does `urlparse` guarantee that the output supports this splitting? we should add unit tests and if we can write one that fails here then we should catch and throw

graphdatascience/aura_sessions.py
```diff
@@ -38,7 +38,13 @@ def create_gds(self, session_name: str, session_password: str) -> GraphDataScien
         if session_name in [session.name for session in self.list_sessions()]:
             raise RuntimeError(f"Session with name `{session_name}` already exists.")
 
-        create_details = self._aura_api.create_instance(AuraSessions._instance_name(session_name))
+        db_instance_id = AuraApi.extract_id(self._db_credentials.uri)
+        db_instance = self._aura_api.list_instance(db_instance_id)
+        if not db_instance:
+            raise ValueError(f"Could not find Aura instance with the url {self._db_credentials.uri}")
```
Mats-SX, line 44:
```suggestion
            raise ValueError(f"Could not find Aura instance with the uri `{self._db_credentials.uri}`")
```

Mats-SX (approved):
My comments are not blocking, but could be used to gain a small improvement

graphdatascience/aura_api.py
FlorentinD: urlparse returns a string and as we always takes the very first after the split, at worst you endup with an invalid string `id`. But I dont think we want to check constraints on the id.

added some test though :)

PR #534, opened 2023-11-14
Do not fail when `type` key is missing

Mats-SX:
It can happen that this key is missing from the response body. It doesn't matter so much for us when this happens during `create` so we can be soft about it.

### DIFF
--- a/graphdatascience/aura_api.py
+++ b/graphdatascience/aura_api.py
@@ -71,7 +71,7 @@ class InstanceCreateDetails(InstanceDetails):
             password=json["password"],
             username=json["username"],
             connection_url=json["connection_url"],
-            type=json["type"],
+            type=json.get("type", ""),
             region=json["region"],
         )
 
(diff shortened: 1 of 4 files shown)


graphdatascience/aura_api.py
```diff
@@ -54,26 +55,19 @@ def fromJson(cls, json: dict[str, Any]) -> InstanceSpecificDetails:
 
 
 @dataclass(repr=True)
-class InstanceCreateDetails(InstanceDetails):
-    password: str
+class InstanceCreateDetails:
+    id: str
     username: str
+    password: str
     connection_url: str
```
FlorentinD, line 62: so we only look at the fields we actually need and not everything that should be returned?

FlorentinD (approved):
Slimming the create result is a good workaround.

graphdatascience/aura_api.py
Mats-SX: Exactly. This protects us from the flakiness of the `region`, `name`, and `type` fields, which we do not need.

PR #535, opened 2023-11-14
Interact with aura through specific gds api object

soerenreichardt:

Make sure:

### DIFF
--- a/graphdatascience/aura_graph_data_science.py
+++ b/graphdatascience/aura_graph_data_science.py
@@ -1,0 +1,81 @@
+from typing import Any, Dict, Optional, Tuple
+
+from neo4j import GraphDatabase
+from pandas import DataFrame
+
+from .query_runner.neo4j_query_runner import Neo4jQueryRunner
+from .server_version.server_version import ServerVersion
+from .version import __version__
+from graphdatascience import GraphDataScience
+from graphdatascience.query_runner.aura_db_arrow_query_runner import (
+    AuraDbArrowQueryRunner,
+    AuraDbConnectionInfo,
+)
+
+
+class AuraGraphDataScience(GraphDataScience):
+    """
+    Primary API class for interacting with Neo4j AuraDB + Graph Data Science.
+    Always bind this object to a variable called `gds`.
+    """
+
+    def __init__(
+        self,
+        endpoint: str,
+        auth: Tuple[str, str],
+        aura_db_connection_info: AuraDbConnectionInfo,
+        database: Optional[str] = None,
+        arrow_disable_server_verification: bool = True,
+        arrow_tls_root_certs: Optional[bytes] = None,
+        bookmarks: Optional[Any] = None,
+    ):
+        super().__init__(
+            endpoint=endpoint,
+            auth=auth,
+            aura_ds=True,
+            database=database,
+            arrow=True,
+            arrow_disable_server_verification=arrow_disable_server_verification,
+            arrow_tls_root_certs=arrow_tls_root_certs,
+            bookmarks=bookmarks,
+        )
+
+        gds_query_runner = self._query_runner
+
+        driver = GraphDatabase.driver(aura_db_connection_info.uri, auth=aura_db_connection_info.auth, **self._config)
+        db_query_runner = Neo4jQueryRunner(driver, auto_close=True, bookmarks=bookmarks)
+        db_query_runner.set_server_version(self._server_version)
+
+        if database:
+            db_query_runner.set_database(database)
+
+        if self._server_version >= ServerVersion(2, 6, 0):
+            self._query_runner = AuraDbArrowQueryRunner(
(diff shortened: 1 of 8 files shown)


soerenreichardt:
[[🍓] Client behaviour for cypher-based GDS features (run_cypher, native_projection)](https://trello.com/c/D4OUpeYH/8931-%F0%9F%8D%93-client-behaviour-for-cypher-based-gds-features-runcypher-nativeprojection)

graphdatascience/aura_graph_data_science.py
```diff
@@ -0,0 +1,80 @@
...
+    Always bind this object to a variable called `gds`.
+    """
+
+    def __init__(
+        self,
+        endpoint: str,
+        auth: Tuple[str, str],
+        aura_db_connection_info: AuraDbConnectionInfo,
+        database: Optional[str] = None,
+        arrow_disable_server_verification: bool = True,
```
DarthMax, line 27: Should be false by default.

graphdatascience/aura_graph_data_science.py
```diff
@@ -0,0 +1,80 @@
...
+from .query_runner.neo4j_query_runner import Neo4jQueryRunner
+from .server_version.server_version import ServerVersion
+from graphdatascience import GraphDataScience
+from graphdatascience.query_runner.aura_db_arrow_query_runner import (
+    AuraDbArrowQueryRunner,
+    AuraDbConnectionInfo,
+)
+
+
+class AuraGraphDataScience(GraphDataScience):
```
DarthMax, line 15: We talked about the set_database method yesterday. Should we also change the semantics here?

graphdatascience/query_runner/aura_db_arrow_query_runner.py
```diff
@@ -64,12 +65,16 @@ def run_query(
...
 
+        query_runner = self._gds_query_runner
+
         if AuraDbArrowQueryRunner.GDS_REMOTE_PROJECTION_PROC_NAME in query:
             token, aura_db_arrow_endpoint = self._get_or_request_auth_pair()
             params["token"] = token
             params["host"] = aura_db_arrow_endpoint
             params["config"] = {"useEncryption": self._encrypted}
 
+            query_runner = self._db_query_runner
```
DarthMax, line 76: This query needs to run in GDS not the DB

graphdatascience/tests/integration/conftest.py
```diff
@@ -24,7 +26,7 @@
 
 DB = os.environ.get("NEO4J_DB", "neo4j")
 
-AURA_DB_URI = os.environ.get("NEO4J_AURA_DB_URI", "bolt://localhost:7689")
+AURA_DB_URI = os.environ.get("NEO4J_AURA_DB_URI", "bolt//localhost:7689")
```
DarthMax, line 29: I don't think this is a valid URI

DarthMax (approved):
Looks good
I only have some minor comments

graphdatascience/aura_graph_data_science.py
soerenreichardt: What exactly do you mean here? the `set_database` and `set_bookmark` calls are delegated to the `AuraDbQueryRunner` here which then delegates to an `_db_query_runner`.

Mats-SX, line 15: Did we consider to compose instead of inherit? We do a lot of inheritance in the client, but maybe this is a place where we would be less limited if we didn't have to support all the same signatures?

graphdatascience/graph_data_science.py
```diff
@@ -39,7 +35,6 @@ def __init__(
         arrow: bool = True,
         arrow_disable_server_verification: bool = True,
         arrow_tls_root_certs: Optional[bytes] = None,
-        aura_db_connection_info: Optional[AuraDbConnectionInfo] = None,
```
Mats-SX, line 42: I like removing this. Technically it is breaking, but nobody could've used this for anything valuable yet anyway. The mistake is not removing it now, but adding it previously and releasing that.

graphdatascience/aura_graph_data_science.py
soerenreichardt: We did consider it, however we need the `Neo4jQueryRunner` and the `ArrowQueryRunner` from the `GraphDataScience` object. If we copy the code we basically copy almost the whole constructor. If we instantiate a gds object instead of extending it we still have to provide the same arguments.
I'm also not super happy with the initalization right now but is was the easiest for now. One other thing would be to disassemble the `GraphDataScience` constructor a bit and pulling things into own classes that we could reuse in the `AuraGraphDataScience` object.

soerenreichardt: See #539

Mats-SX: wow, I love it!

PR #536, opened 2023-11-15
Implement aura instance size selection

soerenreichardt:


soerenreichardt:
[[🍓] Improve instance sizing API](https://trello.com/c/E6Au6njD/8926-%F0%9F%8D%93-improve-instance-sizing-api)

DarthMax (approved):
Looks good, just one question: what happens if the user enters an illegal size string?

soerenreichardt:
> Looks good, just one question: what happens if the user enters an illegal size string?

Good question, there seems to be a way to get available memory configurations. I will try to check if the option is included there

PR #537, opened 2023-11-15
Call_procedure/function endpoints for better checks on gds calls

FlorentinD:
- Add call_procedure/functions above run_cypher
- Fix test_edge_embedding_models query building

### DIFF
--- a/graphdatascience/query_runner/arrow_query_runner.py
+++ b/graphdatascience/query_runner/arrow_query_runner.py
@@ -11,7 +11,7 @@ from pyarrow.flight import ClientMiddleware, ClientMiddlewareFactory
 from ..server_version.server_version import ServerVersion
 from .arrow_graph_constructor import ArrowGraphConstructor
 from .graph_constructor import GraphConstructor
-from .query_runner import QueryRunner
+from .query_runner import EndpointType, QueryRunner
 from graphdatascience.server_version.compatible_with import (
     IncompatibleServerVersionError,
 )
@@ -52,9 +52,21 @@ class ArrowQueryRunner(QueryRunner):
             DeprecationWarning(f"The endpoint '{old_endpoint}' is deprecated. Please use '{new_endpoint}' instead.")
         )
 
-    def run_query(
+    def run_cypher(
         self,
         query: str,
+        params: Optional[Dict[str, Any]] = None,
+        database: Optional[str] = None,
+        custom_error: bool = True,
+    ) -> DataFrame:
+        return self._fallback_query_runner.run_cypher(query, params, database, custom_error)
+
+    def call_endpoint(
+        self,
+        type: EndpointType,
+        endpoint: str,
+        yields: Optional[List[str]] = None,
+        body: Optional[str] = None,
         params: Optional[Dict[str, Any]] = None,
         database: Optional[str] = None,
         custom_error: bool = True,
@@ -66,7 +78,9 @@ class ArrowQueryRunner(QueryRunner):
         no_tier_in_namespace_server_version = ServerVersion(2, 5, 0)
 
         # We need to support the deprecated endpoints until they get removed on the server side
-        if (old_endpoint := ("gds.graph.streamNodeProperty" in query)) or "gds.graph.nodeProperty.stream" in query:
+        if (
+            old_endpoint := ("gds.graph.streamNodeProperty" == endpoint)
+        ) or "gds.graph.nodeProperty.stream" == endpoint:
             graph_name = params["graph_name"]
             property_name = params["properties"]
             node_labels = params["entities"]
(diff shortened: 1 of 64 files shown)


FlorentinD:
inspired by #214

FlorentinD:
[[🍓] Client behaviour for cypher-based GDS features (run_cypher, native_projection)](https://trello.com/c/D4OUpeYH/8931-%F0%9F%8D%93-client-behaviour-for-cypher-based-gds-features-runcypher-nativeprojection)

graphdatascience/query_runner/query_runner.py
```diff
@@ -7,9 +8,65 @@
 from .graph_constructor import GraphConstructor
 
 
+class EndpointType(Enum):
+    PROCEDURE = "PROCEDURE"
+    FUNCTION = "FUNCTION"
+
+
 class QueryRunner(ABC):
+    def call_function(
```
adamnsch, line 17: (nitpick) Might be better to have a bit more elaborate name, since it sounds like it's just calling a Python function

graphdatascience/query_runner/query_runner.py
```diff
@@ -7,9 +8,65 @@
...
+            endpoint=endpoint,
+            yields=yields,
+            body=body,
+            params=params,
+            database=database,
+            custom_error=custom_error,
+        )
+
+    @abstractmethod
+    def call_endpoint(
```
adamnsch, line 56: It looks like this method is not really needed, if the only difference from the other `call_*` is that it takes a type parameter. And you could also remove the enum I think

graphdatascience/query_runner/arrow_query_runner.py
```diff
@@ -52,12 +52,24 @@ def warn_about_deprecation(self, old_endpoint: str, new_endpoint: str) -> None:
...
         custom_error: bool = True,
+    ) -> DataFrame:
+        return self._fallback_query_runner.run_cypher(query, params, database, custom_error)
+
+    def call_endpoint(
+        self,
+        type: EndpointType,
+        endpoint: str,
+        yields: Optional[List[str]] = None,
+        body: Optional[str] = None,
```
adamnsch, line 69: I wonder if `body` is a good name. It's not really a body in the conventional sense of a function body, or if statement body. It looks more like `param_placeholders` or something to me

graphdatascience/query_runner/neo4j_query_runner.py
```diff
@@ -81,14 +81,35 @@ def run_query(
...
+        yields: Optional[List[str]] = None,
+        body: Optional[str] = None,
+        params: Optional[Dict[str, Any]] = None,
+        database: Optional[str] = None,
+        custom_error: bool = True,
+    ) -> DataFrame:
+        call_keyword = "CALL" if type == EndpointType.PROCEDURE else "RETURN"
+
+        if yields is not None and type == EndpointType.FUNCTION:
+            raise ValueError("Functions cannot yield results")
```
adamnsch, line 97: You'll need to specify what you mean by "function" here. Also, a client user may not know what a Cypher function is. If you don't put yields in the API yet, I think this should be a generic `RuntimeError`

graphdatascience/system/config_endpoints.py
```diff
@@ -14,15 +14,14 @@ def set(self, key: str, value: Any, username: Optional[str] = None) -> None:
         self._namespace += ".set"
 
         params = {"key": key, "value": value}
+        body = "$key, $value"
 
         # Checking for explicit None as '' is a valid user
         if username is not None:
-            query = f"CALL {self._namespace}($key, $value, $username)"
+            body += ", $username"
             params["username"] = username
```
adamnsch, line 22: Somehow I think this feels a little bit hacky, with the Cypher parameter placeholder given as a string like this. I think it would be better to have some kind of builder-like class `CypherParamPlaceholders` that is given to the query runner. In this method:
```python
cpp = CypherParamPlaceholders()
cpp.add("key")
...
self._query_runner(...,cpp=cpp,...)
```
Then the in the query runner:
`f"CALL {endpoint}({cpp.build()}) ..."`
or something like this

adamnsch: (just an example, might not be the best way)

adamnsch (commented):
Looks promising! This is a big undertaking, thanks for doing it :)

graphdatascience/query_runner/query_runner.py
FlorentinD: users of query_runner use call_proc/function but implementers only need call_endpoint. Renaming to `_call_endpoint` could make this clearer

graphdatascience/query_runner/arrow_query_runner.py
FlorentinD: I was also not hapy with this. At first i head `positional_args`. thinking about it more though, it is indeed in most cases a 1:1 mapping with the keys in params 🤔 
Should add a check if thats correct

graphdatascience/query_runner/query_runner.py
adamnsch: Sure, but if the enum is needed as an argument, it suggests that the different types need to be handled in different ways anyway

graphdatascience/system/config_endpoints.py
FlorentinD: I went with `CallParameters` which is an ordered dict as a builder

graphdatascience/query_runner/query_runner.py
FlorentinD: short-cutted call_function to run_cypher as its always ended up there

adamnsch (approved):
Looks good, nice work!!

PR #543, opened 2023-11-21
Fix bugs after GDS creation refactoring

DarthMax:


graphdatascience/aura_graph_data_science.py
```diff
@@ -26,43 +26,41 @@ def __init__(
...
-        self._driver_config = gds_query_runner.driver_config()
+        gds_query_runner = ArrowQueryRunner.create(
+            Neo4jQueryRunner.create(endpoint, auth, aura_ds=True),
+            auth,
+            True,
+            arrow_disable_server_verification,
+            arrow_tls_root_certs,
+        )
+
+        gds_query_runner.set_database("neo4j")
```
soerenreichardt, line 41: This is not necessary as this is the default

graphdatascience/aura_graph_data_science.py
```diff
@@ -26,43 +26,41 @@ def __init__(
         endpoint: str,
         auth: Tuple[str, str],
         aura_db_connection_info: AuraDbConnectionInfo,
-        database: Optional[str] = None,
```
soerenreichardt, line 29: Why do you think that this is not necessary anymore? I think we need this

soerenreichardt (changes requested):
I think agree with removing the database selection. Otherwise this looks fine

graphdatascience/aura_graph_data_science.py
FlorentinD: i dont see where this is the default. The default is `None`.

soerenreichardt: https://github.com/neo4j/graph-data-science-client/blob/main/graphdatascience/query_runner/neo4j_query_runner.py#L83

soerenreichardt: Still necessary as the default is `None` and not `neo4j` which is a bit counter-intuitive. I added a comment clarifying this

soerenreichardt (approved):

soerenreichardt:
@DarthMax we thought a bit about how to set the database and so on and preferred to align the remote projection with what other procedures do by having to call `gds.set_database` instead of passing that as an explicit argument to the procedure

PR #546, opened 2023-11-23
Allow gds.graph.drop by graph name

FlorentinD:
Before you had to check gds.graph.exists upfront, as you cannot get a Graph object if the graph does not exists.
This makes it a lot shorter and usage of `failIfMissing=False` parameter actually useful from the client.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -17,6 +17,7 @@
 ## Improvements
 
 * Expose user facing custom types so that they can be directly imported from `graphdatascience`.
+* Allow dropping graphs through `gds.graph.drop` by name and not only based on `Graph` objects.
 
 
 ## Other changes
--- a/graphdatascience/graph/graph_proc_runner.py
+++ b/graphdatascience/graph/graph_proc_runner.py
@@ -294,18 +294,20 @@ class GraphProcRunner(UncallableNamespace, IllegalAttrChecker):
         self._namespace += ".ogbl"
         return OGBLLoader(self._query_runner, self._namespace, self._server_version)
 
-    @graph_type_check
     def drop(
         self,
-        G: Graph,
+        graph: Union[Graph, str],
         failIfMissing: bool = False,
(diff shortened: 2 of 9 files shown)


adamnsch:
This has come up several times before. The reason not do to it I think, is that we don't want people to represent graphs by strings in their code. I think it's still valid

FlorentinD:
in general I do agree, but for `drop`, I see a good excuse, as otherwise you have no way to drop your graph except with `run_cypher`

adamnsch:
> in general I do agree, but for `drop`, I see a good excuse, as otherwise you have no way to drop your graph except with `run_cypher` (or always check exists upfront, which makes the failIfMissing flag not really useful)

The `failIfMissing` flag is really something built for the Cypher interface I think, so that it's not very useful for the Python client I think is ok. The Pythonic way of doing the drop would be:
```python
try:
    gds.graph.get("...").drop()
except:
    pass
```
which I prefer. But it's up to you of course. Hopefully people should rarely end up in a situation where they don't have a graph object reference when using the Python API

FlorentinD:
I did not think of the try + except too much. Not yet thinking pythonic enough you could say.
For that argument, I would update our usage in the test then.

My main motivation is to reduce the `run_cypher` usage for gds endpoints

adamnsch:
> I did not think of the try + except too much. Not yet thinking pythonic enough you could say. For that argument, I would update our usage in the test then.
> 
> My main motivation is to reduce the `run_cypher` usage for gds endpoints

Haha :) It takes some time to get into the mindset I guess. Anyway, if you prefer the solution you suggested I think you should merge it. It's up to you. I think @Mats-SX has also been part of discussions on this topic in the past, so he might have an opinion

Mats-SX:
I'm OK with this. I see a lot of users struggle with how to remove graphs and come up with creative ways. This seems straight-forward to me.

PR #548, opened 2023-11-28
Prepare notebook for field testing

Mats-SX:


FlorentinD (approved):
LGTM

examples/dev/aura-only-features.ipynb
```diff
@@ -387,7 +387,7 @@
     "    nodeSelfInfluence=0.2,\n",
     ")\n",
     "print(f\"Compute millis: {frp_result['computeMillis']}\")\n",
-    "gds.graph.nodeProperties.stream(G, ['pagerank', 'fastRP'], separate_property_columns=True)"
+    "gds.graph.nodeProperties.stream(G, [\"pagerank\", \"fastRP\"], separate_property_columns=True)"
```
Mats-SX: I don't get why escaped double-quotes are stylistically preferred over single-quotes. I guess I need to write a PEP or smth

Mats-SX: @adamnsch do you think this is reasonable?

Mats-SX: or wait, is the escape only in the notebook source, and the backslashes go away when rendering?

PR #556, opened 2023-12-13
AuraDS size estimation tool

soerenreichardt:
~~Based on https://github.com/neo4j/graph-data-science-client/pull/629~~

### DIFF
--- a/graphdatascience/session/aura_api.py
+++ b/graphdatascience/session/aura_api.py
@@ -12,6 +12,7 @@ from urllib.parse import urlparse
 import requests as req
 from requests import HTTPError
 
+from graphdatascience.session.algorithm_category import AlgorithmCategory
 from graphdatascience.version import __version__
 
 
@@ -64,6 +65,21 @@ class InstanceCreateDetails:
 
     @classmethod
     def from_json(cls, json: dict[str, Any]) -> InstanceCreateDetails:
+        fields = dataclasses.fields(cls)
+        if any(f.name not in json for f in fields):
+            raise RuntimeError(f"Missing required field. Expected `{[f.name for f in fields]}` but got `{json}`")
+
+        return cls(**{f.name: json[f.name] for f in fields})
+
+
+@dataclass(repr=True, frozen=True)
+class EstimationDetails:
+    min_required_memory: str
+    recommended_size: str
+    did_exceed_maximum: bool
+
+    @classmethod
+    def from_json(cls, json: dict[str, Any]) -> EstimationDetails:
         fields = dataclasses.fields(cls)
         if any(f.name not in json for f in fields):
             raise RuntimeError(f"Missing required field. Expected `{[f.name for f in fields]}` but got `{json}`")
@@ -246,6 +262,21 @@ class AuraApi:
             time.sleep(sleep_time)
 
         return WaitResult.from_error(f"Instance is not running after waiting for {waited_time} seconds")
+
+    def estimate_size(
+        self, node_count: int, relationship_count: int, algorithm_categories: List[AlgorithmCategory]
+    ) -> EstimationDetails:
+        data = {
+            "node_count": node_count,
+            "relationship_count": relationship_count,
+            "algorithm_categories": algorithm_categories,
+            "tier": "gds",  # fixed as professional or enterprise makes no difference in sizing
+        }
+
+        response = req.post(f"{self._base_uri}/v1/instances/estimations", headers=self._build_header(), json=data)
+        response.raise_for_status()
+
+        return EstimationDetails.from_json(response.json()["data"])
 
     def _get_tenant_id(self) -> str:
         response = req.get(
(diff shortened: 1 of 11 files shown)


soerenreichardt:
[[🍓] Improve instance sizing API](https://trello.com/c/E6Au6njD/8926-%F0%9F%8D%93-improve-instance-sizing-api)

soerenreichardt:
Is the v1 folder the correct place or does this need to go into v1beta4 or some other beta?

graphdatascience/session/aura_api.py
```diff
@@ -247,6 +263,21 @@ def wait_for_instance_running(
...
+        self, node_count: int, relationship_count: int, algorithm_categories: List[AlgorithmCategory]
+    ) -> EstimationDetails:
+        data = {
+            "node_count": node_count,
+            "relationship_count": relationship_count,
+            "algorithm_categories": algorithm_categories,
+            "tier": "gds",  # fixed as professional or enterprise makes no difference in sizing
+        }
+
+        response = req.post(f"{self._base_uri}/v1/instances/estimations", headers=self._build_header(), json=data)
```
FlorentinD, line 276:
```suggestion
        response = req.post(f"{self._base_uri}/v1/instances/sizing", headers=self._build_header(), json=data)
```

graphdatascience/session/aura_api.py
```diff
@@ -247,6 +263,21 @@ def wait_for_instance_running(
...
         return WaitResult.from_error(f"Instance is not running after waiting for {waited_time} seconds")
 
+    def estimate_size(
+        self, node_count: int, relationship_count: int, algorithm_categories: List[AlgorithmCategory]
+    ) -> EstimationDetails:
+        data = {
+            "node_count": node_count,
+            "relationship_count": relationship_count,
+            "algorithm_categories": algorithm_categories,
+            "tier": "gds",  # fixed as professional or enterprise makes no difference in sizing
```
FlorentinD, line 273:
```suggestion
            "instance_type": "gds",  # fixed as professional or enterprise makes no difference in sizing
```

graphdatascience/session/gds_sessions.py
```diff
@@ -82,6 +84,23 @@ def __init__(self, api_credentials: AuraAPICredentials) -> None:
             client_secret=api_credentials.client_secret,
         )
 
+    def estimate(
+        self, node_count: int, relationship_count: int, algorithm_categories: Optional[List[AlgorithmCategory]] = None
+    ) -> SessionSizeByMemory:
+        if algorithm_categories is None:
+            algorithm_categories = []
```
soerenreichardt, line 91: We need to double check, but I think when algorithm categories is empty, the result is always to use 8gb

FlorentinD: good point! is that an EstimationCLI problem?
Shouldnt we at least estimate the projection based on the counts?

soerenreichardt: Not sure where this problem originates, either in console-api or the estimation cli.
Ideally the projection should be taken into account but as far as I recall that wasn't the case

graphdatascience/tests/unit/test_gds_sessions.py
```diff
@@ -424,5 +435,25 @@ def test_invalid_session_name() -> None:
...
+def test_estimate_size_exceeds(aura_api: AuraApi) -> None:
+    aura_api = FakeAuraApi(size_estimation=EstimationDetails("16GB", "8GB", True))
+    sessions = GdsSessions(AuraAPICredentials("", "", "foo"))
+    sessions._aura_api = aura_api
+
+    with pytest.warns(
+        ResourceWarning,
+        match=re.escape("The estimated memory `16GB` exceeds the maximum size supported of your Aura tenant (`8GB`)"),
+    ):
+        sessions.estimate(1, 1, [AlgorithmCategory.CENTRALITY]) == SessionSizeByMemory.M
```
Mats-SX, line 455: missing an assertion

graphdatascience/session/gds_sessions.py
```diff
@@ -82,6 +84,23 @@ def __init__(self, api_credentials: AuraAPICredentials) -> None:
...
+
+        estimation = self._aura_api.estimate_size(node_count, relationship_count, algorithm_categories)
+
+        if estimation.did_exceed_maximum:
+            warnings.warn(
+                f"The estimated memory `{estimation.min_required_memory}` exceeds the maximum size"
+                f" supported of your Aura tenant (`{estimation.recommended_size}`).",
```
Mats-SX, line 98: there's a string interpolation bug in here

Mats-SX: oh it was fixed in another commit

Mats-SX (approved):
The SessionMemory allows too small sizes for AuraDS-based sessions. I will reduce its value space in another PR.

graphdatascience/session/gds_sessions.py
Mats-SX: no the 400 was coming from trying to create an AuraDS in Aura API with a size of 4 or 2 or 1

Mats-SX: yes the Aura API does return only an available size, I think. Although that requires tenant info, which is not there

PR #558, opened 2023-12-19
Simplest version of dumping train config

orazve:

Make sure:

### DIFF
--- a/graphdatascience/model/model_proc_runner.py
+++ b/graphdatascience/model/model_proc_runner.py
@@ -1,3 +1,4 @@
+import json
 from typing import Any, Dict, List, Optional, Tuple
 
 from pandas import DataFrame, Series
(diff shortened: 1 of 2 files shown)


graphdatascience/model/model_proc_runner.py
```diff
@@ -45,6 +46,34 @@ def create(
             relationship_type_embeddings,
         )
 
+    @compatible_with("train", min_inclusive=ServerVersion(2, 5, 0))
+    @client_only_endpoint("gds.model.transe")
+    def train(self,
```
adamnsch, line 51: I'm a little bit confused about how the predict-only TransE we've had previously fits into this. They kind of seem like different things. Also I wonder if it's better if we have a `gds.kge.train` call instead of one dedicated to each particular scoring function

graphdatascience/model/model_proc_runner.py
```diff
@@ -45,6 +46,34 @@ def create(
...
+              proportions: list,
+              embedding_dimension: int,
+              batch_size: int,
+              epochs: int,
+              optimizer: str,
+              optimizer_kwargs: dict,
+              # loss: str
+              ) -> int:
+        config = {'scoring_function': 'TransE',
+                  'proportions': proportions,
```
adamnsch, line 62: I call this `split_ratios` in the python runtime. I think "ratio" is a more common term to refer to this (also what pyKEEN use for example)

adamnsch (commented):
Cool :)

graphdatascience/model/model_proc_runner.py
orazve: I think that we are adding training step in addition to existing prediction one. Maybe it's worth to have a fully separate API like `gds.kge.train` or `gds.kge.transe.train`.

Personally I prefer to mention scoring method in a function name, like `gds.kge.transe.train`, not `gds.kge.train`. Because KGE algorithms are different and supposed to catch different relationship properties.

Our API proposal has `gds.model.transe.train` call, that's why I wrote it that way.

adamnsch: Ok! Fair enough :) I do still prefer `gds.kge.train`: I think it makes sense since the embeddings are what are trained, not TransE which is a scoring function. And since all KGE algos are the same (in terms of eg. hyperparameters) except for the scoring function, I think it would make sense to group them in the API for simplicity, sharing the same docs, etc. In that sense scoring function is just another hyperparam I think, and one may even want to use an ensemble of them. I like how pyKEEN designed their API

PR #566, opened 2024-01-09
Decrease version range for pyarrow

soerenreichardt:


soerenreichardt:
You're right, a bit more room doesn't hurt. I'd like to be somewhat restrictive though as we were seeing some inconsistencies lately between versions. We could go with `>=10` for example which maps to the arrow version released in gds 2.3.0. I'll also update the `tox.ini`

adamnsch:
> You're right, a bit more room doesn't hurt. I'd like to be somewhat restrictive though as we were seeing some inconsistencies lately between versions, also downgrading pip installations seems ok so users can always choose an earlier version of the client. We could go with `>=10` for example which maps to the arrow version released in gds 2.3.0. I'll also update the `tox.ini`.

ah ok! pyarrow is so annoying haha. But you're right, users can use earlier version of the client. But probably there are not many reasons why they shouldn't just use the latest. Maybe if they're on a very old python version 🤔 anyway, >= 10 sounds reasonable!

tox.ini
```diff
@@ -1,8 +1,8 @@
 [tox]
 envlist =
 	py{38,39}-neo4j{4,5}-pandas{1,2}-main-{standard,encrypted},
-	py{38,39}-pandas1-pyarrow{4,5,6,7,8,9,10,11,12,13,14}-main-{standard,encrypted},
```
adamnsch, line 4: This is so great! finally @FlorentinD

adamnsch (approved):
Nice work!

PR #569, opened 2024-01-16
Check AURA_ENV on _init_

FlorentinD:
As otherwise it would be checked to early on importing the library.
Maybe you also switch the env between init calls.

### DIFF
--- a/graphdatascience/gds_session/aura_api.py
+++ b/graphdatascience/gds_session/aura_api.py
@@ -71,9 +71,6 @@ class InstanceCreateDetails:
 
 
 class AuraApi:
-    DEV_ENV = os.environ.get("AURA_ENV")
-    BASE_URI = "https://api.neo4j.io" if not DEV_ENV else f"https://api-{os.environ.get('AURA_ENV')}.neo4j-dev.io"
-
     class AuraAuthToken:
         access_token: str
         expires_in: int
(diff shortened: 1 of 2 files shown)


graphdatascience/gds_session/aura_api.py
```diff
@@ -71,9 +71,6 @@ def from_json(cls, json: dict[str, Any]) -> InstanceCreateDetails:
 
 
 class AuraApi:
-    DEV_ENV = os.environ.get("AURA_ENV")
-    BASE_URI = "https://api.neo4j.io" if not DEV_ENV else f"https://api-{os.environ.get('AURA_ENV')}.neo4j-dev.io"
```
Mats-SX, line 75: ach order of evaluation strikes again

graphdatascience/gds_session/aura_api.py
```diff
@@ -117,7 +116,7 @@ def create_instance(self, name: str, memory: str, cloud_provider: str, region: s
         }
 
         response = req.post(
-            f"{AuraApi.BASE_URI}/v1/instances",
+            f"{self._base_uri}/v1/instances",
```
Mats-SX, line 119: this design is perfectly fine anyway

Mats-SX (approved):
This was messy in the past I recall. Good fix

PR #570, opened 2024-01-16
Gds Sessions updates

Mats-SX:


changelog.md
```diff
@@ -6,7 +6,8 @@
 
 ## New features
 
-* Add gds.graph.project.remoteDb endpoint for projecting graphs from AuraDB instances
+* Add the new concept of GDS Sessions, used to manage GDS computations in Aura, based on data from an AuraDB instance.
+  * Add `gds.graph.project.remote` endpoint for projecting graphs from AuraDB instances
```
FlorentinD:
```suggestion
  * Add a new `gds.graph.project` endpoint to project graphs from AuraDB instances to GDS sessions.
```

Actually from the client perspective, its not called project.remote anymore but only just `gds.graph.project`.

examples/dev/gds-sessions.ipynb
FlorentinD, line 1: notebooks are very hard to review, I found some smaller things which I will put in a commit

graphdatascience/gds_session/schema.py
```diff
@@ -0,0 +1,12 @@
+from enum import Enum
```
FlorentinD, line 1: we should consider to add these as high-level classes (add to __init__.py). Not for this PR, but before promoting the feature

FlorentinD (approved):
LGTM

changelog.md
Mats-SX: you're right of course!

examples/dev/gds-sessions.ipynb
Mats-SX: the name of the method? it still says `remote` 🙈

graphdatascience/gds_session/schema.py
Mats-SX: what does that do? it seems like importing does still work

FlorentinD: instead of `from graphdatascience.session import Schema`instead of `from graphdatascience.session.schema import Schema`. Or even `from graphdatascience import Schema`

examples/dev/gds-sessions.ipynb
FlorentinD: the aggregation function is still called `gds.graph.project.remote`. The client endpoint only is `gds.graph.project`

graphdatascience/gds_session/schema.py
Mats-SX: aha

PR #574, opened 2024-01-19
Parse node IDs correctly for heterogeneous OGBL graphs

adamnsch:

Make sure:

### DIFF
--- a/graphdatascience/graph/ogb_loader.py
+++ b/graphdatascience/graph/ogb_loader.py
@@ -314,8 +314,8 @@ class OGBLLoader(OGBLoader):
                 assert source_labels[i] == source_label
                 assert target_labels[i] == target_label
 
-                source_ids[i] += node_id_offsets[edges["head_type"][i]]
-                target_ids[i] += node_id_offsets[edges["tail_type"][i]]
+                source_ids[i] += node_id_offsets[edges["head_type"][i]] + edges["head"][i]
+                target_ids[i] += node_id_offsets[edges["tail_type"][i]] + edges["tail"][i]
 
                 rel_types.append(f"{edge_type}_{set_type.upper()}")
 
--- a/changelog.md
+++ b/changelog.md
@@ -12,10 +12,10 @@
 
 ## Bug fixes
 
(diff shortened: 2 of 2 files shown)


graphdatascience/graph/ogb_loader.py
```diff
@@ -314,8 +314,8 @@ def _parse_heterogeneous(self, dataset: HeterogeneousOGBLDataset) -> Tuple[List[
                 assert source_labels[i] == source_label
                 assert target_labels[i] == target_label
 
-                source_ids[i] += node_id_offsets[edges["head_type"][i]]
-                target_ids[i] += node_id_offsets[edges["tail_type"][i]]
+                source_ids[i] += node_id_offsets[edges["head_type"][i] + edges["head"][i]]
+                target_ids[i] += node_id_offsets[edges["tail_type"][i] + edges["tail"][i]]
```
FlorentinD, line 318: so we just head the node label offsets here?
The degree distribution must have been crazy.

The fix is supposed to add the actual source/target id. `node_id_offsets` is a dict label -> offset though is it? So I am surprised to have the actual ids added before we access the offset.
Do we have a test? We could count the number of non-zero degree nodes (filter + node_count)
```suggestion
                source_ids[i] += node_id_offsets[edges["head_type"][i]] + edges["head"][i]
                target_ids[i] += node_id_offsets[edges["tail_type"][i]] + edges["tail"][i]
```

adamnsch: yeah, your right! my bad, will fix

FlorentinD: Might use a notebook example?

Just wondering how we can check the correctness for the other datasets as well.
Maybe just as a one-off even for all datasets

adamnsch: Yeah could do that perhaps. We do have some tests that one can run offline, but they happened to not catch this thing. It's not something I feel like working on now though. But we can add a card for it perhaps?

FlorentinD:
🤞 hope it gets green now

PR #578, opened 2024-01-24
Update known issues

nvitucci:


FlorentinD (approved):
Thanks a lot. Not just removal but also improved the rest 🌟

PR #580, opened 2024-01-29
Filter out pandas warnings

FlorentinD:
the load datasets functions use `read_parquet`


FlorentinD:
probably once we do a major release and also review the supported server versions

PR #583, opened 2024-01-30
Arrow Server version support

s1ck:
Starting from GDS 2.6.0, the GDS Arrow Server supports versioned Arrow commands (actions, descriptors, tickets).
The Server is backwards-compatible to the alpha state (unversioned commands).
The Arrow Version is exposed via the `gds.arrow.debug` procedure.

This PR introduces 
* reading the `ArrowVersion` when initializing the `ArrowQueryRunner` and
* using that version when creating actions, tickets and flight descriptors

Make sure:
- [X] Your contribution is covered by tests

### DIFF
--- a/graphdatascience/query_runner/arrow_endpoint_version.py
+++ b/graphdatascience/query_runner/arrow_endpoint_version.py
@@ -1,35 +1,35 @@
 from __future__ import annotations
 
 from enum import Enum
+from typing import List
 
 
 class ArrowEndpointVersion(Enum):
     ALPHA = ""
     V1 = "v1/"
 
-    def name(self) -> str:
+    def version(self) -> str:
         return self._name_.lower()
 
     def prefix(self) -> str:
         return self._value_
 
     @staticmethod
-    def from_arrow_info(arrow_info: Series[Any]) -> ArrowEndpointVersion:
-        supported_arrow_versions = arrow_info.get("versions", [])
+    def from_arrow_info(supported_arrow_versions: List[str]) -> ArrowEndpointVersion:
         # Fallback for pre 2.6.0 servers that do not support versions
         if len(supported_arrow_versions) == 0:
             return ArrowEndpointVersion.ALPHA
 
         # If the server supports versioned endpoints, we try v1 first
-        if ArrowEndpointVersion.V1.name() in supported_arrow_versions:
+        if ArrowEndpointVersion.V1.version() in supported_arrow_versions:
             return ArrowEndpointVersion.V1
 
-        if ArrowEndpointVersion.ALPHA.name() in supported_arrow_versions:
+        if ArrowEndpointVersion.ALPHA.version() in supported_arrow_versions:
             return ArrowEndpointVersion.ALPHA
 
         raise UnsupportedArrowEndpointVersion(supported_arrow_versions)
 
 
 class UnsupportedArrowEndpointVersion(Exception):
-    def __init__(self, server_version):
+    def __init__(self, server_version: List[str]) -> None:
         super().__init__(self, f"Unsupported Arrow endpoint versions: {server_version}")
--- a/graphdatascience/query_runner/arrow_endpoint_version.py
+++ b/graphdatascience/query_runner/arrow_endpoint_version.py
@@ -1,7 +1,7 @@
 from enum import Enum
 
 
-class ArrowVersion(Enum):
+class ArrowEndpointVersion(Enum):
     ALPHA = ""
     V1 = "v1/"
 
@@ -12,6 +12,6 @@ class ArrowVersion(Enum):
         return self._value_
 
 
-class UnsupportedArrowVersion(Exception):
+class UnsupportedArrowEndpointVersion(Exception):
     def __init__(self, server_version):
-        super().__init__(self, f"Unsupported Arrow server version: {server_version}")
+        super().__init__(self, f"Unsupported Arrow endpoint version: {server_version}")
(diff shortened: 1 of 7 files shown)


s1ck:
While introducing this, I came across building the procedure calls for arrow (https://github.com/neo4j/graph-data-science-client/blob/main/graphdatascience/query_runner/arrow_query_runner.py#L115-L227). This code seems rather clunky and I was wondering if we should introduce an abstraction e.g. `arrow_server_commands.py` which is basically an abstract way of saying e.g. `ArrowServerCommand.getNodeProperty(config) -> dict`. This could work well with versioned Arrow Server as we could introduce an implementation per version and then just delegate. I dont really like having "versionization" code spread out across separate files. although it's not a lot (yet). wdyt?

graphdatascience/query_runner/arrow_query_runner.py
```diff
@@ -49,10 +54,20 @@ def create(
...
             return fallback_query_runner
 
+    @staticmethod
+    def read_arrow_version(arrow_info: Series[Any]) -> ArrowVersion:
+        arrow_version = arrow_info.get("version", "alpha")
+        try:
+            arrow_version = ArrowVersion[arrow_version.upper()]
+            return arrow_version
+        except KeyError:
+            raise UnsupportedArrowVersion(arrow_version)
```
FlorentinD, line 65: asumming there is V2 on the server. But the server also supports V1 still. 
With this logic, we make it mandatory to update the client asap. Shouldnt we fallback to the latest known version and try if its supported?

graphdatascience/query_runner/arrow_query_runner.py
```diff
@@ -272,8 +289,15 @@ def _run_arrow_property_get(self, graph_name: str, procedure_name: str, configur
             "procedure_name": procedure_name,
             "configuration": configuration,
         }
-        ticket = flight.Ticket(json.dumps(payload).encode("utf-8"))
 
+        if self._arrow_version == ArrowVersion.V1:
+            payload = {
+                "name": "GET_MESSAGE",
```
FlorentinD, line 291: you could put `GET_MESSAGE` / `PUT_MESSAGE` also behind constants as they are tied to ArrowVersion.V1

FlorentinD: but they are also only used once 🤷

graphdatascience/tests/integration/test_graph_construct.py
```diff
@@ -554,3 +554,9 @@ def test_graph_alpha_construct_backward_compat_with_arrow(gds: GraphDataScience)
 
     with pytest.warns(DeprecationWarning):
         gds.alpha.graph.construct("hello", nodes, relationships)
+
+
+@pytest.mark.enterprise
+@pytest.mark.compatible_with(min_inclusive=ServerVersion(2, 6, 0))
+def test_arrow_endpoint_version(gds: GraphDataScience) -> None:
+    pass
```
FlorentinD, line 562: might want to remove this one

graphdatascience/query_runner/arrow_query_runner.py
s1ck: mh good point. maybe we should return a list of supported versions instead.

graphdatascience/query_runner/arrow_query_runner.py
s1ck: I will update those to the final version once I updated the server side

graphdatascience/query_runner/arrow_version.py
```diff
@@ -0,0 +1,17 @@
+from enum import Enum
+
+
+class ArrowVersion(Enum):
```
adamnsch, line 4: Nitpick: I think the `ArrowVersion` naming here can be confused with the actual version of the Arrow library (on client or server). Would it be better to call it `ArrowEndpointVersion` or even `ServerArrowEndpointVersion`, or something else that disambiguates it

adamnsch (commented):
Cool stuff 👍 Just had a minor comment that you can ignore if you want

graphdatascience/query_runner/arrow_version.py
s1ck: Good point. I need to change the version lookup anyways, so I can also make that rename. `ArrowEndpointVersion` sounds good to me.

graphdatascience/query_runner/arrow_query_runner.py
s1ck: done

s1ck:
@FlorentinD I added a few changed wrt list of versions .. in case you want to have another look

graphdatascience/query_runner/arrow_endpoint_version.py
```diff
@@ -1,17 +1,35 @@
...
-    def prefix(self):
+    def prefix(self) -> str:
         return self._value_
 
+    @staticmethod
+    def from_arrow_info(arrow_info: Series[Any]) -> ArrowEndpointVersion:
+        supported_arrow_versions = arrow_info.get("versions", [])
+        # Fallback for pre 2.6.0 servers that do not support versions
+        if len(supported_arrow_versions) == 0:
+            return ArrowEndpointVersion.ALPHA
```
FlorentinD, line 21: NIT: for empty list you can also do `if not supported_arrow_versions`

s1ck: yeah, I read about that, but preferred the more explicit way

PR #584, opened 2024-02-14
Validate create requests against tenants available configurations

FlorentinD:


graphdatascience/gds_session/aura_api.py
```diff
@@ -110,14 +154,14 @@ def extract_id(uri: str) -> str:
...
-        # TODO should give more control here
+        tenant_details = self.tenant_details()
+
         data = {
             "name": name,
             "memory": memory,
             "version": "5",
             "region": region,
-            # TODO should be figured out from the tenant details in the future
-            "type": self._instance_type(),
```
knutwalker, line 120: The true unicorn in the industry -- actually fixing a TODO comment 🤯

graphdatascience/gds_session/gds_sessions.py
```diff
@@ -45,7 +45,11 @@ def __init__(self, ds_connection: AuraAPICredentials) -> None:
         )
 
     def get_or_create(
-        self, session_name: str, size: SessionSizeByMemory, db_connection: DbmsConnectionInfo
+        self,
+        session_name: str,
+        size: SessionSizeByMemory,
+        db_connection: DbmsConnectionInfo,
+        region: Optional[str] = None,
```
FlorentinD, line 52: this is only needed while we have AuraDS based sessions. Afterwards we dont want to allow this actually

examples/dev/gds-sessions.ipynb
```diff
@@ -542,8 +548,22 @@
...
+   "codemirror_mode": {
+    "name": "ipython",
+    "version": 3
+   },
+   "file_extension": ".py",
+   "mimetype": "text/x-python",
+   "name": "python",
+   "nbconvert_exporter": "python",
+   "pygments_lexer": "ipython3",
+   "version": "3.11.7"
```
Mats-SX, line 566: I think this doesn't hurt, but you can clear it with `makestyle` I think

examples/dev/gds-sessions.ipynb
```diff
@@ -311,6 +316,7 @@
     "    session_name=\"pagerank-compute\",\n",
     "    size=SessionSizes.by_memory().DEFAULT,\n",
     "    db_connection=db_connection_info,\n",
+    "    region=\"europe-west1\",\n",
```
Mats-SX, line 319: yeah I don't know why we want to be anywhere but the same region as the db

graphdatascience/gds_session/gds_sessions.py
```diff
@@ -67,11 +67,11 @@ def get_or_create(
             GdsSessions._instance_name(session_name), size.value, db_instance.cloud_provider, region
         )
         wait_result = self._aura_api.wait_for_instance_running(create_details.id)
-        if wait_result is not None:
-            raise RuntimeError(f"Failed to create session `{session_name}`: {wait_result}")
+        if err := wait_result.error:
```
Mats-SX, line 70: are you Golanging here now? `:=` 😅

Mats-SX (changes requested):
I don't know why we need to have region as an overridable top-level parameter. I suspect that you are thinking of a case where a customer has AuraDB in a region that AuraDS does not support? Is that it? I am not convinced we _need_ this ability even for that case.

graphdatascience/tests/unit/test_gds_sessions.py
```diff
@@ -176,6 +182,31 @@ def test_create_default_session(mocker: MockerFixture, aura_api: AuraApi) -> Non
...
+    )
+    mocker.patch(
+        "graphdatascience.gds_session.gds_sessions.GdsSessions._change_initial_pw", lambda *args, **kwargs: kwargs
+    )
+
+    sessions.get_or_create(
+        "my-session",
+        SessionSizeByMemory.DEFAULT,
+        DbmsConnectionInfo("neo4j+ssc://ffff0.databases.neo4j.io", "dbuser", "db_pw"),
+        region="dresden-2",
```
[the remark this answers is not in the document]
FlorentinD: We could also try to derive the region if not available. 
Checking back, there is only one per continent, so we could have a mapping logic from the db region instead.

Seems cleaner

FlorentinD:
@Mats-SX thanks for your review! I take the closest matching region now. You are right, we dont need to let the user decide, we can make a good decision here.

PR #593, opened 2024-02-27
Add example for db_node_properties

nvitucci:


doc/modules/ROOT/pages/graph-object.adoc
```diff
@@ -452,18 +452,54 @@ gds.graph.drop(G)
...
 
 are greatly sped up if https://neo4j.com/docs/graph-data-science/current/installation/installation-apache-arrow/[Apache Arrow Flight Server] of GDS is enabled.
 
 Additionally, setting the client only optional keyword parameter `separate_property_columns=True` (it defaults to `False`) for `gds.graph.streamNodeProperties` and `gds.graph.streamRelationshipProperties` returns a pandas `DataFrame` in which each property requested has its own column.
 Note that this is different from the default behavior for which there would only be one column called `propertyValue` that contains all properties requested interleaved for each node or relationship.
 
-`gds.graph.nodeProperties.stream` also takes an additional optional client only parameter `db_node_properties`, which takes a list of string representing node properties that are only in the Neo4j DB, but not on the in-memory graph.
-This can be used as a convenience parameter to fetch DB-only properties (for example, `String` type node properties that are not supported in-memory) and enrich streamed results.
+
+[[graph-object-streaming-db-properties]]
```
FlorentinD, line 466: we need to rebase this PR once #595  is in. (and also adjust the docs)

doc/modules/ROOT/pages/graph-object.adoc
```diff
@@ -452,18 +452,54 @@ gds.graph.drop(G)
...
 
 are greatly sped up if https://neo4j.com/docs/graph-data-science/current/installation/installation-apache-arrow/[Apache Arrow Flight Server] of GDS is enabled.
 
 Additionally, setting the client only optional keyword parameter `separate_property_columns=True` (it defaults to `False`) for `gds.graph.streamNodeProperties` and `gds.graph.streamRelationshipProperties` returns a pandas `DataFrame` in which each property requested has its own column.
 Note that this is different from the default behavior for which there would only be one column called `propertyValue` that contains all properties requested interleaved for each node or relationship.
 
-`gds.graph.nodeProperties.stream` also takes an additional optional client only parameter `db_node_properties`, which takes a list of string representing node properties that are only in the Neo4j DB, but not on the in-memory graph.
-This can be used as a convenience parameter to fetch DB-only properties (for example, `String` type node properties that are not supported in-memory) and enrich streamed results.
```
FlorentinD, line 466: we removed the section for the `db_node_properties` but not for the `separate_property_columns`.

I would argue, these descriptions could both belong to the api docs (at `graph.rst`).

and here we could have a shorter syntax table if at all?

doc/modules/ROOT/pages/graph-object.adoc
```diff
@@ -452,18 +452,54 @@ gds.graph.drop(G)
...
+
+[[graph-object-streaming-db-properties]]
+==== Including node properties from Neo4j
+
+Graph projections cannot include non-numeric node properties such as names and addresses, because `String`-type node properties are not supported in memory.
+To fetch additional node properties directly from the Neo4j database, you can use the `db_node_properties` client-only parameter of the `gds.graph.nodeProperties.stream` method.
```
FlorentinD, line 470: How about starting with rather the reason of why you need these properties now - `Not every property in your database, is needed for running algorithms, but can be vital to understand the result. Examples could be string type properties such as names but also other numerical properties`.

doc/modules/ROOT/pages/graph-object.adoc
```diff
@@ -452,18 +452,54 @@ gds.graph.drop(G)
...
+
+In the following example, the `City` nodes have both a numeric and a `String` property; attempting to include the `name` property in the projection would fail with an `UnsupportedOperationException`.
```
FlorentinD, line 472: this goes very much into the limitations of GDS. I would rather argue, GDS cannot make use of string properties, so why use up the memory in the first place?

doc/modules/ROOT/pages/graph-object.adoc
FlorentinD: i would use the plural in the example.

FlorentinD (commented):
Thanks for adding the example. I have some ideas around phrasing

doc/modules/ROOT/pages/graph-object.adoc
nvitucci: Sure, if there is no other reason to have them here (for example to draw extra attention to them, like for the `db_node_properties`) they can be moved to the API docs.

FlorentinD: For extra attention, we shouldn't we convert it to an example as well?
 I think its a neat feature

nvitucci: We can, but I'd do it in a different PR then :)

doc/modules/ROOT/pages/graph-object.adoc
nvitucci: I've rephrased it a bit.

doc/modules/ROOT/pages/graph-object.adoc
nvitucci: Rephrased as well.

FlorentinD (approved):
LGTM.
Suprised, there is no merge conflict

nvitucci:
> LGTM. Suprised, there is no merge conflict

Maybe because I have rebased already :)

PR #603, opened 2024-03-07
robinhood

jjaderberg:
- **Optionally override arrow address**
- **Rename listen address to connection string**


graphdatascience/tests/unit/test_arrow_runner.py
```diff
@@ -0,0 +1,40 @@
+import pytest
+from pandas import DataFrame
+from pyarrow.flight import FlightUnavailableError
+
+from .conftest import CollectingQueryRunner
+from graphdatascience.query_runner.arrow_query_runner import ArrowQueryRunner
+from graphdatascience.server_version.server_version import ServerVersion
+
+
+@pytest.mark.parametrize("server_version", [ServerVersion(2, 6, 0)])
```
FlorentinD, line 10: this test should work also for earlier GDS server versions

FlorentinD (approved):
LGTM

You could use your fork for branching next time :)

graphdatascience/tests/unit/test_arrow_runner.py
knutwalker: not all of them, return fields require >= 2.2.1
We didn't add a test for the older version because they are no longer supported

PR #609, opened 2024-03-20
Encode build id in AuraDS test instance

FlorentinD:


FlorentinD:
[Encode build id in test instances](https://trello.com/c/gKrdTZPC/9205-encode-build-id-in-test-instances)

PR #612, opened 2024-03-27
Handle empty rels streamed back via arrow

FlorentinD:
- **Ignore credential files**
- **Handle empty table in arrow based stream**


FlorentinD:
[Customer would like a more helpful error message when streaming relationships where a mutate call created zero relationships.](https://trello.com/c/k2Uya8C8/9335-customer-would-like-a-more-helpful-error-message-when-streaming-relationships-where-a-mutate-call-created-zero-relationships)

DarthMax (approved):
LGTM

PR #626, opened 2024-04-05
Use correct authorization key to get the token

Mats-SX:


graphdatascience/query_runner/arrow_query_runner.py
```diff
@@ -377,10 +377,17 @@ def __init__(self, factory: AuthFactory, *args: Any, **kwargs: Any) -> None:
         self._factory = factory
 
     def received_headers(self, headers: Dict[str, Any]) -> None:
-        auth_header: str = headers.get("Authorization", None)
+        auth_header = headers.get("authorization", None)
```
FlorentinD, line 380: for the AuraDbArrowQueryRunner we did use the lower case already

FlorentinD (approved):
Great find!

graphdatascience/query_runner/arrow_query_runner.py
Mats-SX: indeed

PR #629, opened 2024-04-11
Change size to memory and improve error messages

FlorentinD:
- **Remove memory normalization**
- **Test tenantid is mandatory for multiple tenants**


FlorentinD:
[[🍓] Session tweaks in client](https://trello.com/c/3fbPc60h/9249-%F0%9F%8D%93-session-tweaks-in-client)

graphdatascience/session/gds_sessions.py
```diff
@@ -32,13 +32,12 @@ class SessionInfo:
...
     @classmethod
     def from_specific_instance_details(cls, instance_details: InstanceSpecificDetails) -> SessionInfo:
-
         size = ""
         try:
             # instance creation also allows for GB but instance details returns size as GiB
-            memory = instance_details.memory.replace("GiB", "GB").replace(" ", "")
+            memory = instance_details.memory
             if memory:
-                size = SessionSizeByMemory(memory).name
```
FlorentinD, line 41: do you want to change the SessionInfo to also say `memory` instead of size?

graphdatascience/session/gds_sessions.py
```diff
@@ -106,13 +105,13 @@ def get_or_create(
         if existing_session:
             session_id = existing_session.id
             existing_size = SessionInfo.from_specific_instance_details(existing_session).size
```
FlorentinD, line 107: no need to use the SessionInfo indirection now.
Could directly use `existing_session.memory`

graphdatascience/session/session_sizes.py
```diff
@@ -1,33 +1,20 @@
...
+    _4GB = "4GB"
+    _8GB = "8GB"
+    _16GB = "16GB"
+    _32GB = "32GB"
+    _64GB = "64GB"
+    _128GB = "128GB"
+    _256GB = "256GB"
+    _320GB = "320GB"
+    _512GB = "512GB"
     DEFAULT = "8GB"
```
FlorentinD, line 20: why do we even have the default option?

graphdatascience/session/session_sizes.py
```diff
@@ -1,33 +1,20 @@
...
-class SessionSizes:
-    @staticmethod
-    def by_memory() -> Type[SessionSizeByMemory]:
-        """
-        Helper method to specify the memory size of a session.
-
-        Returns:
-            Type[SessionSizeByMemory]: The SessionSizeByMemory class.
-        """
-        return SessionSizeByMemory
```
FlorentinD, line 33: you want to update the api docs as well to reflect these changes

FlorentinD (commented):
LGTM, just missing the api docs and some things can be simpler 👍 

As the PR owner, I cant approve it.

graphdatascience/session/gds_sessions.py
Mats-SX: Yes, good catch.

graphdatascience/session/gds_sessions.py
Mats-SX: Yes, you're right!

graphdatascience/session/session_sizes.py
Mats-SX: I thought it was a usability feature. If the user really doesn't know, then this size is a good starting point.

We can remove it if you do not believe in it.

Mats-SX: But not in this PR.

graphdatascience/session/session_sizes.py
Mats-SX: Yes, will do.

graphdatascience/session/session_sizes.py
FlorentinD: I think, if we make the size an optional parameter on the get_and_create, than its not needed to have an enum value for it.

But than its not much overhead to have it, so I dont mind.

Mats-SX: Actually, as it turns out, we cannot have the same value twice in the enum. So DEFAULT is going out now.

Mats-SX:
@FlorentinD I've added several commits to address a number of issues that were discovered thanks to your review. Please, could you cast an eye on the new changes as well?

FlorentinD (commented):
LGTM 👍 

Only minor wording suggestions

PR #630, opened 2024-04-12
Add dedicated session endpoints

FlorentinD:
- **Move aura-api response data classes**
- **Map new aura-api session endpoints**
- **Wait for session to be ready**

### DIFF
--- a/graphdatascience/session/aura_api.py
+++ b/graphdatascience/session/aura_api.py
@@ -1,12 +1,9 @@
 from __future__ import annotations
 
-import dataclasses
 import logging
 import os
 import time
-from collections import defaultdict
-from dataclasses import dataclass
-from typing import Any, List, NamedTuple, Optional, Set
+from typing import Any, List, Optional
 from urllib.parse import urlparse
 
 import requests as req
@@ -14,122 +11,6 @@ from requests import HTTPError
 
 from graphdatascience.session.algorithm_category import AlgorithmCategory
 from graphdatascience.version import __version__
-
-
-@dataclass(repr=True, frozen=True)
-class InstanceDetails:
-    id: str
-    name: str
-    tenant_id: str
-    cloud_provider: str
-
-    @classmethod
-    def fromJson(cls, json: dict[str, Any]) -> InstanceDetails:
-        return cls(
-            id=json["id"],
-            name=json["name"],
-            tenant_id=json["tenant_id"],
-            cloud_provider=json["cloud_provider"],
-        )
-
-
-@dataclass(repr=True, frozen=True)
-class InstanceSpecificDetails(InstanceDetails):
-    status: str
-    connection_url: str
-    memory: str
-    type: str
-    region: str
-
-    @classmethod
-    def fromJson(cls, json: dict[str, Any]) -> InstanceSpecificDetails:
-        return cls(
-            id=json["id"],
-            name=json["name"],
-            tenant_id=json["tenant_id"],
-            cloud_provider=json["cloud_provider"],
-            status=json["status"],
-            connection_url=json.get("connection_url", ""),
-            memory=json.get("memory", ""),
-            type=json["type"],
-            region=json["region"],
-        )
-
-
-@dataclass(repr=True, frozen=True)
-class InstanceCreateDetails:
-    id: str
-    username: str
-    password: str
-    connection_url: str
-
-    @classmethod
-    def from_json(cls, json: dict[str, Any]) -> InstanceCreateDetails:
-        fields = dataclasses.fields(cls)
-        if any(f.name not in json for f in fields):
-            raise RuntimeError(f"Missing required field. Expected `{[f.name for f in fields]}` but got `{json}`")
-
-        return cls(**{f.name: json[f.name] for f in fields})
-
-
-@dataclass(repr=True, frozen=True)
-class EstimationDetails:
-    min_required_memory: str
-    recommended_size: str
-    did_exceed_maximum: bool
-
-    @classmethod
-    def from_json(cls, json: dict[str, Any]) -> EstimationDetails:
-        fields = dataclasses.fields(cls)
-        if any(f.name not in json for f in fields):
-            raise RuntimeError(f"Missing required field. Expected `{[f.name for f in fields]}` but got `{json}`")
-
-        return cls(**{f.name: json[f.name] for f in fields})
-
-
-class WaitResult(NamedTuple):
-    connection_url: str
-    error: str
-
-    @classmethod
-    def from_error(cls, error: str) -> WaitResult:
-        return cls(connection_url="", error=error)
-
-    @classmethod
-    def from_connection_url(cls, connection_url: str) -> WaitResult:
-        return cls(connection_url=connection_url, error="")
-
-
-@dataclass(repr=True, frozen=True)
-class TenantDetails:
-    id: str
-    ds_type: str
-    regions_per_provider: dict[str, Set[str]]
-
-    @classmethod
-    def from_json(cls, json: dict[str, Any]) -> TenantDetails:
-        regions_per_provider = defaultdict(set)
-        instance_types = set()
-        ds_type = None
-
-        for configs in json["instance_configurations"]:
-            type = configs["type"]
-            if type.split("-")[1] == "ds":
-                regions_per_provider[configs["cloud_provider"]].add(configs["region"])
-                ds_type = type
-            instance_types.add(configs["type"])
-
-        id = json["id"]
-        if not ds_type:
-            raise RuntimeError(
-                f"Tenant with id `{id}` cannot create DS instances. Available instances are `{instance_types}`."
-            )
-
-        return cls(
-            id=id,
-            ds_type=ds_type,
-            regions_per_provider=regions_per_provider,
-        )
 
 
 class AuraApi:
--- a/graphdatascience/session/aura_api.py
+++ b/graphdatascience/session/aura_api.py
@@ -73,11 +73,14 @@ class AuraApi:
 
         return SessionDetails.fromJson(response.json())
 
-    def list_session(self, session_id: str, dbid: str) -> SessionDetails:
+    def list_session(self, session_id: str, dbid: str) -> Optional[SessionDetails]:
         response = req.get(
             f"{self._base_uri}/v1beta5/data-science/sessions/{session_id}?instanceId={dbid}",
             headers=self._build_header(),
         )
+
+        if response.status_code == 404:
+            return None
 
         response.raise_for_status()
 
@@ -92,6 +95,27 @@ class AuraApi:
         response.raise_for_status()
 
         return [SessionDetails.fromJson(s) for s in response.json()]
+
+    def wait_for_session_running(
+        self, session_id: str, dbid: str, sleep_time: float = 0.2, max_sleep_time: float = 300
+    ) -> WaitResult:
+        waited_time = 0.0
+        while waited_time <= max_sleep_time:
+            session = self.list_session(session_id, dbid)
+            if session is None:
+                return WaitResult.from_error("Session is not found -- please retry")
+            elif session.status == "Ready" and session.host:  # check host needed until dns based routing
+                return WaitResult.from_connection_url(session.bolt_connection_url())
+            else:
+                self._logger.debug(
+                    f"Session `{session_id}` is not yet running. "
+                    f"Current status: {session.status} Host: {session.host}. "
+                    f"Retrying in {sleep_time} seconds..."
+                )
+            waited_time += sleep_time
+            time.sleep(sleep_time)
+
+        return WaitResult.from_error(f"Session is not running after waiting for {waited_time} seconds")
 
     def delete_session(self, session_id: str, dbid: str) -> bool:
         response = req.delete(
--- a/graphdatascience/session/aura_api.py
+++ b/graphdatascience/session/aura_api.py
@@ -10,6 +10,15 @@ import requests as req
 from requests import HTTPError
 
 from graphdatascience.session.algorithm_category import AlgorithmCategory
+from graphdatascience.session.aura_api_responses import (
+    EstimationDetails,
+    InstanceCreateDetails,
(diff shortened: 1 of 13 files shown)


FlorentinD:
[[🍓] Update the GDS Python Client with the new Aura API endpoints](https://trello.com/c/MxTFNDve/9166-%F0%9F%8D%93-update-the-gds-python-client-with-the-new-aura-api-endpoints)

graphdatascience/session/aura_api.py
```diff
@@ -53,6 +62,53 @@ def extract_id(uri: str) -> str:
...
+        response = req.get(
+            f"{self._base_uri}/v1beta5/data-science/sessions?instanceId={dbid}",
+            headers=self._build_header(),
+        )
+
+        response.raise_for_status()
+
+        return [SessionDetails.fromJson(s) for s in response.json()]
+
+    def delete_session(self, session_id: str, dbid: str) -> bool:
```
Mats-SX: Very elegant

graphdatascience/tests/unit/test_aura_api.py
```diff
@@ -6,16 +6,162 @@
...
+    result = api.create_session("name-0", "dbid-1", "pwd-2")
+
+    assert result == SessionDetails(
+        id="id0",
+        name="name-0",
+        status="Creating",
+        instance_id="dbid-1",
+        created_at="1970-01-01T00:00:00Z",
+        host="1.2.3.4",
+        memory="4G",  # TODO parse into sizing
```
Mats-SX, line 47: a todo

graphdatascience/session/aura_api.py
```diff
@@ -53,6 +62,53 @@ def extract_id(uri: str) -> str:
...
+
+    def list_sessions(self, dbid: str) -> List[SessionDetails]:
+        response = req.get(
+            f"{self._base_uri}/v1beta5/data-science/sessions?instanceId={dbid}",
+            headers=self._build_header(),
+        )
+
+        response.raise_for_status()
+
+        return [SessionDetails.fromJson(s) for s in response.json()]
```
Mats-SX, line 97: Should we not display a more brief set of fields on the list-many endpoint?

graphdatascience/session/aura_api.py
```diff
@@ -93,6 +96,27 @@ def list_sessions(self, dbid: str) -> List[SessionDetails]:
...
         return [SessionDetails.fromJson(s) for s in response.json()]
 
+    def wait_for_session_running(
+        self, session_id: str, dbid: str, sleep_time: float = 0.2, max_sleep_time: float = 300
+    ) -> WaitResult:
+        waited_time = 0.0
+        while waited_time <= max_sleep_time:
+            session = self.list_session(session_id, dbid)
+            if session is None:
+                return WaitResult.from_error("Session is not found -- please retry")
```
Mats-SX, line 106: Let's include the session id and dbid in this error message. I suggest

```
"Session `session_id` for database `dbid` not found -- please retry"
```

graphdatascience/session/aura_api.py
```diff
@@ -93,6 +96,27 @@ def list_sessions(self, dbid: str) -> List[SessionDetails]:
...
+            else:
+                self._logger.debug(
+                    f"Session `{session_id}` is not yet running. "
+                    f"Current status: {session.status} Host: {session.host}. "
+                    f"Retrying in {sleep_time} seconds..."
+                )
+            waited_time += sleep_time
+            time.sleep(sleep_time)
+
+        return WaitResult.from_error(f"Session is not running after waiting for {waited_time} seconds")
```
Mats-SX, line 118: Also here. I suggest

```
Session `session_id` for database `dbid` is not running after ...
```

graphdatascience/session/gds_sessions_v2.py
```diff
@@ -0,0 +1,150 @@
...
+from graphdatascience.session.aura_graph_data_science import AuraGraphDataScience
+from graphdatascience.session.dbms_connection_info import DbmsConnectionInfo
+from graphdatascience.session.gds_sessions import AuraAPICredentials
+
+
+# TODO come up with a good name (explain diff to previous version)
+class GdsSessionsV2:
+    """
+    Class for managing GDS sessions v2 hosted in Neo4j Aura.
+    """
```
Mats-SX: need to solve these

graphdatascience/session/gds_sessions_v2.py
```diff
@@ -0,0 +1,150 @@
...
+        session_name: str,
+        db_connection: DbmsConnectionInfo,
+    ) -> AuraGraphDataScience:
+        """
+        Retrieves an existing session with the given session name and database connection,
+        or creates a new session if one does not exist.
+
+        Args:
+            session_name (str): The name of the session.
+            size (SessionSizeByMemory): The size of the session specified by memory.
```
Mats-SX: there's no such parameter yet

graphdatascience/session/gds_sessions_v2.py
```diff
@@ -0,0 +1,150 @@
...
+            A list of SessionInfo objects representing the GDS sessions.
+        """
+        dbs = self._aura_api.list_instances()
+
+        sessions: List[SessionDetails] = []
+        for db in dbs:
+            sessions.extend(self._aura_api.list_sessions(db.id))
+
+        # TODO SessionDetails vs SessionInfo?
+        # TODO return dataframe?
```
Mats-SX: todo

graphdatascience/session/gds_sessions_v2.py
```diff
@@ -0,0 +1,150 @@
...
+
+        if len(matched_sessions) > 1:
+            self._fail_ambiguous_session(session_name, matched_sessions)
+
+        return matched_sessions[0]
+
+    def _create_session(self, session_name: str, dbid: str, dburi: str, pwd: str) -> SessionDetails:
+        db_instance = self._aura_api.list_instance(dbid)
+        if not db_instance:
+            raise ValueError(f"Could not find Aura instance with the uri `{dburi}`")
```
Mats-SX: It should be `AuraDB instance` I think

graphdatascience/session/dedicated_sessions.py
```diff
@@ -2,57 +2,37 @@
...
-            db_connection (DbmsConnectionInfo): The database connection information.
-
-        Returns:
-            AuraGraphDataScience: The session.
-        """
-
         dbid = AuraApi.extract_id(db_connection.uri)
         existing_session = self._find_existing_session(session_name, dbid)
 
+        # TODO configure session size (and check existing_session has same size)
```
Mats-SX, line 35: we should track this also on a card I think

graphdatascience/session/dedicated_sessions.py
```diff
@@ -91,22 +63,14 @@ def delete(self, session_name: str, dbid: Optional[str] = None) -> bool:
...
         dbs = self._aura_api.list_instances()
 
         sessions: List[SessionDetails] = []
         for db in dbs:
             sessions.extend(self._aura_api.list_sessions(db.id))
 
-        # TODO SessionDetails vs SessionInfo?
-        # TODO return dataframe?
-        return sessions
+        return [SessionInfo.from_session_details(i) for i in sessions]
```
Mats-SX, line 73: I like this choice

graphdatascience/session/gds_sessions.py
```diff
@@ -69,38 +35,25 @@ class GdsSessions:
...
             api_credentials (AuraAPICredentials): The Aura API credentials used for establishing a connection.
         """
-        self._aura_api = AuraApi(
-            tenant_id=api_credentials.tenant,
-            client_id=api_credentials.client_id,
-            client_secret=api_credentials.client_secret,
+        aura_api = AuraApi(api_credentials.client_id, api_credentials.client_secret, api_credentials.tenant)
+        session_type_flag = os.environ.get("USE_DEDICATED_SESSIONS", "false").lower() == "true"
+        self._impl: Union[DedicatedSessions, AuraDsSessions] = (
+            DedicatedSessions(aura_api) if session_type_flag else AuraDsSessions(aura_api)
```
Mats-SX, line 48: neatly done

graphdatascience/tests/unit/test_aurads_sessions.py
```diff
@@ -103,48 +97,27 @@ def aura_api() -> AuraApi:
...
-        tenant_id="tenant_id",
-        cloud_provider="gcp",
-        status="RUNNING",
-        connection_url="",
-        memory="16GB",
-        type="gds",
-        region="",
-    )
+def test_list_session(aura_api: AuraApi) -> None:
+    aura_api = FakeAuraApi()
```
Mats-SX, line 101: why do we pass in `aura_api` and then immediately overwrite it?

graphdatascience/tests/unit/test_aurads_sessions.py
```diff
@@ -426,8 +403,7 @@ def test_create_waiting_forever() -> None:
 def test_invalid_session_name() -> None:
     aura_api = FakeAuraApi(status_after_creating="updating")
     _setup_db_instance(aura_api)
-    sessions = GdsSessions(AuraAPICredentials("", "", "foo"))
-    sessions._aura_api = aura_api
+    sessions = AuraDsSessions(aura_api)
```
Mats-SX, line 406: looks like many of these tests became simpler after the refactoring. nice!

graphdatascience/tests/unit/test_aurads_sessions.py
```diff
@@ -457,29 +431,5 @@ def test_estimate_size_exceeds(aura_api: AuraApi) -> None:
...
-            )
-        )
-    with pytest.raises(ValueError, match="Invalid session name: `not-a-gds-session-foo`"):
-        SessionInfo.from_specific_instance_details(
-            InstanceSpecificDetails(
-                "id", "not-a-gds-session-foo", "tenant", "cp", "status", "bolt", "24GB", "type", "region"
-            )
-        )
-
-
```
Mats-SX, line 483: removing these tests because the method tested is removed -> fine

the replacement method is `SessionDetails.fromJson`, which is untested. We should add back these tests for that method. I suppose that ends up solving the TODO about parsing session memory as well.

Mats-SX (changes requested):
I think we can do a little bit more here before we merge

graphdatascience/session/aura_api.py
Mats-SX: I think this is solved on the session-level API.

graphdatascience/session/aura_api_responses.py
```diff
@@ -0,0 +1,147 @@
...
+@dataclass(repr=True, frozen=True)
+class SessionDetails:
+    id: str
+    name: str
+    instance_id: str
+    memory: str  # TODO parse session size?
+    status: str
+    host: Optional[str]  # TODO non-optional once dns based routing
+    expiry_date: Optional[str]  # TODO parse time into datetime using dateutil?
+    created_at: str  # TODO parse time into datetime using dateutil?
```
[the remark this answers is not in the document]
FlorentinD: actually on the aura_api level, I would leave them in the basic types.

For the sessionInfo, we can use extended types. 
The drawback of parsing into datetime is the str repr.
In general a list of SessionInfo can look very verbose in a notebook. Maybe we should spend some-time into having a better visualization there (similar to DataFrames).

graphdatascience/session/dedicated_sessions.py
FlorentinD: its tracked on the enablement wall

graphdatascience/tests/unit/test_aurads_sessions.py
FlorentinD: I realized, we dont need to parse the memory field, as in the end we endup using the str version and not the enum.
Imagine for having a better str representation?

graphdatascience/session/aura_api_responses.py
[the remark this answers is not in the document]
Mats-SX: My opinion here is to keep the dates as strings, but parse the memory into our own type.

Yes, happy to change the repr and str methods of SessionInfo, but it isn't necessary on this PR>

graphdatascience/tests/unit/test_aurads_sessions.py
Mats-SX: You're right we don't need to, but I think it's nice if we do it anyway, for general type safety.

graphdatascience/session/aura_api_responses.py
```diff
@@ -11,23 +12,25 @@ class SessionDetails:
...
     instance_id: str
-    memory: str  # TODO parse session size?
+    memory: str
     status: str
-    host: Optional[str]  # TODO non-optional once dns based routing
-    expiry_date: Optional[str]  # TODO parse time into datetime using dateutil?
-    created_at: str  # TODO parse time into datetime using dateutil?
+    host: Optional[str]
+    expiry_date: Optional[datetime]
+    created_at: datetime
```
Mats-SX, line 19: this is also nice. small cost, better types 👍

Mats-SX: in theory, this could help us catch bugs if we mess up the datetime fields in Aura, too :)

Mats-SX (approved):

Mats-SX:
> ValueError: Invalid isoformat string: '2022-01-01T00:00:00Z'

graphdatascience/tests/unit/test_aurads_sessions.py
FlorentinD: How will we add new memory options to the sessions?

First we add them to Aura, but then (old) clients will complain about the new unknown size.
We would first need to add them to the client and get everyone to upgrade.

IMO thats not a good UX. Thats why I would rather keep the str format and only offer the typed size on the API calls. This makes it easier for users to chose which size they can select, but not throw on new sizes

Mats-SX: That makes a lot of sense. Yes, better to not parse into the enum!

PR #383, opened 2023-05-29
Add testing for server endpoint coverage in ref docs

adamnsch:
Thank you for your contribution to the Graph Data Science Client project.

Before submitting this PR, please read [Contributing to the Neo4j Ecosystem](https://github.com/neo4j/graph-data-science-client/blob/main/CONTRIBUTING.md). 

Make sure:
- [x] You signed the [Neo4j CLA](https://neo4j.com/developer/cla/#sign-cla) (Contributor License Agreement) so that we are allowed to ship your code in our library
- [x] Your contribution is covered by tests

PR #631, opened 2024-04-18
Add __init__ to ref docs

FlorentinD:
Previously assumed it would be added twice without this exclusion (https://github.com/neo4j/graph-data-science-client/pull/383#discussion_r1209258179).

### DIFF
--- a/doc/sphinx/source/graphdatascience.rst
+++ b/doc/sphinx/source/graphdatascience.rst
@@ -4,4 +4,3 @@ GraphDataScience
 .. autoclass:: graphdatascience.GraphDataScience
     :members:
     :inherited-members:
-    :exclude-members: __init__


FlorentinD:
[Document constructor and parameters for GraphDataScience in the client](https://trello.com/c/XpAvsVz5/9411-document-constructor-and-parameters-for-graphdatascience-in-the-client)

Mats-SX:
Not really this PR's fault, but the docs that get displayed as an effect of this could use some tweaks. Since this PR is small and does expose the content in the API refdocs, I think we can pile on a few things, if you agree:

>aura_ds (bool, default False) – A flag that indicates that that the client is used to connect to a Neo4j Aura instance.

Maybe AuraDS?

>arrow (Union[str, bool], default True) – Arrow connection information. Either a flag that indicates whether the client should use Apache Arrow for data streaming if it is available on the server. True means discover the connection URI from the server. A connection URI (str) can also be provided.

Remove the word "Either"? The sentence doesn't compile. It is a union so maybe there should be an "either". How about:

GDS Arrow Server connection. This is either a bool or a string. If it is a string, it will be interpreted as a connection URL to a GDS Arrow Server. If it is a bool, `True` will make the client discover the connection URI via the Bolt server, while `False` will make the client use Bolt for all operations.

>arrow_disable_server_verification (bool, default True) – A flag that indicates that, if the flight client is connecting with TLS, that it skips server verification. If this is enabled, all other TLS settings are overridden.

Too many "that" in the first sentence. I suggest to replace both sentences with:

A flag that overrides other TLS settings and disables server verification for TLS connections.

>arrow_tls_root_certs (Optional[bytes], default None) – PEM-encoded certificates that are used for the connecting to the Arrow Flight server.

"the connecting" -> "the connection"
"the Arrow Flight server" -> "the GDS Arrow Server"

Mats-SX (changes requested):
Not really this PR's fault, but the docs that get displayed as an effect of this could use some tweaks. Since this PR is small and does expose the content in the API refdocs, I think we can pile on a few things, if you agree:

>aura_ds (bool, default False) – A flag that indicates that that the client is used to connect to a Neo4j Aura instance.

Maybe AuraDS?

>arrow (Union[str, bool], default True) – Arrow connection information. Either a flag that indicates whether the client should use Apache Arrow for data streaming if it is available on the server. True means discover the connection URI from the server. A connection URI (str) can also be provided.

Remove the word "Either"? The sentence doesn't compile. It is a union so maybe there should be an "either". How about:

GDS Arrow Server connection. This is either a bool or a string. If it is a string, it will be interpreted as a connection URL to a GDS Arrow Server. If it is a bool, `True` will make the client discover the connection URI via the Bolt server, while `False` will make the client use Bolt for all operations.

>arrow_disable_server_verification (bool, default True) – A flag that indicates that, if the flight client is connecting with TLS, that it skips server verification. If this is enabled, all other TLS settings are overridden.

Too many "that" in the first sentence. I suggest to replace both sentences with:

A flag that overrides other TLS settings and disables server verification for TLS connections.

>arrow_tls_root_certs (Optional[bytes], default None) – PEM-encoded certificates that are used for the connecting to the Arrow Flight server.

"the connecting" -> "the connection"
"the Arrow Flight server" -> "the GDS Arrow Server"

FlorentinD:
Good idea. I did briefly check it, but you found some good spots to improve 👍

Mats-SX (approved):

PR #632, opened 2024-04-18
Adapt memory configurations to AuraDS offering

Mats-SX:
~~Based on #556~~


FlorentinD (changes requested):
Good idea to match these.

For dedicated sessions we can have a different enum.

You missed the docs (https://github.com/neo4j/graph-data-science-client/blob/1b29cdd985416a211ed691692e4087a1442948c1/doc/modules/ROOT/pages/gds-session.adoc#session-memory-configuration)

Mats-SX:
@FlorentinD thanks, I fix

PR #641, opened 2024-05-02
Hash db password before usage as session pw

FlorentinD:
Avoid reusing the exact same pw. Better if session or db pw should leak.

### DIFF
--- a/graphdatascience/session/dedicated_sessions.py
+++ b/graphdatascience/session/dedicated_sessions.py
@@ -46,7 +46,7 @@ class DedicatedSessions:
         existing_session = self._find_existing_session(session_name, dbid)
 
         # hashing the password to avoid storing the actual db password in Aura
-        password = hashlib.md5(db_connection.password.encode()).hexdigest()
+        password = hashlib.sha256(db_connection.password.encode()).hexdigest()
 
         # TODO configure session size (and check existing_session has same size)
         if existing_session:
--- a/graphdatascience/session/dedicated_sessions.py
+++ b/graphdatascience/session/dedicated_sessions.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import hashlib
 import warnings
 from typing import List, Optional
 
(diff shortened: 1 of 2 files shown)


FlorentinD:
[Handle the session passwords better](https://trello.com/c/AzbZiVCu/39-handle-the-session-passwords-better)

graphdatascience/session/aurads_sessions.py
```diff
@@ -67,15 +68,25 @@ def get_or_create(
...
 
         if not existing_session:
             gds_user = create_details.username
             self._change_initial_pw(
-                gds_url=gds_url, gds_user=gds_user, initial_pw=create_details.password, new_pw=db_connection.password
+                gds_url=gds_connection.uri,
+                gds_user=gds_user,
+                initial_pw=create_details.password,
+                new_pw=gds_connection.password,
             )
```
FlorentinD, line 85: for AuraDS sessions, its about parity with dedicated sessions.

however, this would break existing sessions - upgrading the client -> cannot connect anymore.
Maybe we should only introduce it for the dedicated sessions

graphdatascience/session/dedicated_sessions.py
```diff
@@ -44,19 +45,28 @@ def get_or_create(
         dbid = AuraApi.extract_id(db_connection.uri)
         existing_session = self._find_existing_session(session_name, dbid)
 
+        # hashing the password to avoid storing the actual db password in Aura
+        password = hashlib.md5(db_connection.password.encode()).hexdigest()
```
soerenreichardt, line 49: nit: md5 is generally not recommended as hasing function. I think the reasoning does not apply here (missing salt while hashing for example), but still: Do we have the option of using another algorithm here?

soerenreichardt (approved):
👍

graphdatascience/session/dedicated_sessions.py
FlorentinD: we could totally use another one. need to make sure it also exists in other languages for later APIs.
But we cannot use salting here as we want the determinism. (the user would need to provide the salt, which is no different to a password at this point).

sha256 looks like an option

PR #647, opened 2024-05-29
Use exponential backoff for waiting

FlorentinD:
Trying every 0.2s is not needed if the average waiting time is around a minute or for instances even 3


DarthMax (approved):
Nice, :ship: :it:

PR #659, opened 2024-06-10
Fix documentation for remote projection

Mats-SX:
- no more property schemas
- concurrency config


Mats-SX:
Also remove `GdsPropertyTypes` class

FlorentinD (approved):

PR #660, opened 2024-06-11
KGE documentation - main terms and parameters

orazve:

Make sure:


doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
```diff
@@ -0,0 +1,283 @@
+= Knowledge graph embeddings
+
+Knowledge Graph Embeddings (KGE) refer to a family of algorithms designed to learn low-dimensional representations of entities and relations within a knowledge graph.
+These embeddings are utilized for tasks including link prediction, entity classification, and entity clustering.
```
adamnsch, line 4:
```suggestion
These embeddings are utilized for tasks such as link prediction, entity classification, and entity clustering.
```

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
```diff
@@ -0,0 +1,283 @@
...
+
+This chapter provides an overview of the available shallow embedding algorithms in the GDS Python runtime, such as `TransE` and `DistMult`.
+
+* `TransE`
+** https://proceedings.neurips.cc/paper_files/paper/2013/file/1cecc7a77928ca8133fa24680a88d2f9-Paper.pdf[Bordes, Antoine, et al. "Translating embeddings for modeling multi-relational data." Advances in neural information processing systems 26 (2013).^]
```
adamnsch, line 9: I think we have said in the past that we should not have links to papers as they may be broken. Instead just state the name of the authors and the paper

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
```diff
@@ -0,0 +1,283 @@
...
+* `DistMult`
+** https://arxiv.org/pdf/1412.6575[Yang, Bishan, et al. "Embedding entities and relations for learning and inference in knowledge bases." arXiv preprint arXiv:1412.6575 (2014).^]
+
+A knowledge graph is a directed, multi-relational graph.
+It consists of triples `(head, relation, tail)` where `head` and `tail` are entities and `relation` is the relationship between them.
+
+Shallow models of KGE represented here are _inductive_ algorithms for computing node and relationship embeddings (low-dimensional representations) for knowledge graphs.
+Shallow means that matrix lookups represent the entity and relation encoders.
+The algorithm is inductive, meaning that it can be trained on one graph and then applied to this exact graph or another graph with the same schema.
```
adamnsch, line 18: KGE models are not inductive. You get node embeddings as the output that only pertain the graph it was trained on

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
```diff
@@ -0,0 +1,283 @@
...
+
+* https://arxiv.org/pdf/1706.02216.pdf[William L. Hamilton, Rex Ying, and Jure Leskovec. "Inductive Representation Learning on Large Graphs." 2018.^]
+* https://arxiv.org/pdf/1911.10232.pdf[Amit Pande, Kai Ni and Venkataramani Kini. "SWAG: Item Recommendations using Convolutions on Weighted Graphs." 2019.^]
+
+
+[[algorithms-embeddings-kge-considerations]]
+== Scoring functions
+
+For KGE algorithms, the scoring function is used to compute the score of a triple `(head, relation, tail)`.
+This score is used for computing loss during training and during prediction to rank the tail entities for a given head and relation.
```
adamnsch, line 31: I think we want to say "relationship type" instead of "relation", so as to adjust the terminology to Neo4j

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
```diff
@@ -0,0 +1,283 @@
...
+
+The scoring function is a function of the embeddings of the head, relation, and tail entities.
+
+Let the embeddings of the head, relation, and tail entities be denoted by `h`, `r`, and `t` respectively.
+To compute the score of the triple `(head, relation, tail)`, the scoring function takes the embeddings `h`, `r`, and `t` as input and computes a scalar score using the corresponding formula.
+
+=== `TransE`
```
adamnsch, line 38: (nitpick) Our ascidoc convention is to have two blank lines before a new header

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
```diff
@@ -0,0 +1,283 @@
...
+* Closed World Assumption (CWA) assumes that all unobserved facts are false.
+
+* Local Closed World Assumption (LCWA) assumes that all observed facts are true.
+All corrupted triplets, which are generated by replacing the head or tail entity of a positive triplet, are false.
+
+* Stochastic Local Closed World Assumption (sLCWA) assumes that all observed facts are true.
+Some corrupted triplets are false and some are true.
+The number of corrupted triplets for each true triplet is set by the `negative_sampling_size` parameter.
+
+Knowledge graph embedding models are trained under the sLCWA assumption in current implementation.
```
adamnsch, line 78: Do we need to introduce all this terminology if we only use sLCWA?

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
```diff
@@ -0,0 +1,283 @@
...
+
+=== Loss function
+The loss function is crucial for guiding the training process of KGE models.
+It determines how the difference between predicted and actual values is calculated and minimized.
+There are several loss functions that can be used for training KGE models, see below for more details.
+
+==== Margin ranking loss
+
+Margin ranking loss is a pairwise loss function that compares the scores based on the difference between the scores of a positive triple and a negative triple.
+When negative sampling size is more than one, the loss is computed for positive triple and each negative triple and the average loss is computed.
```
adamnsch, line 89:
```suggestion
When negative sampling size is more than one, the loss is computed for a positive triple and each of its negative triples, and the average loss is computed.
```

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
```diff
@@ -0,0 +1,283 @@
...
+
+=== Optimizer
+
+Any pytorch optimizer can be used for training the model.
+To use non-default optimizer, specify the optimizer class name as a string in the `optimizer` parameter.
+All optimizer parameters except `params` can be passed as `optimizer_kwargs`.
+
+=== Negative sampling
+
+Loss function requires negative samples to compute the loss.
```
adamnsch, line 109:
```suggestion
The loss function requires negative samples to compute the loss.
```

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
```diff
@@ -0,0 +1,283 @@
...
+| Parameter | Type | Default value | Description
+
+| num_epochs
+| int
+| N/A
+| Number of epochs for training (must be greater than 0)
+
+| embedding_dimension
+| int
+| 256
```
adamnsch, line 150: I don't think there should be a default value for embedding dimension. Better to force the user to specify as it depends so much on the data

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
```diff
@@ -0,0 +1,283 @@
...
+| Function used to score embeddings
+
+| p_norm
+| float
+| 1.0
+| Norm to use in TransE scoring function
+
+| batch_size
+| int
+| 512
```
adamnsch, line 180: I'm not sure it makes sense to have a default for this either? I think either we need to infer it intelligently from things like embedding size, GPU memory, etc, or not have a default. Wdyt? Same for `test_batch_size`

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
```diff
@@ -0,0 +1,283 @@
...
+
+| optimizer
+| str
+| "adam"
+| Optimizer to use for training
+
+| optimizer_kwargs
+| dict[str, Any]
+| {lr=0.01, weight_decay=0.0005}
+| Additional arguments for the optimizer
```
adamnsch, line 196:
```suggestion
| Arguments for the optimizer
```

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
```diff
@@ -0,0 +1,283 @@
...
+For some datasets, this normalization might not be beneficial.
+To avoid this normalization, set `inner_norm` to `False`.
+
+
+[[algorithms-embeddings-kge-syntax]]
+== Syntax
+
+[source,python]
+----
+gds.model.transe.train(G,
```
adamnsch, line 133: As I've said before I would prefer something like `gds.kge.train(....)`

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
```diff
@@ -0,0 +1,283 @@
...
+
+| do_test
+| bool
+| True
+| Whether to perform testing
+
+| filtered_metrics
+| bool
+| False
+| Whether to use filtered metrics during evaluation
```
adamnsch, line 246: Perhaps this needs to be explained further?

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
```diff
@@ -0,0 +1,283 @@
...
+| bool
+| False
+| Whether to sample negative nodes with the same label as the corresponding positive node
+
+| k_value
+| int
+| 10
+| Value of k used in evaluation metrics
+
+| do_validation
```
adamnsch, line 233: What kind of validation is this referring to? For each epoch? Should not the `split_ratios` dictate whether validation should be done?

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
```diff
@@ -0,0 +1,283 @@
...
+
+| load_from_checkpoint
+| Optional[tuple[str, int]]
+| None
+| Checkpoint to load from, specified as a tuple (path, epoch)
+
+| split_ratios
+| dict[str, float]
+| {TRAIN=0.8, TEST=0.2}
+| Ratios for splitting the dataset into training and test sets
```
adamnsch, line 166: There's also validation right?

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
```diff
@@ -0,0 +1,283 @@
...
+
+| epochs_per_val
+| int
+| 50
+| Number of epochs between validations (must be greater than or equal to 0)
+
+| disable_tqdm
+| bool
+| True
+| Whether to disable tqdm progress bars
```
adamnsch, line 256: TQDM is not something that is user facing right? I think what the user will see is something like MLFlow?

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
```diff
@@ -0,0 +1,283 @@
...
+
+| inner_norm
+| bool
+| True
+| Whether to apply normalization to embeddings
+
+| init_bound
+| Optional[float]
+| None
+| Initial bound for embeddings (if any)
```
adamnsch, line 266: What does this mean?

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
```diff
@@ -0,0 +1,283 @@
...
+=== Learning rate scheduler
+
+Any pytorch learning rate scheduler can be used for training the model.
+To use non-default learning rate scheduler, specify the scheduler class name as a string in the `lr_scheduler` parameter.
+All scheduler parameters except `optimizer` can be passed as `lr_scheduler_kwargs`.
+
+=== Inner normalisation
+
+In the original `TransE` paperfootnote:[Bordes, Antoine, et al. "Translating embeddings for modeling multi-relational data." Advances in neural information processing systems 26 (2013).]
+in `Algorithm 1`, line 5, the entity embeddings are normalized to have `Lp` norm of 1.
```
adamnsch, line 123: What is the value of `p` in this case?

adamnsch (commented):
Looks good, nice work! Just added some comments :)

doc/modules/ROOT/pages/python-runtime/introduction.adoc
```diff
@@ -0,0 +1,7 @@
+= Introduction
+
+Python runtime works with GDS sessions.
```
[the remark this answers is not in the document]
orazve: I agree, I'm not sure how we should start here. How do want to call this bunch of algorithms which are available with GDS sessions only?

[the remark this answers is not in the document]
adamnsch: That's a good question. I see there's a page for sessions: https://github.com/neo4j/graph-data-science-client/blob/main/doc/modules/ROOT/pages/gds-session.adoc Maybe we should just consider the KGE stuff to be another algorithm, and specify that users can only use them through sessions for now. I guess in the future, there will only be sessions, at least in Aura 🤔

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
adamnsch: Ok 👍

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
adamnsch: Ok! Yes, I agree it's good to let the user override this if they need to :)

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
adamnsch: I think it's ok to not have it be default. Just specify a sum < 1.0 and the what remains should be allocated as validation set?

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
adamnsch: ah ok!

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
adamnsch: Ok! Can the user set which `p` they want?

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
orazve: Yes, if TRAIN + TEST < 1.0 we have to do validation, I'll put a clear statement about it.

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
orazve: Yes, it's the same as `p_norm` value, I'll add information about.

adamnsch: ok 👍

doc/modules/ROOT/pages/python-runtime/introduction.adoc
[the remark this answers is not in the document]
orazve: See `algorithms.adoc` page with this suggestion

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
adamnsch: Any comment on this?

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
```diff
@@ -0,0 +1,296 @@
...
+
+* `TransE`
+** Bordes, Antoine, et al. "Translating embeddings for modeling multi-relational data." Advances in neural information processing systems 26 (2013).
+* `DistMult`
+** Yang, Bishan, et al. "Embedding entities and relations for learning and inference in knowledge bases." arXiv preprint arXiv:1412.6575 (2014)
+
+A knowledge graph is a directed, multi-relational graph.
+It consists of triples `(head, relation, tail)` where `head` and `tail` are entities and `relation` is the relationship between them.
+
+Shallow models of KGE represented here are _transductive_ algorithms for computing node and relationship embeddings (low-dimensional representations) for knowledge graphs.
```
adamnsch, line 16:
```suggestion
Shallow models of KGE represented here are _transductive_ algorithms for computing node and relationship embeddings (low-dimensional representations) for knowledge graphs.
```
"relationship" -> "relation" or "relationship type"?

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
```diff
@@ -0,0 +1,296 @@
...
+
+
+=== Filtered metrics
+
+When we evaluate (compute metrics) the model, we compute scores for all possible triples with the same head or tail and relationship type as the test triple.
+Ranking the test triple among other triples is used for computing metrics, such as Mean Rank, Mean Reciprocal Rank, and Hit@k.
+
+When `filtered_metrics` is set to `False`, the ranking is done among all possible triples.
+
+When `filtered_metrics` is set to `True`, the ranking is done among only the triples that are not present in the training set.
```
adamnsch, line 140: Nice explanation :)

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
```diff
@@ -0,0 +1,296 @@
...
+
+[[algorithms-embeddings-kge-syntax]]
+== Syntax
+
+[source, python, role=no-test]
+----
+gds.model.train(G,
```
adamnsch, line 147: I don't think that `model` is sufficiently clear, as there are several types of models in the library. Did you consider `kge` or something instead? Or perhaps something else

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
```diff
@@ -0,0 +1,296 @@
...
+| None
+| Checkpoint to load from, specified as a tuple (path, epoch)
+
+| split_ratios
+| dict[str, float]
+| {TRAIN=0.8, TEST=0.2}
+| Ratios for splitting the dataset into training and test sets.
+When sum of values is less than 1, the remaining data is used for validation.
+Validation size can be set explicitly with the key `VALID`.
+When all three keys are present, the sum of values must be equal to 1.
```
adamnsch, line 184:
```suggestion
When all three keys are present, the sum of values must be equal to 1.0.
```

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
```diff
@@ -0,0 +1,296 @@
...
+| Optional[tuple[str, int]]
+| None
+| Checkpoint to load from, specified as a tuple (path, epoch)
+
+| split_ratios
+| dict[str, float]
+| {TRAIN=0.8, TEST=0.2}
+| Ratios for splitting the dataset into training and test sets.
+When sum of values is less than 1, the remaining data is used for validation.
+Validation size can be set explicitly with the key `VALID`.
```
adamnsch, line 183:
```suggestion
The validation set ratio can be set explicitly with the key `VALID`.
```

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
```diff
@@ -0,0 +1,296 @@
...
+| load_from_checkpoint
+| Optional[tuple[str, int]]
+| None
+| Checkpoint to load from, specified as a tuple (path, epoch)
+
+| split_ratios
+| dict[str, float]
+| {TRAIN=0.8, TEST=0.2}
+| Ratios for splitting the dataset into training and test sets.
+When sum of values is less than 1, the remaining data is used for validation.
```
adamnsch, line 182:
```suggestion
When the sum of the ratios is less than 1.0, the remaining examples are used for validation.
```

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
```diff
@@ -0,0 +1,296 @@
...
+Validation size can be set explicitly with the key `VALID`.
+When all three keys are present, the sum of values must be equal to 1.
+
+| scoring_function
+| str
+| "transe"
+| Function used to score embeddings
```
adamnsch, line 189:
```suggestion
| Function used to score embeddings of triples
```
Just to clarify :)

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
```diff
@@ -0,0 +1,296 @@
...
+
+==== Negative Sampling Self-Adversarial Loss
+
+Negative Sampling Self-Adversarial Lossfootnote:[Sun, Zhiqing, et al. "Rotate: Knowledge graph embedding by relational rotation in complex space." arXiv preprint arXiv:1902.10197 (2019).] is a setwise loss function that compares the scores based on the difference between the scores of a positive triple and a set of negative triples.
+`loss_function_kwargs` can be used to set the `adversarial_temperature` and `margin` parameters.
+
+
+=== Optimizer
+
+Any pytorch optimizer can be used for training the model.
```
adamnsch, line 104: I'm worried about this. It really depends on the PyTorch version. I think it's quite dangerous to expose this sort of thing

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
```diff
@@ -0,0 +1,296 @@
...
+
+| epochs_per_checkpoint
+| int
+| max(num_epochs / 10, 1)
+| Number of epochs between checkpoints (must be greater than or equal to 0)
+
+| load_from_checkpoint
+| Optional[tuple[str, int]]
+| None
+| Checkpoint to load from, specified as a tuple (path, epoch)
```
adamnsch, line 176: I forget how it works.
```suggestion
| Checkpoint to load from, specified as a tuple (modelname, epoch)
```
I think it's just the model name right?

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
```diff
@@ -0,0 +1,296 @@
...
+
+| use_node_type_aware_sampler
+| bool
+| False
+| Whether to sample negative nodes with the same label as the corresponding positive node
+
+| k_value
+| int
+| 10
+| Value of k used in evaluation metrics
```
adamnsch, line 249:
```suggestion
| Value of k used in Hits@k evaluation metric
```
Or are there more?

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
```diff
@@ -0,0 +1,296 @@
...
+
+| inner_norm
+| bool
+| True
+| Whether to apply normalization to embeddings, see <<_inner_normalisation, inner normalization>>
+
+| init_bound
+| Optional[float]
+| None
+| The value for the range [-bound; bound] of the uniform distribution used to initialize the embeddings. Xavier's initialization is used if None.
```
adamnsch, line 279:
```suggestion
| The value for the range [-init_bound; init_bound] of the uniform distribution used to initialize the embeddings. Xavier initialization is used if None.
```
Also, should we explain or refer to somewhere for more on Xavier initialization?

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
```diff
@@ -0,0 +1,296 @@
...
+|====
+
+
+.Results
+[opts="header",cols="2m,1,6"]
+|===
+| Name              | Type    | Description
+| modelInfo         | Map     | Details of the trained model.
+| configuration     | Map     | The configuration used to run the procedure.
+| trainMillis       | Integer | Milliseconds to train the model.
```
adamnsch, line 289: I don't think this is exactly what we have atm? We have some `metadata` field atm I think. Also, Should we still use the Cypher data types when this is just going to be all Python?

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
```diff
@@ -0,0 +1,296 @@
+= Knowledge graph embeddings
+
```
adamnsch, line 2: Very nice intro!

adamnsch (commented):
Very nice, great work! Have mostly minor comments :)

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
orazve: Thanks for noticing it again, I missed this point, now it's fixed

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
```diff
@@ -0,0 +1,296 @@
...
+* `TransE`
+** Bordes, Antoine, et al. "Translating embeddings for modeling multi-relational data." Advances in neural information processing systems 26 (2013).
+* `DistMult`
+** Yang, Bishan, et al. "Embedding entities and relations for learning and inference in knowledge bases." arXiv preprint arXiv:1412.6575 (2014)
+
+A knowledge graph is a directed, multi-relational graph.
+It consists of triples `(head, relation, tail)` where `head` and `tail` are entities and `relation` is the relationship between them.
+
+Shallow models of KGE represented here are _transductive_ algorithms for computing node and relationship embeddings (low-dimensional representations) for knowledge graphs.
+Shallow means that matrix lookups represent the entity and relation encoders.
```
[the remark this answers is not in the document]
orazve: Term "shallow" doesn't relate to the method used to train the model. It is about the fact that the model on inference stage returns embedding as result of lookup operation, without any additional computing/processing.
Ref: https://arxiv.org/pdf/1709.05584
image

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
```diff
@@ -0,0 +1,296 @@
...
+
+The scoring function is a function of the embeddings of the head, relation, and tail entities.
+
+Let the embeddings of the head, relation, and tail entities be denoted by `h`, `r`, and `t` respectively.
+To compute the score of the triple `(head, relation, tail)`, the scoring function takes the embeddings `h`, `r`, and `t` as input and computes a scalar score using the corresponding formula.
+
+
+=== `TransE`
+
+`TransE` is a translational distance interaction model.
```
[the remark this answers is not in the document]
orazve: Yes, ref: https://arxiv.org/pdf/2006.13365 Chapter 3.1

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
```diff
@@ -0,0 +1,296 @@
...
+** Yang, Bishan, et al. "Embedding entities and relations for learning and inference in knowledge bases." arXiv preprint arXiv:1412.6575 (2014)
+
+A knowledge graph is a directed, multi-relational graph.
+It consists of triples `(head, relation, tail)` where `head` and `tail` are entities and `relation` is the relationship between them.
+
+Shallow models of KGE represented here are _transductive_ algorithms for computing node and relationship embeddings (low-dimensional representations) for knowledge graphs.
+Shallow means that matrix lookups represent the entity and relation encoders.
+The algorithm is inductive, meaning that it can be trained on one graph and then applied to this exact graph or another graph with the same schema.
+
+A KGE model can return the score for the triple which can be used to predict the missing tail entity for a given head and relationship type.
```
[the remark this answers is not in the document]
orazve: Currently we compute scores for tails. Head can be predicted as part of additional prediction function or as tail for inverse relationship which is not in the scope now.

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
```diff
@@ -0,0 +1,296 @@
...
+The embeddings can be used to perform various tasks, such as link prediction, entity classification, and entity clustering.
+
+
+[[algorithms-embeddings-kge-considerations]]
+== Scoring functions
+
+For KGE algorithms, the scoring function is used to compute the score of a triple `(head, relation, tail)`.
```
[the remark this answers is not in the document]
orazve: Repetition is the mother of learning 😄

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
orazve: Yes, I think we have to mention `kge`.

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
orazve: Yes

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
orazve: No, nothing more

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
orazve: I removed this chapter, until we have the right one

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
orazve: Added footnote

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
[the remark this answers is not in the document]
adamnsch: I don't disagree with this. But nevertheless I think the word shallow is used in relation to its opposite "deep" for deep learning, such as GNNs

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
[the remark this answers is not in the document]
adamnsch: Ok. It's a fairly common use-case but I guess we could add it later

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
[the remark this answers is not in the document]
adamnsch: OK 👍

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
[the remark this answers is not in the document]
adamnsch: Interesting, I've never heard that terminology anywhere else in the literature. Are we sure this is not just something that these particular authors invented for this paper?

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
adamnsch: Better, but I think we should probably explicitly list exactly the optimizers we support. Similar for schedulers of course. And what happens if these optimizers changes name between PyTorch versions...

In general I'm not super fond of us exposing the implementation details like this. But it's something Gian wants I guess

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
```diff
@@ -0,0 +1,289 @@
...
+| inner_norm
+| bool
+| True
+| Whether to apply normalization to embeddings, see <<_inner_normalisation, inner normalization>>
+
+| init_bound
+| Optional[float]
+| None
+| The value for the range [-init_bound; init_bound] of the uniform distribution used to initialize the embeddings.
+Xavier initializationfootnote:[Xavier Glorot, Yoshua Bengio "Understanding the difficulty of training deep feedforward neural networks" Proceedings of the Thirteenth International Conference on Artificial Intelligence and Statistics, PMLR 9:249-256, 2010]
```
adamnsch, line 282: Nice :)

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
adamnsch, line 1: I don't think we should expose "python-runtime" as part of this path as it's a very internal thing

adamnsch: (same for the other files)

doc/modules/ROOT/pages/python-runtime/introduction.adoc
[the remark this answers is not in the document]
adamnsch: Sounds ok for now I think :)

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
orazve: We use python runtime term in text too: `his chapter provides an overview of the available shallow embedding algorithms in the GDS Python runtime`
And we show PyTorch bind in the doc too
Do you want me to remove `python runtime` mentions from paths and text? Which term for change you suggest?

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
adamnsch: I think we have to do that, or state all names of the optimizers/schedulers, and exactly what parameters they can take. Let's think about it

doc/modules/ROOT/pages/python-runtime/knowledge-graph-embeddings.adoc
adamnsch: Yes, we should remove Python runtime from any user facing communication. Do we need another term at all? The only thing that's special about these algorithms for the user is that they can only run them through GDS Sessions

orazve: I have removed "runtime" mentioning from the doc

adamnsch (approved):

PR #662, opened 2024-06-13
Dedicated session memory checks

DarthMax:
- **Use actual types for instance memory specifications**
- **Check if a dedicated session exists with a different memory configuration**


graphdatascience/session/aura_api_responses.py
```diff
@@ -9,13 +9,15 @@
...
+from .session_sizes import SessionMemory
+
 
 @dataclass(repr=True, frozen=True)
 class SessionDetails:
     id: str
     name: str
     instance_id: str
-    memory: str
+    memory: SessionMemory
```
FlorentinD, line 20: we were not parsing this by choice. Should have put a comment.
If we introduce a new Session size option in Aura, we dont want to break client when listing sessions.

I get the point for get_or_create comparison, but for simple list it would be not ideal IMO.

graphdatascience/session/aurads_sessions.py
```diff
@@ -54,9 +54,9 @@ def get_or_create(
         if existing_session:
             session_id = existing_session.id
             # 0MB is AuraAPI default value for memory if none can be retrieved
-            if existing_session.memory != "0MB" and existing_session.memory != memory.value:
+            if existing_session.memory != memory:
```
FlorentinD, line 57: Right thats the other reason why we dont parse.
Newly created instances will return `0MB` on the instance details endpoint. So this will now fail if you quickly call the function twice

graphdatascience/session/session_sizes.py
```diff
@@ -7,6 +7,7 @@ class SessionMemory(Enum):
     Enumeration representing session main memory configurations.
     """
 
+    m_4GB = "4GB"
     m_8GB = "8GB"
```
FlorentinD, line 11: its a valid one for dedicated but not AuraDS 🤔

FlorentinD (changes requested):
The check itself looks good.

I am have some comments about the refactoring / api-change.

graphdatascience/session/aura_api_responses.py
DarthMax: I added another commit that uses a dedicated class to represent the memory. We can keep the enum to make the size selection easier. In Java i would use constant static instances of the class instead but it does not seem like this exists in Python, so maybe there is still a better way than the enum.

FlorentinD: I have an HTTP call which creates 4GB instances

graphdatascience/session/dedicated_sessions.py
```diff
@@ -139,6 +140,15 @@ def _check_expiry_date(self, session: SessionDetails) -> None:
...
                 raise Warning(f"Session `{session.name}` is expiring in less than a day.")
 
+    def _check_memory_configuration(
+        self, existing_session: SessionDetails, requested_memory: SessionMemoryValue
+    ) -> None:
+        if existing_session.memory != requested_memory:
+            raise RuntimeError(
+                f"Session `{existing_session.name}` exists with a different memory configuration. "
+                f"Current: {existing_session.memory}, Requested: {requested_memory}."
+            )
```
FlorentinD, line 150: NIT: we could share this function between the impls

graphdatascience/session/session_sizes.py
```diff
@@ -1,26 +1,53 @@
...
-    m_24GB = "24GB"
-    m_32GB = "32GB"
-    m_48GB = "48GB"
-    m_64GB = "64GB"
-    m_96GB = "96GB"
-    m_128GB = "128GB"
-    m_192GB = "192GB"
-    m_256GB = "256GB"
-    m_384GB = "384GB"
+    m_4GB = SessionMemoryValue("4GB")
```
FlorentinD, line 36: as this type is user-facing, we should put it into the `__init__.py` of this package

FlorentinD (approved):
LGTM. Only some small suggestions

PR #585, opened 2024-02-14
Support GDS sessions on self-managed DBMS

FlorentinD:
~based on #584~

PR #664, opened 2024-06-18
WIP - support for self-managed DBMS sources of GDS Sessions

Mats-SX:
Replaces #585


FlorentinD:
replaced by #712 . as most of the validation happens on the server-side, I would omit the suggestion for dedicated sessions.

PR #665, opened 2024-06-24
Avoid fetching deprecated return values

soerenreichardt:


soerenreichardt:
[Deprecation warnings in the Python client](https://trello.com/c/ePI9r5V3/9578-deprecation-warnings-in-the-python-client)

graphdatascience/query_runner/gds_arrow_client.py
```diff
@@ -18,8 +18,11 @@
 class GdsArrowClient:
     @staticmethod
     def is_arrow_enabled(query_runner: QueryRunner) -> bool:
-        arrow_info = query_runner.call_procedure(endpoint="gds.debug.arrow", custom_error=False).squeeze().to_dict()
-        return not not arrow_info["running"]
+        arrow_server_running = (
+            query_runner.call_procedure(endpoint="gds.debug.arrow", custom_error=False, yields=["running"])
+            .squeeze()
+        )
+        return not not arrow_server_running
```
knutwalker, line 25: Is this the typed python equivalent to the `!!foo` pattern in JS to turn a truthy value into True? 🙈

soerenreichardt: I guess so, but I think that is not needed anymore now that only a single value is returned

graphdatascience/query_runner/gds_arrow_client.py
```diff
@@ -30,14 +32,20 @@ def create(
         tls_root_certs: Optional[bytes] = None,
         connection_string_override: Optional[str] = None,
     ) -> "GdsArrowClient":
-        arrow_info = query_runner.call_procedure(endpoint="gds.debug.arrow", custom_error=False).squeeze().to_dict()
+        arrow_info = (
```
knutwalker, line 35: Is there a minimum supported GDS version policy for the client? Since `advertisedListendAddress` has been renamed in 2.3, do we need a check for server versions before that, or is it ok to fail on those?

soerenreichardt: I thought listen address was there from the beginning

knutwalker: Ah, yes, I think you're right. We always had `listenAddress`, then added the `advertisedListenAddress`, then soft-removed the latter by having them both return the same content.

PR #675, opened 2024-07-04
Handle empty memory on GET instance details

FlorentinD:
The endpoint does not return the actual memory config during creation


Mats-SX (approved):
Looks fine. Looking forward to removing AuraDS-based sessions.

PR #682, opened 2024-07-15
Improve Aura-API error message / Refactor AuraAPI

FlorentinD:
* use requests.Session for reusage of header
* adapter to insert Auth token
* include response error text in errors


FlorentinD:
[[🍓] Python Aura API client use request session and retry logic](https://trello.com/c/NBlMUcmo/9635-%F0%9F%8D%93-python-aura-api-client-use-request-session-and-retry-logic)

PR #687, opened 2024-07-17
Sessions - disallow topological lp + support util functions

FlorentinD:
- **Remove outdated limitations from README**
- **Omit TopologicalLPAlphaEndpoints from GDS sessions**
- **Support call_functions in QueryRunner**
- **Support gds.util.asNode(s) in Sessions**

### DIFF
--- a/doc/modules/ROOT/pages/gds-session.adoc
+++ b/doc/modules/ROOT/pages/gds-session.adoc
@@ -255,6 +255,11 @@ gds.graph.nodeProperties.stream(G, db_node_properties=["name"], node_properties=
 For a full list of the available algorithms, see the https://neo4j.com/docs/graph-data-science-client/{page-version}/api/[API reference].
 
 
+=== Limitations
+
+Topological Link Prediction algorithms are not yet supported.
+
+
 == Remote write-back
 
 The GDS Session's in-memory graph was projected from data in AuraDB, so write-back operations will persist the data back to the same AuraDB instance.
@@ -282,5 +287,3 @@ Therefore, you will not be able to call any GDS procedures from the `run_cypher(
 ----
 gds.run_cypher("MATCH (n:User) RETURN n.name, n.embedding")
 ----
-
-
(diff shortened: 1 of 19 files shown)


graphdatascience/utils/util_remote_proc_runner.py
```diff
@@ -0,0 +1,48 @@
...
+            node_ids: The ids of the nodes to get.
+
+        Returns:
+            The nodes with the given ids.
+
+        """
+        query = "MATCH (n) WHERE id(n) IN $nodeIds RETURN collect(n)"
+        params = {"nodeIds": node_ids}
+
+        return self._query_runner.run_cypher(query=query, params=params).squeeze()  # type: ignore
```
soerenreichardt, line 44: Why are these only in the remote space?

doc/modules/ROOT/pages/gds-session.adoc
```diff
@@ -255,6 +255,11 @@ gds.graph.nodeProperties.stream(G, db_node_properties=["name"], node_properties=
 For a full list of the available algorithms, see the https://neo4j.com/docs/graph-data-science-client/{page-version}/api/[API reference].
 
 
+=== Limitations
+
+Topological Link Prediction algorithms are not yet supported.
```
soerenreichardt, line 260: What's the reason for this?

soerenreichardt (approved):

doc/modules/ROOT/pages/gds-session.adoc
FlorentinD: The problem is, that these GDS functions work directly over 2 nodes on the db and not the GDS graph. 

We would have to execute them on the db or fetch the relationships on demand.
As this feature is untouched since GDS 1.0, we want to wait for user requests.

graphdatascience/utils/util_remote_proc_runner.py
FlorentinD: They are also on the `util_proc_runner` but there we use the existing GDS functions such as `gds.util.asNode`

PR #697, opened 2024-07-30
Catch invalid db tier on list-session

FlorentinD:
currently we iterate over all instances to find the right session by name.


RafalSkolasinski (approved):
LGTM

Side note: ideally we would base logic on explicit error codes - if HTTP status codes are not enough we could introduce "error_code" into the response message.

PR #710, opened 2024-08-13
Pass dbid only when needed to aura-api

FlorentinD:
latest changes in Aura allow to not always send the instanceId along.
Also the uniqenuess constraint is checked inside Aura now.

### DIFF
--- a/graphdatascience/session/aura_api.py
+++ b/graphdatascience/session/aura_api.py
@@ -71,21 +71,23 @@ class AuraApi:
 
         return SessionDetails.fromJson(response.json()["data"])
 
-    def list_session(self, session_id: str, dbid: str) -> Optional[SessionDetails]:
-        response = self._request_session.get(
-            f"{self._base_uri}/v1beta5/data-science/sessions/{session_id}?instanceId={dbid}",
-        )
-
-        if response.status_code == 404:
-            return None
+    def list_session(self, session_id: str) -> Optional[SessionDetails]:
+        response = self._request_session.get(f"{self._base_uri}/v1beta5/data-science/sessions/{session_id}")
 
         self._check_resp(response)
 
         return SessionDetails.fromJson(response.json()["data"])
 
-    def list_sessions(self, dbid: str) -> List[SessionDetails]:
+    def list_sessions(self, dbid: Optional[str] = None) -> List[SessionDetails]:
+        params = {
+            "tenantId": self._tenant_id,
+        }
+
+        if dbid:
+            params["instanceId"] = dbid
+
         response = self._request_session.get(
-            f"{self._base_uri}/v1beta5/data-science/sessions?instanceId={dbid}",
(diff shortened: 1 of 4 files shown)


graphdatascience/session/aura_api.py
```diff
@@ -71,21 +71,23 @@ def create_session(self, name: str, dbid: str, pwd: str, memory: SessionMemoryVa
...
+        params = {
+            "tenantId": self._tenant_id,
+        }
+
+        if dbid:
+            params["instanceId"] = dbid
+
         response = self._request_session.get(
-            f"{self._base_uri}/v1beta5/data-science/sessions?instanceId={dbid}",
+            f"{self._base_uri}/v1beta5/data-science/sessions?instanceId={dbid}", params=params
```
DarthMax, line 90:
```suggestion
            f"{self._base_uri}/v1beta5/data-science/sessions?", params=params
```

graphdatascience/session/aura_api.py
```diff
@@ -71,21 +71,23 @@ def create_session(self, name: str, dbid: str, pwd: str, memory: SessionMemoryVa
...
         return SessionDetails.fromJson(response.json()["data"])
 
-    def list_session(self, session_id: str, dbid: str) -> Optional[SessionDetails]:
-        response = self._request_session.get(
-            f"{self._base_uri}/v1beta5/data-science/sessions/{session_id}?instanceId={dbid}",
-        )
-
-        if response.status_code == 404:
-            return None
+    def list_session(self, session_id: str) -> Optional[SessionDetails]:
```
DarthMax, line 74: The is not super related to this PR, but I think list_session is a bad name, listing imho returns a list not a single item.

graphdatascience/session/dedicated_sessions.py
```diff
@@ -71,56 +71,26 @@ def get_or_create(
...
-                if e.status_code == 404 or (e.status_code == 400 and "of tier" in e.message):
-                    continue
-                raise e
+    def list(self, dbid: Optional[str] = None) -> List[SessionInfo]:
+        sessions: List[SessionDetails] = self._aura_api.list_sessions(dbid)
 
         return [SessionInfo.from_session_details(i) for i in sessions]
 
-    def _find_existing_session(self, session_name: str, dbid: str) -> Optional[SessionDetails]:
+    def _find_existing_session(self, session_name: str) -> Optional[SessionDetails]:
```
DarthMax, line 87: I guess we could make the dbid here optional as well. It would speed up the filter by a bit should the user have many many sessions

DarthMax (approved):
Finally! I guess we only need to change that one URL

graphdatascience/session/dedicated_sessions.py
FlorentinD: would be nice. I would like to do this once we dont have AuraDS based sessions in the client anymore

PR #711, opened 2024-08-14
Fetch protocol version and use versioned remote procedures

soerenreichardt:


graphdatascience/query_runner/aura_db_query_runner.py
```diff
@@ -50,8 +53,15 @@ def call_procedure(
         if params is None:
             params = CallParameters()
 
-        if AuraDbQueryRunner.GDS_REMOTE_PROJECTION_PROC_NAME == endpoint:
-            return self._remote_projection(endpoint, params, yields, database, logging)
+        if AuraDbQueryRunner.GDS_REMOTE_PROJECTION_PROC_NAME in endpoint:
+            if self._endpoint_has_supported_version(endpoint):
+                return self._remote_projection(endpoint, params, yields, database, logging)
```
Mats-SX, line 58: so in this PR we send the same remote projection query regardless of protocol version?

graphdatascience/query_runner/aura_db_query_runner.py
```diff
@@ -120,32 +130,43 @@ def _remote_write_back(
...
-            "jobId": job_id,
-            "arrowConfiguration": db_arrow_config,
-        }
+        graph_name = params["graph_name"]
+        if ProtocolVersion.V2 in self._protocol_versions:
+            db_write_proc_params = self._write_back_params_v2(graph_name, job_id, db_arrow_config, config)
+            protocol_version = ProtocolVersion.V2
+        elif ProtocolVersion.V1 in self._protocol_versions:
+            db_write_proc_params = self._write_back_params_v1(graph_name, job_id, db_arrow_config)
+            protocol_version = ProtocolVersion.V1
```
Mats-SX, line 154: I was expecting to see something similar to this for projections as well.

Mats-SX (changes requested):
I think we need to think about the error cases and what instruction we should give the user in those cases.

Also, we need to run different queries for the different protocol versions also when projecting, right?

graphdatascience/query_runner/aura_db_query_runner.py
```diff
@@ -50,8 +53,15 @@ def call_procedure(
...
-        if AuraDbQueryRunner.GDS_REMOTE_PROJECTION_PROC_NAME == endpoint:
-            return self._remote_projection(endpoint, params, yields, database, logging)
+        if AuraDbQueryRunner.GDS_REMOTE_PROJECTION_PROC_NAME in endpoint:
+            if self._endpoint_has_supported_version(endpoint):
+                return self._remote_projection(endpoint, params, yields, database, logging)
+            else:
+                raise RuntimeError(
+                    f"Unsupported procedure: {endpoint}. \
+                    The database supports the following procedure protocol versions: {self._protocol_versions}. \
+                    Please use a matching project procedure."
```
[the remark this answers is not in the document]
soerenreichardt: No you're right. But we should instead point out that the user should update it's client version.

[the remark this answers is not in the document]
soerenreichardt: I also removed the sentence above. The user should not be informed about any versions, as they are purely internal

graphdatascience/query_runner/aura_db_query_runner.py
```diff
@@ -120,32 +130,43 @@ def _remote_write_back(
         logging: bool = False,
         custom_error: bool = True,
     ) -> DataFrame:
-        if params["config"] is None:
-            params["config"] = {}
```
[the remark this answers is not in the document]
soerenreichardt: I missed that we pass the whole `params` dict in the procedure call below. Fixed it now.

graphdatascience/query_runner/aura_db_query_runner.py
soerenreichardt: Yes, you're right I confused a bit which of the projection procedures and aggregation functions is called in which order 🤷

graphdatascience/query_runner/aura_db_query_runner.py
soerenreichardt: Fixed now

graphdatascience/session/aura_graph_data_science.py
```diff
@@ -194,3 +202,20 @@ def close(self) -> None:
         Close the GraphDataScience object and release any resources held by it.
         """
         self._query_runner.close()
+
+    def _db_protocol_versions(self) -> List[ProtocolVersion]:
```
Mats-SX, line 206: we should test this behaviour

graphdatascience/query_runner/aura_db_query_runner.py
```diff
@@ -123,29 +199,46 @@ def _remote_write_back(
...
-        }
+        self._inject_arrow_config(db_arrow_config)
+
+        graph_name = params["graph_name"]
+        if ProtocolVersion.V2 in self._server_protocol_versions:
+            db_write_proc_params = self._write_back_params_v2(graph_name, job_id, db_arrow_config, config)
+            protocol_version = ProtocolVersion.V2
+        elif ProtocolVersion.V1 in self._server_protocol_versions:
+            db_write_proc_params = self._write_back_params_v1(
+                graph_name, self._gds_query_runner.database(), job_id, db_arrow_config
```
Mats-SX, line 224:
```suggestion
                graph_name, self._db_query_runner.database(), job_id, db_arrow_config
```

Mats-SX (approved):

Mats-SX (changes requested):
Before we merge, we need to manually integration test with 5.24-dev

Mats-SX (approved):

graphdatascience/tests/integration/test_graph_construct.py
```diff
@@ -499,8 +499,6 @@ def test_graph_construct_with_arrow_no_db() -> None:
         gds.close()
         pytest.skip("Arrow server not enabled")
 
-    assert not gds.database()
```
Mats-SX, line 502: let's revert this

PR #712, opened 2024-08-20
Support for self-managed DBMS sources of GDS Sessions (take 3)

FlorentinD:
enable creating sessions against stand-alone

Co-Authored-By: Rafal Skolasinski <rafal.skolasinski@neo4j.com>

### DIFF
--- a/graphdatascience/session/__init__.py
+++ b/graphdatascience/session/__init__.py
@@ -8,6 +8,7 @@ __all__ = [
     "GdsSessions",
     "SessionInfo",
     "DbmsConnectionInfo",
+    "CloudLocation",
     "AuraAPICredentials",
     "SessionMemory",
     "SessionMemoryValue",
--- a/graphdatascience/session/gds_sessions.py
+++ b/graphdatascience/session/gds_sessions.py
@@ -77,7 +77,7 @@ class GdsSessions:
         self,
         session_name: str,
         memory: SessionMemory,
-        # TODO support TTL and cloud_location
+        # TODO support TTL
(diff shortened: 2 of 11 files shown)


FlorentinD:
[[🍓] Enable client to create sessions against self-managed dbms](https://trello.com/c/8FRF05Jk/9295-%F0%9F%8D%93-enable-client-to-create-sessions-against-self-managed-dbms)

graphdatascience/session/__init__.py
```diff
@@ -8,6 +8,7 @@
     "GdsSessions",
     "SessionInfo",
     "DbmsConnectionInfo",
+    "CloudLocation",
```
FlorentinD, line 11: lets make sure we have API ref as well

graphdatascience/session/gds_sessions.py
```diff
@@ -70,7 +71,9 @@ def get_or_create(
         self,
         session_name: str,
         memory: SessionMemory,
+        # TODO support TTL and cloud_location
         db_connection: DbmsConnectionInfo,
+        cloud_location: Optional[CloudLocation] = None,
```
FlorentinD, line 76: lets add a notebook example (making sure its part of CI but also for user-visibility)

graphdatascience/session/__init__.py
RafalSkolasinski: done

graphdatascience/session/gds_sessions.py
FlorentinD: -- will do in follow up PR

PR #713, opened 2024-08-21
Avoid reading env in AuraAPI constructor

FlorentinD:
makes it harder to reuse.
Also simplifies the constructor

### DIFF
--- a/graphdatascience/session/aura_api.py
+++ b/graphdatascience/session/aura_api.py
@@ -32,6 +32,8 @@ class AuraApiError(Exception):
 
 class AuraApi:
 
+    API_VERSION = "v1beta5"
+
     def __init__(
         self, client_id: str, client_secret: str, tenant_id: Optional[str] = None, aura_env: Optional[str] = None
     ) -> None:
@@ -68,7 +70,7 @@ class AuraApi:
 
     def create_session(self, name: str, dbid: str, pwd: str, memory: SessionMemoryValue) -> SessionDetails:
         response = self._request_session.post(
-            f"{self._base_uri}/v1beta5/data-science/sessions",
+            f"{self._base_uri}/{AuraApi.API_VERSION}/data-science/sessions",
             json={"name": name, "instance_id": dbid, "password": pwd, "memory": memory.value},
         )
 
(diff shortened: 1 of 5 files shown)


graphdatascience/session/aura_api.py
```diff
@@ -32,15 +32,22 @@ def __init__(self, message: str, status_code: int):
...
 class AuraApi:
-    def __init__(self, client_id: str, client_secret: str, tenant_id: Optional[str] = None) -> None:
-        self._aura_env = os.environ.get("AURA_ENV")
 
-        if not self._aura_env or self._aura_env == "production":
-            self._base_uri = "https://api.neo4j.io"
-        elif self._aura_env == "staging":
-            self._base_uri = "https://api-staging.neo4j.io"
+    @staticmethod
+    def create(client_id: str, client_secret: str, tenant_id: Optional[str] = None) -> AuraApi:
```
RafalSkolasinski, line 37: would expect `__init__` to be first method, even before staticmethods

graphdatascience/session/gds_sessions.py
```diff
@@ -42,7 +42,7 @@ def __init__(self, api_credentials: AuraAPICredentials) -> None:
         Args:
             api_credentials (AuraAPICredentials): The Aura API credentials used for establishing a connection.
         """
-        aura_api = AuraApi(api_credentials.client_id, api_credentials.client_secret, api_credentials.tenant_id)
+        aura_api = AuraApi.create(api_credentials.client_id, api_credentials.client_secret, api_credentials.tenant_id)
```
RafalSkolasinski, line 45: 💡 could keep using the constructor and instead of `.create` method provide a helper that determines URI from environment:
```python
aura_api = AuraApi(base_uri=base_uri(env="..."), ...)
```

RafalSkolasinski: or 
```python
base_uri = determine_base_uri(env="...")
aura_api = AuraApi(base_uri=base_uri, ...)
```

FlorentinD: would also work, but I am slightly leaning towards the static method. Nearly always you want to have the same logic.

RafalSkolasinski: Sure, in this case I'd probably use `classmethod` instead. Name `new` would also be shorter ;) 
```
    @classmethod
    def create(cls: Type[AuraApi], client_id: str, client_secret: str, tenant_id: Optional[str] = None) -> AuraApi:
        ...
        return cls(base_uri, client_id, client_secret, tenant_id)
```

PR #717, opened 2024-08-21
sphinx fail on error

FlorentinD:
- **Fail on sphinx build errors**

### DIFF
--- a/graphdatascience/graph_data_science.py
+++ b/graphdatascience/graph_data_science.py
@@ -49,11 +49,12 @@ class GraphDataScience(DirectEndpoints, UncallableNamespace):
         database: Optional[str], default None
             The Neo4j database to query against.
         arrow : Union[str, bool], default True
-            Arrow connection information. This is either a bool or a string.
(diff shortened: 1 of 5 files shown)


graphdatascience/graph_data_science.py
```diff
@@ -48,12 +48,14 @@ def __init__(
             to a Neo4j AuraDS instance.
         database: Optional[str], default None
             The Neo4j database to query against.
+            # TODO whats off with the arrow part?
         arrow : Union[str, bool], default True
             Arrow connection information. This is either a bool or a string.
```
nvitucci, line 53:
```suggestion
            Arrow connection information. This is either a string or a bool.
```
Nitpick, to match the order in the Union type and in the list.

graphdatascience/graph_data_science.py
```diff
@@ -48,12 +48,14 @@ def __init__(
             to a Neo4j AuraDS instance.
         database: Optional[str], default None
             The Neo4j database to query against.
+            # TODO whats off with the arrow part?
```
nvitucci, line 51: Can this line be removed now?

PR #719, opened 2024-08-22
sessions for self-managed dbms docs (manual + notebook)

FlorentinD:
based on #712

PR #724, opened 2024-08-26
Make TTL configurable

DarthMax:
~~Based on #719~~

### DIFF
--- a/doc/modules/ROOT/pages/gds-session.adoc
+++ b/doc/modules/ROOT/pages/gds-session.adoc
@@ -62,7 +62,9 @@ You can use the `sessions.estimate()` method to estimate the size required.
 This is a `DbmsConnectionInfo` object that contains the URI of an Neo4j instance, a username, and a password.
 
 - **TTL**.
-This optional parameter specifies the time-to-live of the session. The session will be automatically deleted after the TTL has expired after the last user activity.
+This optional parameter specifies the time-to-live of the session.
+The session will be automatically deleted if the session was unused for the provided duration.
+Usage is defined as the computation of an algorithm or the projection of a graph.
 
 - **Cloud location**.
 This is a `CloudLocation` object that specifies the cloud provider and region where the GDS Session will run. Required if the DBMS connection is for a self-managed database.
@@ -94,6 +96,7 @@ gds = sessions.get_or_create(
 .Creating a GDS Session for a self-managed Neo4j database:
 [source,python,role=no-test]
 ----
+from datetime import timedelta
 from graphdatascience.session import DbmsConnectionInfo, AlgorithmCategory, CloudLocation
 
 name = "my-new-session-sm"
@@ -109,6 +112,7 @@ gds = sessions.get_or_create(
     session_name=name,
     memory=memory,
     db_connection=db_connection_info,
+    ttl=timedelta(hours=5),
     cloud_location=cloud_location,
 )
 ----
--- a/doc/modules/ROOT/pages/gds-session.adoc
+++ b/doc/modules/ROOT/pages/gds-session.adoc
@@ -61,6 +61,9 @@ You can use the `sessions.estimate()` method to estimate the size required.
 - **DBMS connection**.
 This is a `DbmsConnectionInfo` object that contains the URI of an Neo4j instance, a username, and a password.
 
+- **TTL**.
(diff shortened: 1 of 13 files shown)


DarthMax:
[[🍓] Enable client to create sessions against self-managed dbms](https://trello.com/c/8FRF05Jk/9295-%F0%9F%8D%93-enable-client-to-create-sessions-against-self-managed-dbms)

doc/modules/ROOT/pages/gds-session.adoc
```diff
@@ -61,6 +61,9 @@ You can use the `sessions.estimate()` method to estimate the size required.
 - **DBMS connection**.
 This is a `DbmsConnectionInfo` object that contains the URI of an Neo4j instance, a username, and a password.
 
+- **TTL**.
+This optional parameter specifies the time-to-live of the session. The session will be automatically deleted after the TTL has expired after the last user activity.
```
FlorentinD, line 65:
```suggestion
This optional parameter specifies the time-to-live of the session. The session will be automatically deleted if the session was unused for the provided duration.
```

doc/modules/ROOT/pages/gds-session.adoc
```diff
@@ -83,6 +87,7 @@ gds = sessions.get_or_create(
     session_name=name,
     memory=memory,
     db_connection=db_connection_info,
+    ttl=timedelta(hours=5),
```
FlorentinD, line 90: how about we also add it to the notebook?

graphdatascience/session/aura_api.py
```diff
@@ -68,11 +68,13 @@ def base_uri(aura_env: Optional[str] = None) -> str:
...
-    def create_attached_session(self, name: str, dbid: str, pwd: str, memory: SessionMemoryValue) -> SessionDetails:
-        return self._create_session(name, pwd, memory, instance_id=dbid)
+    def create_attached_session(
+        self, name: str, dbid: str, pwd: str, memory: SessionMemoryValue, ttl: Optional[str]
+    ) -> SessionDetails:
+        return self._create_session(name, pwd, memory, instance_id=dbid, ttl=ttl)
 
     def create_standalone_session(
-        self, name: str, pwd: str, memory: SessionMemoryValue, cloud_location: CloudLocation
+        self, name: str, pwd: str, memory: SessionMemoryValue, ttl: Optional[str], cloud_location: CloudLocation
```
FlorentinD, line 77: similar to memory, we could pass the ttl as the typed timedelta here

graphdatascience/session/dedicated_sessions.py
```diff
@@ -55,10 +56,11 @@ def get_or_create(
             self._check_expiry_date(existing_session)
             self._check_memory_configuration(existing_session, memory.value)
             self._check_dbid(existing_session, dbid)
+            # TODO check cloud location?
```
FlorentinD, line 59: will be covered in https://trello.com/c/8oUk8poI

graphdatascience/session/gds_sessions.py
```diff
@@ -89,12 +90,13 @@ def get_or_create(
...
             memory (SessionMemory): The size of the session specified by memory.
             db_connection (DbmsConnectionInfo): The database connection information.
+            ttl: Optional[timedelta]: The sessions time to live after inactivity in seconds.
             cloud_location (Optional[CloudLocation]): The cloud location. Required if the GDS session is for a self-managed database.
 
         Returns:
             AuraGraphDataScience: The session.
         """
-        return self._impl.get_or_create(session_name, memory, db_connection, cloud_location=cloud_location)
+        return self._impl.get_or_create(session_name, memory, db_connection, ttl=ttl, cloud_location=cloud_location)
```
FlorentinD, line 99: should throw if its used for AuraDS based sessions

FlorentinD (approved):
Looks good.

Main suggestion is to add it to the notebooks as well for user visibility (and CI testing)

Mats-SX (changes requested):
I think we should define what activities refresh the TTL. But other than that, looks very good.

doc/modules/ROOT/pages/gds-session.adoc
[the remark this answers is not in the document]
DarthMax: We are currently not describing the example at all. I think that if we add this line, we would also have to describe the rest of the parameters, which would be a duplication of the description above.

PR #729, opened 2024-08-27
Return cloud_location and mark instance_id as optional

FlorentinD:


FlorentinD:
[[🍓] Return cloud provider and region](https://trello.com/c/8oUk8poI/9742-%F0%9F%8D%93-return-cloud-provider-and-region)

DarthMax (approved):

PR #733, opened 2024-08-28
fix(attached-sessions): fix misspelled instanceId with correct instan…

RafalSkolasinski:
Fixing accidental type introduced in https://github.com/neo4j/graph-data-science-client/pull/724/files#diff-81be788a211e438e22d728f9450366110ef62b6971b76cc36cce01d14ba3cd6f and addiing tests to ensure it will be caught early on in the future.

### DIFF
--- a/graphdatascience/session/aura_api.py
+++ b/graphdatascience/session/aura_api.py
@@ -97,7 +97,7 @@ class AuraApi:
         json = {"name": name, "password": pwd, "memory": memory.value}
 
         if dbid:
-            json["instanceId"] = dbid
+            json["instance_id"] = dbid
 
         if ttl:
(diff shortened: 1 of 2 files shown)


graphdatascience/session/aura_api.py
```diff
@@ -97,7 +97,7 @@ def create_session(
         json = {"name": name, "password": pwd, "memory": memory.value}
 
         if dbid:
-            json["instanceId"] = dbid
+            json["instance_id"] = dbid
```
FlorentinD, line 100: saw this one too late. (pushed a fix to main)

FlorentinD (approved):
Good additional testing 👍

PR #747, opened 2024-09-04
Local progress task store

soerenreichardt:
This PR adds a local task store that will be consolidated first when we try to query for progress updates given a job-id. As remote projection does not have a job-id right now we cannot query for any progress in the session. We use this local store instead that returns a task name and `n/a` progress, which is the equivalent of an unknown task volume

### DIFF
--- a/graphdatascience/query_runner/progress/progress_provider.py
+++ b/graphdatascience/query_runner/progress/progress_provider.py
@@ -1,0 +1,16 @@
+from abc import ABC, abstractmethod
+from dataclasses import dataclass
+from typing import Optional
+
+
+@dataclass
+class TaskWithProgress:
+    task_name: str
+    progress_percent: str
+
+
+class ProgressProvider(ABC):
+    @abstractmethod
+    def root_task_with_progress(self, job_id: str, database: Optional[str] = None) -> TaskWithProgress:
+        """Return the task with progress for the given job_id."""
+        pass
--- a/graphdatascience/query_runner/progress/query_progress_logger.py
+++ b/graphdatascience/query_runner/progress/query_progress_logger.py
@@ -6,11 +6,11 @@ from uuid import uuid4
 from pandas import DataFrame
 from tqdm.auto import tqdm
 
-from ..server_version.server_version import ServerVersion
+from ...server_version.server_version import ServerVersion
+from .progress_provider import ProgressProvider
+from .query_progress_provider import CypherQueryFunction, QueryProgressProvider, ServerVersionFunction
+from .static_progress_provider import StaticProgressProvider, StaticProgressStore
 
-# takes a query str, optional db str and returns the result as a DataFrame
-CypherQueryFunction = Callable[[str, Optional[str]], DataFrame]
-ServerVersionFunction = Callable[[], ServerVersion]
 DataFrameProducer = Callable[[], DataFrame]
 
 
@@ -24,6 +24,8 @@ class QueryProgressLogger:
     ):
         self._run_cypher_func = run_cypher_func
         self._server_version_func = server_version_func
+        self._static_progress_provider = StaticProgressProvider()
+        self._query_progress_provider = QueryProgressProvider(run_cypher_func, server_version_func)
 
     @staticmethod
     def extract_or_create_job_id(params: Dict[str, Any]) -> str:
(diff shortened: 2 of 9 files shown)


soerenreichardt:
[[🍓] Progress logging for remote projection](https://trello.com/c/n7CVJ5PI/9771-%F0%9F%8D%93-progress-logging-for-remote-projection)

graphdatascience/query_runner/progress/progress_provider.py
```diff
@@ -0,0 +1,16 @@
+from abc import ABC, abstractmethod
+from dataclasses import dataclass
+from typing import Optional
+
+
+@dataclass
+class TaskWithProgress:
+    task_name: str
+    progress_percent: str
```
Mats-SX, line 9: should it be a `float` instead?

graphdatascience/query_runner/progress/query_progress_logger.py
```diff
@@ -45,31 +47,47 @@ def run_with_progress_logging(
...
                 raise future.exception()  # type: ignore
             else:
                 return future.result()
 
-    def _log(self, future: "Future[Any]", job_id: str, database: Optional[str] = None) -> None:
+    def _select_progress_provider(self, job_id) -> ProgressProvider:
+        return (
+            self._static_progress_provider
+            if StaticProgressStore.contains_job_id(job_id)
+            else self._query_progress_provider
```
Mats-SX, line 68: pretty clever

graphdatascience/query_runner/progress/progress_provider.py
Mats-SX: no -- it is sometimes `"n/a"` so that wouldn't work

graphdatascience/query_runner/progress/static_progress_provider.py
```diff
@@ -0,0 +1,27 @@
...
+
+    @staticmethod
+    def contains_job_id(job_id: str) -> bool:
+        return job_id in StaticProgressStore._progress_store
+
+
+class StaticProgressProvider(ProgressProvider):
+    def root_task_with_progress(self, job_id: str, database: Optional[str] = None) -> TaskWithProgress:
+        if not StaticProgressStore.contains_job_id(job_id):
+            raise ValueError(f"Task with job id {job_id} not found in progress store")
```
Mats-SX, line 25: I checked this error type (that we use a lot in this repo) and I don't think it's correct to use it here. I think instead let's just use `Exception`
```suggestion
            raise Exception(f"Task with job id `{job_id}` not found in progress store")
```

Mats-SX: In fact, it's not correct to use in many of our current usages.

graphdatascience/query_runner/progress/static_progress_provider.py
```diff
@@ -0,0 +1,27 @@
+from typing import Optional
+
+from .progress_provider import ProgressProvider, TaskWithProgress
+
+
+class StaticProgressStore:
```
Mats-SX, line 6: this class isn't meant to stick around, but you never know... a unit test or two could be useful I think

graphdatascience/query_runner/progress/static_progress_provider.py
```diff
@@ -0,0 +1,27 @@
...
+    @staticmethod
+    def get_task_with_volume(job_id: str) -> TaskWithProgress:
+        return StaticProgressStore._progress_store[job_id]
+
+    @staticmethod
+    def contains_job_id(job_id: str) -> bool:
+        return job_id in StaticProgressStore._progress_store
+
+
+class StaticProgressProvider(ProgressProvider):
```
Mats-SX, line 22: also for this one

graphdatascience/query_runner/progress/static_progress_provider.py
Mats-SX: although unit tests of these static singletons can be a bit annoying

graphdatascience/query_runner/session_query_runner.py
```diff
@@ -37,7 +38,8 @@ def __init__(
         self._gds_arrow_client = arrow_client
         self._resolved_protocol_version = ProtocolVersionResolver(db_query_runner).resolve()
         self._progress_logger = QueryProgressLogger(
-            self._gds_query_runner.run_cypher, self._gds_query_runner.server_version
+            lambda query, database: self._gds_query_runner.run_cypher(query=query, database=database),
```
Mats-SX, line 41: did you have to make this `lambda` or was it a choice?

Mats-SX (commented):
Looks good. Defensively I'd like to see a unit test for the static store and its logger. Would be simple to add I'm hoping.

graphdatascience/query_runner/session_query_runner.py
soerenreichardt: It could have been a method reference like the one you did for the `neo4j_query_runner` as well

soerenreichardt: Yes, but it contained the bug that it relied on positional arguments rather than named arguments, so we would pass the database value into the `params` argument on `run_cypher`

PR #748, opened 2024-09-09
Retrieve ArrowInfo only once

FlorentinD:
Reduce number of connections for client init

### DIFF
--- a/graphdatascience/query_runner/arrow_info.py
+++ b/graphdatascience/query_runner/arrow_info.py
@@ -20,7 +20,7 @@ class ArrowInfo:
         if query_runner.server_version() > ServerVersion(2, 6, 0):
             debugYields.append("versions")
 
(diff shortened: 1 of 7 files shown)


graphdatascience/query_runner/arrow_info.py
```diff
@@ -0,0 +1,32 @@
...
+
+    @staticmethod
+    def create(query_runner: QueryRunner) -> ArrowInfo:
+        debugYields = ["listenAddress", "enabled", "running"]
+        if query_runner.server_version() > ServerVersion(2, 6, 0):
+            debugYields.append("versions")
+
+        procResult = query_runner.call_procedure(
+            endpoint="gds.debug.arrow", custom_error=False, yields=debugYields
+        ).iloc[0]
```
Mats-SX, line 25: `iloc` -- that's a first for me

Mats-SX (approved):
This is good, approved, but I think we can make ArrowQueryRunner a bit simpler by checking the ArrowInfo before we call `create`

PR #750, opened 2024-09-11
Refactor protocol implementations

soerenreichardt:
In the V3 protocol versions also the behaviour of how to call the procedures will change in a way that the execution will be asynchronous. This PR prepares the code to make the V3 implementation easier.

### DIFF
--- a/graphdatascience/query_runner/protocol/project_protocols.py
+++ b/graphdatascience/query_runner/protocol/project_protocols.py
@@ -1,0 +1,88 @@
+from abc import ABC, abstractmethod
+from typing import Any, Dict, List, Optional
+
+from pandas import DataFrame
+
+from graphdatascience import QueryRunner
+from graphdatascience.call_parameters import CallParameters
+from graphdatascience.session.dbms.protocol_version import ProtocolVersion
+
+
+class ProjectProtocol(ABC):
+    @abstractmethod
+    def project_params(
+        self, graph_name: str, query: str, params: Dict[str, Any], arrow_config: Dict[str, Any]
+    ) -> CallParameters:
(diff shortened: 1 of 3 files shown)


soerenreichardt:
[[🍓] Make write-back async in dbms procedure](https://trello.com/c/fAvBtIHi/9789-%F0%9F%8D%93-make-write-back-async-in-dbms-procedure)

graphdatascience/query_runner/protocol/project_protocols.py
```diff
@@ -0,0 +1,88 @@
...
+        yields: Optional[List[str]] = None,
+        database: Optional[str] = None,
+        logging: bool = False,
+    ) -> DataFrame:
+        """Returns the procedure name for the corresponding protocol version."""
+        pass
+
+    @staticmethod
+    def select(protocol_version: ProtocolVersion) -> "ProjectProtocol":
+        return {ProtocolVersion.V1: ProjectProtocolV1(), ProtocolVersion.V2: ProjectProtocolV2()}[protocol_version]
```
DarthMax, line 34: I think I would prefer a match statement here. Not sure how clever the type checker is but maybe this would then tell us if we forgot a version?

DarthMax (approved):
Looks very clean

graphdatascience/query_runner/protocol/project_protocols.py
soerenreichardt: I wanted to use that, but `match` is only supported from python 3.10 on. We currently require python `>=3.8`

DarthMax: Argh damn it :D

PR #756, opened 2024-09-25
Project/write protocols V3

soerenreichardt:


graphdatascience/query_runner/protocol/project_protocols.py
```diff
@@ -86,3 +92,42 @@ def run_projection(
...
+                ProtocolVersion.V3.versioned_procedure_name(endpoint), params, yields, database, logging, False
+            )
+
+        while True:
+            result = project_fn().squeeze()
+            if result["status"] == Status.DONE.name:
+                return result
+
+            # wait for 100 ms, then retry
+            time.sleep(0.1)
```
soerenreichardt, line 133: maybe this should be some form of exponential backoff or similar?

graphdatascience/query_runner/protocol/project_protocols.py
```diff
@@ -31,12 +33,16 @@ def run_projection(
 
     @staticmethod
     def select(protocol_version: ProtocolVersion) -> "ProjectProtocol":
-        return {ProtocolVersion.V1: ProjectProtocolV1(), ProtocolVersion.V2: ProjectProtocolV2()}[protocol_version]
+        return {
+            ProtocolVersion.V1: ProjectProtocolV1(),
+            ProtocolVersion.V2: ProjectProtocolV2(),
+            ProtocolVersion.V3: ProjectProtocolV3(),
+        }[protocol_version]
```
FlorentinD, line 40: not changed in this PR, but we could improve the error thrown to the user.
Could happen if the client does not know the servers version (client too old).

```suggestion
        return {
            ProtocolVersion.V1: ProjectProtocolV1(),
            ProtocolVersion.V2: ProjectProtocolV2(),
            ProtocolVersion.V3: ProjectProtocolV3(),
        }.get(protocol_version)
```
this would return an Optional[ProjectProtocol]

graphdatascience/query_runner/protocol/project_protocols.py
```diff
@@ -31,12 +33,16 @@ def run_projection(
 
     @staticmethod
     def select(protocol_version: ProtocolVersion) -> "ProjectProtocol":
```
FlorentinD, line 35: you dont need to use the string version if you import `from __future__ import annotations` at the top.

graphdatascience/query_runner/protocol/write_protocols.py
```diff
@@ -94,3 +100,35 @@ def run_write_back(
...
+        database: Optional[str] = None,
+    ) -> CallParameters:
+        return RemoteWriteBackV2().write_back_params(graph_name, job_id, config, arrow_config, database)
+
+    def run_write_back(
+        self, query_runner: QueryRunner, parameters: CallParameters, yields: Optional[List[str]]
+    ) -> DataFrame:
+        def is_not_completed(result: DataFrame) -> bool:
+            result = write_fn().squeeze()
+            return result["status"] != Status.COMPLETED.name
```
FlorentinD, line 121: could extract this into `status.py` as it seems a very generic function :)

requirements/base/base.txt
```diff
@@ -7,3 +7,4 @@ textdistance >= 4.0, < 5.0
 tqdm >= 4.0, < 5.0
 typing-extensions >= 4.0, < 5.0
 requests
+tenacity >= 9.0
```
FlorentinD, line 10: @RafalSkolasinski also suggested using tenacity.

Good to keep it in mind when reviewing existing retry logic or adding new one 👍

graphdatascience/query_runner/protocol/write_protocols.py
```diff
@@ -94,3 +100,35 @@ def run_write_back(
...
+
+        @retry(retry=retry_if_result(is_not_completed), wait=wait_incrementing(start=0.2, increment=0.2, max=2))
```
FlorentinD, line 123: any particular reason for these retry values?

graphdatascience/query_runner/protocol/project_protocols.py
```diff
@@ -86,3 +92,41 @@ def run_projection(
...
+    ) -> DataFrame:
+        def is_not_done(result: DataFrame) -> bool:
+            result = project_fn().squeeze()
+            return result["status"] != Status.DONE.name
+
+        @retry(retry=retry_if_result(is_not_done), wait=wait_incrementing(start=0.2, increment=0.2, max=2))
+        def project_fn() -> DataFrame:
+            return query_runner.call_procedure(
+                ProtocolVersion.V3.versioned_procedure_name(endpoint), params, yields, database, logging, False
+            )
```
FlorentinD, line 130: I did not follow this work to closely.
How do we avoid that the client will not send the previous batches again?

Or is it rather, that the Arrow client will continue with the projection even on a connection loss?
And we will run the cypher part up to the aggregation function, which find an existing projection and returns early?
Ideally we could check upfront if a projection with the job-id already exists?

FlorentinD (approved):
The changes look sane.
Just some minor comments and interested in how the projection retry works

graphdatascience/query_runner/protocol/write_protocols.py
soerenreichardt: Let's do that when we have more versions. Unfortunately someone didn't align with the existing return status of the write-back, so that the project procedure will return `DONE` instead, so we at least cannot share everything between write and project

graphdatascience/query_runner/protocol/write_protocols.py
soerenreichardt: Nope

graphdatascience/query_runner/protocol/project_protocols.py
soerenreichardt: The arrow part is unrelated to this change and will be implemented in another PR (we didn't implement it yet on the dbms or session side). What we did here is to guard the procedure execution from connection losses. Before, when the bolt connection to a project/write procedure was interrupted, the transaction of that procedure was killed and therefore the arrow connection, leading to a complete interruption of the work.
Now, the procedures will spawn tasks that run in the background instead, independent of transactions. When calling a procedure one of the following 2 things will happen:
1. If there is no process with the given job-id already running, spawn a process and return
2. If a process is running, check if it is terminated and return an appropriate status

Here in the python client we merely have to implement the status polling, everything else is handled within the procedure implementations.

graphdatascience/query_runner/protocol/project_protocols.py
soerenreichardt: Makes sense, but I think we should tackle that in a PR that is focused on the UX for outdated client versions, as you suggested

PR #762, opened 2024-10-02
Warn on old server version

FlorentinD:


FlorentinD:
[Check for min server version](https://trello.com/c/xtc5uKPw/9811-check-for-min-server-version)

graphdatascience/tests/unit/conftest.py
```diff
@@ -20,7 +20,7 @@
 from graphdatascience.session.dbms_connection_info import DbmsConnectionInfo
 
 # Should mirror the latest GDS server version under development.
-DEFAULT_SERVER_VERSION = ServerVersion(2, 6, 0)
+DEFAULT_SERVER_VERSION = ServerVersion(2, 10, 0)
```
Mats-SX, line 23: >under development

We are developing 2.12.0 of the plugin now. 2.11.0 is the latest stable.
Should this be 2.12 or 2.11?

graphdatascience/tests/unit/test_graph_construct.py
```diff
@@ -28,54 +26,22 @@ def test_graph_project_based_construct_without_arrow(runner: CollectingQueryRunn
...
-        "$relationship_query, "
-        "{readConcurrency: $read_concurrency, parameters: { nodes: $nodes, relationships: $relationships }})"
+    expected_query = (
+        "UNWIND $data AS data WITH data, "
+        "CASE WHEN data[6] THEN data[5] ELSE null END AS sourceNodeLabels, "
+        "CASE WHEN data[3] THEN data[2] ELSE null END AS relationshipType, "
+        "CASE WHEN data[10] THEN data[9] ELSE null END AS targetNodeId, "
+        "\nCASE WHEN data[8] THEN data[7] ELSE null END AS sourceNodeProperties, "
+        "\nCASE WHEN data[1] THEN data[0] ELSE null END AS relationshipProperties "
+        "RETURN gds.graph.project($graph_name, data[4], targetNodeId, {sourceNodeLabels: sourceNodeLabels, targetNodeLabels: NULL, sourceNodeProperties: sourceNodeProperties, targetNodeProperties: NULL, relationshipType: relationshipType, relationshipProperties: relationshipProperties}, $configuration)"
```
Mats-SX, line 36: did you make the client use new Cypher projection in this PR? I cannot see any changes that correspond to that.

Mats-SX: I am thinking that your removal of the parameterised ServerVersion(2.1.0) makes `graph.construct` take a different path through the code.

I think we should follow this work with removals of code that is parameterised on ServerVersion older than 2.6. That would remove the old Cypher projection from `graph.construct` is my hypothesis. Otherwise I don't understand how things hang together.

graphdatascience/tests/unit/test_graph_construct.py
```diff
@@ -28,54 +26,22 @@ def test_graph_project_based_construct_without_arrow(runner: CollectingQueryRunn
...
-            ServerVersion(2, 4, 0),
-            "",
-            ", targetNodeLabels: NULL",
-            ", targetNodeProperties: NULL",
-            "relationshipProperties",
-            ", ",
-        ),
-    ],
-    ids=["2.3.0 - Alpha Cypher Aggregation", "2.4.0 - New Cypher projection"],
-)
```
Mats-SX, line 70: nice to get rid of this messy thing

Mats-SX (approved):
Very nice initiative. I think `DEFAULT_SERVER_VERSION` should be something else but not sure what

graphdatascience/tests/unit/conftest.py
FlorentinD: now with monthly updates, this can get quite annoying to update.
As its only a version for unit-testing we could also just use a very high number

graphdatascience/tests/unit/test_graph_construct.py
FlorentinD: This is a unit-test and only worked based the old server-version.
Yes we can remove the old impl code as well, but IMO we should do this in 1.13.
We will raise awareness in 1.12 and act on 1.13? I assume many missed the docs about supported versions in the past

PR #776, opened 2024-10-29
Session - get_or_create: Reuse db driver

FlorentinD:
We can save a bit of time by reusing the driver we create to test the db credentials.

Also realized, that the sessions.estimate call was taking up to 5s.
By moving it into its own cell its better to see how fast get_or_create can get.


FlorentinD:
[[🍓] Python client - Avoid `Failed to write data to connection` messages](https://trello.com/c/IyJzO4SO/9967-%F0%9F%8D%93-python-client-avoid-failed-to-write-data-to-connection-messages)

PR #778, opened 2024-10-30
Fix v3 protocols

soerenreichardt:


soerenreichardt:
[Session API V3 projection bug](https://trello.com/c/3qiA49xz/9999-session-api-v3-projection-bug)

FlorentinD (approved):
👍

PR #781, opened 2024-11-12
Forward SessionError 

FlorentinD:
- **Parse errors and add to SessionDetails**
- **Fail on get_or_create session for SessionErrors**

### DIFF
--- a/graphdatascience/session/aura_api.py
+++ b/graphdatascience/session/aura_api.py
@@ -88,7 +88,7 @@ class AuraApi:
             base_uri = f"https://api-{aura_env}.neo4j-dev.io"
         return base_uri
 
-    def create_session(
+    def get_or_create_session(
         self,
         name: str,
         pwd: str,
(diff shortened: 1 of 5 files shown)


graphdatascience/session/aura_api.py
```diff
@@ -114,7 +115,9 @@ def create_session(
 
         self._check_resp(response)
 
-        return SessionDetails.from_json(response.json()["data"])
+        raw_json: Dict[str, Any] = response.json()
+
+        return SessionDetails.from_json(raw_json["data"], raw_json.get("errors", []))
```
RafalSkolasinski, line 120: What if `data` key is missing? Do we handle it somewhere else?

FlorentinD: nope. 
Arguably, we could also expect "errors" to be present. 
If its not, thats a breaking aura-api change.

Especially as this change is already live in Aura we could always expect errors to be present

RafalSkolasinski (approved):

PR #783, opened 2024-11-18
Bring back mypy

FlorentinD:
- **Bring back mypy**
- **Check for DictionaryArray + ignore some warnings**
- **Fix pandas typing issue for sort by columns**

### DIFF
--- a/graphdatascience/query_runner/gds_arrow_client.py
+++ b/graphdatascience/query_runner/gds_arrow_client.py
@@ -6,12 +6,13 @@ import re
 import time
 import warnings
 from dataclasses import dataclass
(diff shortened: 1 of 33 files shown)


graphdatascience/query_runner/gds_arrow_client.py
```diff
@@ -131,7 +132,7 @@ def request_token(self) -> Optional[str]:
     def get_node_properties(
         self,
         graph_name: str,
-        database: str,
+        database: Optional[str],
```
DarthMax, line 135: I don't think we should make this change. The database here is required.

DarthMax (commented):
Looks good for the most part. But the optional part for database params in GdsArrowClient needs to be discussed.

PR #784, opened 2024-11-22
Remove APOC dependency from notebook

Mats-SX:


Mats-SX:
[Rewrite node regression notebook to not use apoc](https://trello.com/c/GqJFjxn9/10031-rewrite-node-regression-notebook-to-not-use-apoc)

FlorentinD (approved):

PR #785, opened 2024-11-26
Drop Python 3.8 support

FlorentinD:
3.8 is EOL


tox.ini
```diff
@@ -1,19 +1,18 @@
 [tox]
 envlist =
-	py{38,39}-neo4j{4,5}-pandas{1,2}-main-{standard,encrypted},
-	py{38,39}-pandas1-pyarrow{11,12,13,14,15}-main-{standard,encrypted},
-	py310-pandas1-pyarrow{11,12,13,14,15}-main-{standard,encrypted},
-	py311-pandas1-pyarrow{11,12,13,14,15}-main-{standard,encrypted},
```
FlorentinD, line 6: we tested some old pyarrow versions the whole time.
according to base.txt we support `pyarrow >= 14.0.1, < 17.0`

PR #788, opened 2024-11-27
Reduce retries on fetching progress

FlorentinD:
Polling interval is 0.5s so connectivity retry we dont need to do for 10 minutes


soerenreichardt (approved):
👍

FlorentinD:
[[🍓] Report error to client in case of a pod crash](https://trello.com/c/fr6Zic4r/9849-%F0%9F%8D%93-report-error-to-client-in-case-of-a-pod-crash)

PR #789, opened 2024-11-27
Improve error message for connecction loss

FlorentinD:


FlorentinD:
[[🍓] Report error to client in case of a pod crash](https://trello.com/c/fr6Zic4r/9849-%F0%9F%8D%93-report-error-to-client-in-case-of-a-pod-crash)

DarthMax (approved):
:+1:

PR #800, opened 2024-12-04
Filter duplicate node_properties in params

FlorentinD:


FlorentinD:
[Handling of duplicate properties given in Arrow endpoints (stream nodeproperties)](https://trello.com/c/l9kq4KmG/9954-handling-of-duplicate-properties-given-in-arrow-endpoints-stream-nodeproperties)

graphdatascience/graph/graph_entity_ops_runner.py
```diff
@@ -104,6 +107,12 @@ def stream(
     ) -> DataFrame:
         self._namespace += ".stream"
 
+        # find if list contain duplicates
+        if len(set(node_properties)) != len(node_properties):
+            raise ValueError(f"Duplicate values in node_properties `{node_properties}`.")
+        if len(set(db_node_properties)) != len(db_node_properties):
+            raise ValueError(f"Duplicate values in db_node_properties `{db_node_properties}`.")
```
Mats-SX, line 114:
```suggestion
            raise ValueError(f"The provided db_node_properties contain duplicate property names: `{db_node_properties}`.")
```

graphdatascience/tests/integration/test_graph_ops.py
```diff
@@ -424,14 +424,23 @@ def test_graph_nodeProperty_stream_raise_error_with_duplicate_keys(gds: GraphDat
...
 
-@pytest.mark.enterprise
+def test_graph_nodeProperty_fail_on_duplicate_node_properties(gds: GraphDataScience) -> None:
+    G, _ = gds.graph.project(GRAPH_NAME, {"Node": {"properties": ["x", "y"]}}, "*")
+
+    with pytest.raises(ValueError, match="Duplicate values in db_node_properties."):
+        gds.graph.nodeProperty.stream(G, "x", db_node_properties=["z", "z", "name"], concurrency=2)
+
+
+# @pytest.mark.enterprise
```
Mats-SX, line 434: intended comment here or?

Mats-SX: if intended, we should delete the line

Mats-SX: ah it's arrow, so it should not be commented
```suggestion
@pytest.mark.enterprise
```

graphdatascience/tests/integration/test_graph_ops.py
```diff
@@ -510,7 +519,7 @@ def test_graph_nodeProperties_stream_with_arrow_separate_property_columns(gds: G
     result = gds.graph.nodeProperties.stream(
         G, ["x", "y"], db_node_properties=["z", "name"], separate_property_columns=True, concurrency=2
     )
-    assert list(result.keys()) == ["nodeId", "x", "y", "z", "name"]
+    assert set(result.keys()) == {"nodeId", "x", "y", "z", "name"}
```
Mats-SX, line 522: why did you change this assertion? if `result.keys()` contains a duplicate, then it will still pass this test. is that intentional?

Mats-SX (changes requested):
There's a mistakenly committed comment on one line that needs to be addressed.

Then I suggested a reworded error message, but that's optional.

graphdatascience/tests/integration/test_graph_ops.py
FlorentinD: i think we shouldnt care about the ordering of the columns here.

PR #804, opened 2024-12-18
fix: remove flight client from serialization

Emrehzl94:
This PR fixes serialization issues with the FlightClient in the GDSArrowClient class. Due to its non-serializability.

- Added a `_instantiate_flight_client()` method to create the `FlightClient` with proper configuration.
- Implemented a `_client()` method for lazy construction of the `FlightClient`, ensuring it’s only instantiated when needed.
- Modified the `__getstate__()` method to exclude the `FlightClient` from the serialized state.

### DIFF
--- a/graphdatascience/query_runner/gds_arrow_client.py
+++ b/graphdatascience/query_runner/gds_arrow_client.py
@@ -102,18 +102,12 @@ class GdsArrowClient:
         self._flight_client = self._instantiate_flight_client()
 
     def _instantiate_flight_client(self) -> flight.FlightClient:
-        print("inside instantiate_flight_client")
         location = flight.Location.for_grpc_tls(self._host, self._port) if self._encrypted else flight.Location.for_grpc_tcp(self._host, self._port)
-        print("after flight location")
         client_options: dict[str, Any] = {"disable_server_verification": self._disable_server_verification}
-        print("after client options init")
         if self._auth:
             client_options["middleware"] = [AuthFactory(self._auth_middleware), UserAgentFactory(useragent=self._user_agent)]
-            print("self. auth middleware")
         if self._tls_root_certs:
             client_options["tls_root_certs"] = self._tls_root_certs
-            print("self.tls roots")
-        print("returning flight client")
         return flight.FlightClient(location, **client_options)
 
     def connection_info(self) -> tuple[str, int]:
@@ -556,30 +550,13 @@ class GdsArrowClient:
         if "_flight_client" in state:
             del state["_flight_client"]
         return state
-    
-    # def copy(self) -> "GDSArrowClient":
-    #     client = GdsArrowClient(
-    #         self._host,
-    #         self._port,
-    #         self._auth,
-    #         self._encrypted,
-    #         self._disable_server_verification,
-    #         self._tls_root_certs,
-    #         self._arrow_endpoint_version,
-    #         self._user_agent
-    #     )
-    #     client.state = self.state
-    #     return client
 
-    
     def _client(self) -> flight.FlightClient:
         """
         Lazy client construction to help pickle this class because a PyArrow
         FlightClient is not serializable.
         """
-        print("checking flight client")
         if not hasattr(self, "_flight_client") or not self._flight_client:
-            print("instantiating flight client")
             self._flight_client = self._instantiate_flight_client()
         return self._flight_client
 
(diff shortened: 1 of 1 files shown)


graphdatascience/query_runner/gds_arrow_client.py
```diff
@@ -89,19 +89,35 @@ def __init__(
...
-            client_options["tls_root_certs"] = tls_root_certs
 
-        self._flight_client = flight.FlightClient(location, **client_options)
+        self._flight_client = self._instantiate_flight_client()
+
+    def _instantiate_flight_client(self) -> flight.FlightClient:
+        location = (
+            flight.Location.for_grpc_tls(self._host, self._port)
+            if self._encrypted
+            else flight.Location.for_grpc_tcp(self._host, self._port)
```
Mats-SX, line 106: ah yes this is the nicer style to do "get or else" that I was looking for

graphdatascience/query_runner/gds_arrow_client.py
```diff
@@ -89,19 +89,35 @@ def __init__(
...
+        )
+        client_options: dict[str, Any] = {"disable_server_verification": self._disable_server_verification}
+        if self._auth:
+            user_agent = f"neo4j-graphdatascience-v{__version__} pyarrow-v{arrow_version}"
+            if self._user_agent:
+                user_agent = self._user_agent
```
Mats-SX, line 112: so it could be
```suggestion
            user_agent = (
                self._user_agent
                if self._user_agent
                else f"neo4j-graphdatascience-v{__version__} pyarrow-v{arrow_version}"
            )
```

but it's not necessary. also maybe this will trip up the typer -- not sure.

graphdatascience/query_runner/gds_arrow_client.py
```diff
@@ -537,11 +553,28 @@ def upload_triplets(
...
+        return state
+
+    def _client(self) -> flight.FlightClient:
+        """
+        Lazy client construction to help pickle this class because a PyArrow
+        FlightClient is not serializable.
+        """
+        if not hasattr(self, "_flight_client") or not self._flight_client:
+            self._flight_client = self._instantiate_flight_client()
+        return self._flight_client
```
Mats-SX, line 570: makes sense

graphdatascience/query_runner/gds_arrow_client.py
```diff
@@ -537,11 +553,28 @@ def upload_triplets(
...
+        state = self.__dict__.copy()
+        # Remove the FlightClient as it isn't serializable
+        if "_flight_client" in state:
+            del state["_flight_client"]
+        return state
+
+    def _client(self) -> flight.FlightClient:
+        """
+        Lazy client construction to help pickle this class because a PyArrow
+        FlightClient is not serializable.
```
Mats-SX, line 566: Makes sense, but I am curious; in which situations do you need to serialise this class?

graphdatascience/query_runner/gds_arrow_client.py
```diff
@@ -537,11 +553,28 @@ def upload_triplets(
...
+        if not hasattr(self, "_flight_client") or not self._flight_client:
+            self._flight_client = self._instantiate_flight_client()
+        return self._flight_client
+
     def _send_action(self, action_type: str, meta_data: dict[str, Any]) -> dict[str, Any]:
         action_type = self._versioned_action_type(action_type)
 
         try:
-            result = self._flight_client.do_action(flight.Action(action_type, json.dumps(meta_data).encode("utf-8")))
+            client = self._client()
```
Mats-SX, line 576: so this either just gets the field, or if this class was restored from a serialised state, then it re-initialises the flight client. Cool.

Mats-SX (approved):
Looks good.

graphdatascience/query_runner/gds_arrow_client.py
Emrehzl94: yeah looks much better this way

graphdatascience/query_runner/gds_arrow_client.py
Emrehzl94: yep

graphdatascience/query_runner/gds_arrow_client.py
Emrehzl94: I'm not sure at this point but we might have needed it this way on BigQuery Connector

Emrehzl94: yes sure, I will let you know why when I make sure about the need on the connector side

Emrehzl94: In the BigQuery connector, we use Spark jobs to process data, and Spark requires a serialized version of this class to distribute the job across different workers. Since FlightClient is not inherently serializable, we needed this lazy initialization.

PR #808, opened 2025-01-02
fix: use serialisation helper function

dhrudevalia:
## [Issue](https://trello.com/c/cUx7Uz5L/805-move-to-gds-client-library-instead-of-neo4jarrow#)
When calling the GDS read functions we are not using the lazy serialisation client. 
This is required for the BigQueryConnector which uses this library instead of neo4j-arrow.
Follow-up from: #804 

## Fix
Use `self._client()` instead of the `_flight_client` directly so that it can be lazy initialised if needed.

### DIFF
--- a/graphdatascience/query_runner/gds_arrow_client.py
+++ b/graphdatascience/query_runner/gds_arrow_client.py
@@ -654,8 +654,9 @@ class GdsArrowClient:
 
         ticket = flight.Ticket(json.dumps(payload).encode("utf-8"))
 
+        client = self._client()
         try:
-            get = self._flight_client.do_get(ticket)
+            get = client.do_get(ticket)
             arrow_table = get.read_all()
         except Exception as e:
             self.handle_flight_error(e)
--- a/graphdatascience/query_runner/gds_arrow_client.py
+++ b/graphdatascience/query_runner/gds_arrow_client.py
@@ -140,7 +140,8 @@ class GdsArrowClient:
             a token from the server and returns it.
         """
         if self._auth:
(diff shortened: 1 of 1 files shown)


graphdatascience/query_runner/gds_arrow_client.py
```diff
@@ -654,8 +654,9 @@ def _do_get(
 
         ticket = flight.Ticket(json.dumps(payload).encode("utf-8"))
 
+        client = self._client()
         try:
-            get = self._flight_client.do_get(ticket)
+            get = client.do_get(ticket)
```
FlorentinD, line 659: makes sense.

FlorentinD (commented):
Checking on other usage, I think we also need to fix some other places:

-  `__exit__` should just call `self.close()
- `close()` .. only close the client if its not None
- https://github.com/neo4j/graph-data-science-client/blob/c774985d5b364b8547cac5f71ae2c909ddffd5b9/graphdatascience/query_runner/gds_arrow_client.py#L143 .. should also use the helper function to fetch the token?

dhrudevalia:
> Checking on other usage, I think we also need to fix some other places:
> 
> * `__exit__` should just call `self.close()
> * `close()` .. only close the client if its not None
> * https://github.com/neo4j/graph-data-science-client/blob/c774985d5b364b8547cac5f71ae2c909ddffd5b9/graphdatascience/query_runner/gds_arrow_client.py#L143
>    .. should also use the helper function to fetch the token?

Yep good shout. I've updated the token fetching + close method as suggested.

FlorentinD (approved):

PR #819, opened 2025-01-19
Fix progress bar not shown without jobId

FlorentinD:
Before we did generate a jobId but never passed it to the algorithm call.
This lead to the client never being able to fetch progress unless the jobId was provided by the user


graphdatascience/algo/single_mode_algo_endpoints.py
```diff
@@ -9,6 +9,7 @@ class SingleModeAlgoEndpoints(CallerBase):
     @property
     @compatible_with("triangles", min_inclusive=ServerVersion(2, 5, 0))
     def triangles(self) -> StreamModeRunner:
+        # todo insert job id
```
Mats-SX, line 12: unresolved todo

graphdatascience/call_parameters.py
```diff
@@ -1,9 +1,40 @@
...
+
+    def get_job_id(self) -> Optional[str]:
+        config = self["config"] if "config" in self else {}
+
+        job_id = None
+        if "jobId" in config:
+            job_id = config["jobId"]
+
+        if "job_id" in config:
+            job_id = config["job_id"]
```
Mats-SX, line 20: duplicated statement
```suggestion
        if "jobId" in config:
            job_id = config["jobId"]
```

Mats-SX: wait ... this is not quite duplicate. It caters for two different syntaxes `jobId` and `job_id`

graphdatascience/tests/unit/test_call_params.py
```diff
@@ -14,3 +16,47 @@ def test_empty_params() -> None:
...
+
+    # job_id given
+    params = CallParameters(config={"job_id": "foo"})
+    job_id = params.get_job_id()
+    assert job_id == "foo"
+
+    # jobId given
+    params = CallParameters(config={"jobId": "bar"})
+    job_id = params.get_job_id()
+    assert job_id == "bar"
```
Mats-SX, line 40: I wonder why we've got both of these

Mats-SX (approved):
I don't quite see where the bug is, but I trust you've fixed it :)

FlorentinD:
> I don't quite see where the bug is, but I trust you've fixed it :)

The problem was the `neo4j_query_runner` created a jobId but this new jobId was never passed to the algorithm config params.
If no jobId was given by the user, we would generate a jobId and try to fetch the progress, but internally in GDS we created a new one as none was passed in the configuration -> never possible to find any progress for the  jobId generated in the neo4j_query_runner

graphdatascience/tests/unit/test_call_params.py
FlorentinD: also wondered why. I could see some usage of `job_id` inside the ArrowQueryRunner, but these are not passed inside a config param.
But for this PR I didnt want to touch as its more easy to introduce another bug if we just remove the case.

Mats-SX: yeah that's good thinking.

PR #821, opened 2025-01-20
Allow str for graph parameter to gds.graph.list

FlorentinD:
- **Allow str for graph filter in gds.graph.list**
- **Cleanup signatures**
- **Fix test and format code**


FlorentinD:
[gds.graph.get(graphName) return Optional[Graph] instead of throwing](https://trello.com/c/1BHpLlAx/10177-gdsgraphgetgraphname-return-optionalgraph-instead-of-throwing)

PR #831, opened 2025-02-01
Pyarrow to pandas less copy

FlorentinD:


graphdatascience/query_runner/gds_arrow_client.py
```diff
@@ -674,7 +676,11 @@ def _do_get(
             message=r"Passing a BlockManager to DataFrame is deprecated",
         )
 
-        return self._sanitize_arrow_table(arrow_table).to_pandas()  # type: ignore
+        if ServerVersion.from_string(pandas.__version__) >= ServerVersion(2, 0, 0):
+            return arrow_table.to_pandas(types_mapper=pandas.ArrowDtype)  # type: ignore
+        else:
+            arrow_table = self._sanitize_arrow_table(arrow_table)
+            return arrow_table.to_pandas()  # type: ignore
```
FlorentinD, line 683: want to run a local check if the address was different before

soerenreichardt: What address did you check?

soerenreichardt (approved):
👍

graphdatascience/query_runner/gds_arrow_client.py
FlorentinD: i used https://fosdem.org/2025/events/attachments/fosdem-2025-6092-what-can-pyarrow-do-for-you-array-interchange-storage-compute-and-transport/slides/236214/Python_tr_DviDDwC.pdf as the base for the check.
Essentially from the node-id column the first value for the numpy-based bandas and the pyarrow column

graphdatascience/query_runner/gds_arrow_client.py
```diff
@@ -674,7 +676,11 @@ def _do_get(
             message=r"Passing a BlockManager to DataFrame is deprecated",
         )
 
-        return self._sanitize_arrow_table(arrow_table).to_pandas()  # type: ignore
+        if ServerVersion.from_string(pandas.__version__) >= ServerVersion(2, 0, 0):
```
DarthMax, line 679: using ServerVersion here is a little confusing. Maybe we should rename the class?

FlorentinD: right we should name it so just SemanticVersion :+1

FlorentinD: #832

PR #855, opened 2025-03-26
Allow creation of standalone sessions

soerenreichardt:


graphdatascience/session/aura_graph_data_science.py
```diff
@@ -66,28 +66,35 @@ def create(
...
+            )
         else:
-            db_bolt_query_runner = Neo4jQueryRunner.create_for_db(
-                db_endpoint.uri,
-                db_endpoint.auth(),
-                aura_ds=True,
-                show_progress=False,
-                database=db_endpoint.database,
+            return cls(
+                query_runner=session_bolt_query_runner,
```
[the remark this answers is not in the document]
FlorentinD: This would lead to `run_cypher` being forwarded to the session wouldnt it? I think we should throw a unsupported operation error or something in such cases

soerenreichardt:
Good points, I implemented a dedicated query runner now that blocks all database related operations

FlorentinD (commented):
LGTM :) 

only would like to have an integration test as well. Maybe in `test_remote_graph_ops.py` but with a graph construct.

+ a changelog entry

soerenreichardt:
> only would like to have an integration test as well 
To test what exactly? :) I don't see what an integration test would give us that the unit test didn't capture in this case.

FlorentinD:
> > only would like to have an integration test as well
> 
> To test what exactly? :) I don't see what an integration test would give us that the unit test didn't capture in this case.

I was thinking of making sure its wired up correctly (i.e. such as the previous error in the query runner).

FlorentinD (approved):

graphdatascience/query_runner/standalone_session_query_runner.py
```diff
@@ -0,0 +1,76 @@
...
+        raise NotImplementedError
+
+    def set_database(self, database: str) -> None:
+        raise NotImplementedError
+
+    def set_bookmarks(self, bookmarks: Optional[Any]) -> None:
+        raise NotImplementedError
+
+    def database(self) -> Optional[str]:
+        raise NotImplementedError
```
FlorentinD, line 67: thats accessed on graph projection .. makes me wonder if there is a bug for sessions in general as this database can be viewed as either the db database or the session database (which doesnt make sense as its fixed)

soerenreichardt: I think in this case there is only one posisbility but for the other query runners this is ambiguous

soerenreichardt: I think this is not a bug in sessions. The code that uses the db information looks like this:
```
        if len(info) == 0:
            raise ValueError(f"There is no projected graph named '{self.name()}'")
        if len(info) > 1:
            # for multiple dbs we can have the same graph name. But db + graph name is unique
            info = info[info["database"] == self._db]
```
So only when we are getting more than one result back from graph.list where we specify a graph name we will check the db. That could in theory be failing as we are using the database name from the db query runner, however on a session we cannot have the same graphName on multiple projections so we should be safe here 🤞. 
We could do better though by maybe having a special `Graph` implementation for sessions?

FlorentinD: Alternatively we can split the `database` method into `gds_database` and `db_database`.

With its usage only on list this should never be an issue for the session case.

For sessions we anyway plan to rewrite quite a bit of the client and maybe we will endup with a different Graph implementation as well 🤔

PR #856, opened 2025-03-26
Raise SessionStatusError on get(_or_create) endpoint

FlorentinD:
Instead of manually handling specific state cases


FlorentinD:
[[🍓] Consolidate 'inactivity' as 'Expired' state, with extra detail](https://trello.com/c/kRpTPOHR/10348-%F0%9F%8D%93-consolidate-inactivity-as-expired-state-with-extra-detail)

soerenreichardt (approved):

PR #859, opened 2025-03-28
Create explicit WCC endpoints

DarthMax:
- **Introduce explict endpoints for Wcc**
- **Implement Wcc endpoints using Cypher procedures**

### DIFF
--- a/graphdatascience/graph_data_science.py
+++ b/graphdatascience/graph_data_science.py
@@ -7,6 +7,9 @@ from typing import Any, Optional, Type, Union
 from neo4j import Driver
 from pandas import DataFrame
 
+from graphdatascience.procedure_surface.api.wcc_endpoints import WccEndpoints
+from graphdatascience.procedure_surface.cypher.wcc_proc_runner import WccCypherEndpoints
+
 from .call_builder import IndirectCallBuilder
 from .endpoints import AlphaEndpoints, BetaEndpoints, DirectEndpoints
 from .error.uncallable_namespace import UncallableNamespace
@@ -106,9 +109,15 @@ class GraphDataScience(DirectEndpoints, UncallableNamespace):
         self._query_runner.set_show_progress(show_progress)
         super().__init__(self._query_runner, namespace="gds", server_version=self._server_version)
 
+        self._wcc_endpoints = WccCypherEndpoints(self._query_runner)
+
     @property
     def graph(self) -> GraphProcRunner:
         return GraphProcRunner(self._query_runner, f"{self._namespace}.graph", self._server_version)
+
+    @property
+    def wcc(self) -> WccEndpoints:
+        return self._wcc_endpoints
 
     @property
     def util(self) -> UtilProcRunner:
--- a/graphdatascience/procedure_surface/api/wcc_endpoints.py
+++ b/graphdatascience/procedure_surface/api/wcc_endpoints.py
@@ -1,0 +1,242 @@
+from abc import ABC, abstractmethod
+from typing import Any, List, Optional
+
+from pandas import DataFrame, Series
+
+from ...graph.graph_object import Graph
+
+
+class WccEndpoints(ABC):
+    """
+    Abstract base class defining the API for the Weakly Connected Components (WCC) algorithm.
+    """
+
+    @abstractmethod
+    def mutate(
+        self,
+        G: Graph,
+        mutate_property: str,
+        threshold: Optional[float] = None,
+        relationship_types: Optional[List[str]] = None,
+        node_labels: Optional[List[str]] = None,
+        sudo: Optional[bool] = None,
+        log_progress: Optional[bool] = None,
+        username: Optional[str] = None,
+        concurrency: Optional[Any] = None,
+        job_id: Optional[Any] = None,
+        seed_property: Optional[str] = None,
+        consecutive_ids: Optional[bool] = None,
+        relationship_weight_property: Optional[str] = None,
+    ) -> Series[Any]:
+        """
+        Executes the WCC algorithm and writes the results to the in-memory graph as node properties.
+
+        Parameters
+        ----------
+        G : Graph
+            The graph to run the algorithm on
+        mutate_property : str
+            The property name to store the component ID for each node
+        threshold : Optional[float], default=None
+            The minimum required weight to consider a relationship during traversal
+        relationship_types : Optional[List[str]], default=None
+            The relationship types to project
+        node_labels : Optional[List[str]], default=None
+            The node labels to project
+        sudo : Optional[bool], default=None
+            Run analysis with admin permission
+        log_progress : Optional[bool], default=None
+            Whether to log progress
+        username : Optional[str], default=None
+            The username to attribute the procedure run to
+        concurrency : Optional[Any], default=None
+            The number of concurrent threads
+        job_id : Optional[Any], default=None
+            An identifier for the job
+        seed_property : Optional[str], default=None
+            Defines node properties that are used as initial component identifiers
+        consecutive_ids : Optional[bool], default=None
+            Flag to decide whether component identifiers are mapped into a consecutive id space
+        relationship_weight_property : Optional[str], default=None
+            The property name that contains weight
+
+        Returns
+        -------
+        Series
+            Algorithm metrics and statistics
+        """
+        pass
+
+    @abstractmethod
+    def stats(
+        self,
+        G: Graph,
+        threshold: Optional[float] = None,
+        relationship_types: Optional[List[str]] = None,
+        node_labels: Optional[List[str]] = None,
+        sudo: Optional[bool] = None,
+        log_progress: Optional[bool] = None,
+        username: Optional[str] = None,
+        concurrency: Optional[Any] = None,
+        job_id: Optional[Any] = None,
+        seed_property: Optional[str] = None,
+        consecutive_ids: Optional[bool] = None,
+        relationship_weight_property: Optional[str] = None,
+    ) -> Series[Any]:
+        """
+        Executes the WCC algorithm and returns statistics.
+
+        Parameters
+        ----------
+        G : Graph
+            The graph to run the algorithm on
+        threshold : Optional[float], default=None
+            The minimum required weight to consider a relationship during traversal
+        relationship_types : Optional[List[str]], default=None
+            The relationship types to project
+        node_labels : Optional[List[str]], default=None
+            The node labels to project
+        sudo : Optional[bool], default=None
+            Run analysis with admin permission
+        log_progress : Optional[bool], default=None
+            Whether to log progress
+        username : Optional[str], default=None
+            The username to attribute the procedure run to
+        concurrency : Optional[Any], default=None
+            The number of concurrent threads
+        job_id : Optional[Any], default=None
+            An identifier for the job
+        seed_property : Optional[str], default=None
+            Defines node properties that are used as initial component identifiers
+        consecutive_ids : Optional[bool], default=None
+            Flag to decide whether component identifiers are mapped into a consecutive id space
+        relationship_weight_property : Optional[str], default=None
+            The property name that contains weight
+
+        Returns
+        -------
+        Series
+            Algorithm metrics and statistics
+        """
+        pass
+
+    @abstractmethod
+    def stream(
+        self,
+        G: Graph,
+        min_component_size: Optional[int] = None,
+        threshold: Optional[float] = None,
+        relationship_types: Optional[List[str]] = None,
+        node_labels: Optional[List[str]] = None,
+        sudo: Optional[bool] = None,
+        log_progress: Optional[bool] = None,
+        username: Optional[str] = None,
+        concurrency: Optional[Any] = None,
+        job_id: Optional[Any] = None,
+        seed_property: Optional[str] = None,
+        consecutive_ids: Optional[bool] = None,
+        relationship_weight_property: Optional[str] = None,
+    ) -> DataFrame:
+        """
+        Executes the WCC algorithm and returns a stream of results.
+
+        Parameters
+        ----------
+        G : Graph
+            The graph to run the algorithm on
+        min_component_size : Optional[int], default=None
+            Don't stream components with fewer nodes than this
+        threshold : Optional[float], default=None
+            The minimum required weight to consider a relationship during traversal
+        relationship_types : Optional[List[str]], default=None
+            The relationship types to project
+        node_labels : Optional[List[str]], default=None
+            The node labels to project
+        sudo : Optional[bool], default=None
+            Run analysis with admin permission
+        log_progress : Optional[bool], default=None
+            Whether to log progress
+        username : Optional[str], default=None
+            The username to attribute the procedure run to
+        concurrency : Optional[Any], default=None
+            The number of concurrent threads
+        job_id : Optional[Any], default=None
+            An identifier for the job
+        seed_property : Optional[str], default=None
+            Defines node properties that are used as initial component identifiers
+        consecutive_ids : Optional[bool], default=None
+            Flag to decide whether component identifiers are mapped into a consecutive id space
+        relationship_weight_property : Optional[str], default=None
+            The property name that contains weight
+
+        Returns
+        -------
+        DataFrame
+            DataFrame with the algorithm results
+        """
+        pass
+
+    @abstractmethod
+    def write(
+        self,
+        G: Graph,
+        write_property: str,
+        min_component_size: Optional[int] = None,
+        threshold: Optional[float] = None,
+        relationship_types: Optional[List[str]] = None,
+        node_labels: Optional[List[str]] = None,
+        sudo: Optional[bool] = None,
+        log_progress: Optional[bool] = None,
+        username: Optional[str] = None,
+        concurrency: Optional[Any] = None,
+        job_id: Optional[Any] = None,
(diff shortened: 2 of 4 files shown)


graphdatascience/procedure_surface/api/wcc_endpoints.py
```diff
@@ -0,0 +1,276 @@
...
+        node_labels: Optional[List[str]] = None,
+        sudo: Optional[bool] = None,
+        log_progress: Optional[bool] = None,
+        username: Optional[str] = None,
+        concurrency: Optional[Any] = None,
+        job_id: Optional[Any] = None,
+        seed_property: Optional[str] = None,
+        consecutive_ids: Optional[bool] = None,
+        relationship_weight_property: Optional[str] = None,
+    ) -> WccMutateResult:
```
DarthMax, line 33: This constitutes a breaking change. Before we would have returned a Dataframe. 
However I think this API is much easier to use as it is more strict on the types and should enabled better IDE auto completion.

graphdatascience/procedure_surface/arrow/wcc_arrow_endpoints.py
```diff
@@ -0,0 +1,236 @@
...
+        mutate_result = MutationClient.mutate_node_property(self._arrow_client, job_id, mutate_property)
+        computation_result = JobClient.get_summary(self._arrow_client, job_id)
+
+        return WccMutateResult(
+            computation_result["componentCount"],
+            computation_result["componentDistribution"],
+            computation_result["preProcessingMillis"],
+            computation_result["computeMillis"],
+            computation_result["postProcessingMillis"],
+            0,
```
DarthMax, line 63: TODO: measure the actual time

graphdatascience/procedure_surface/arrow/wcc_arrow_endpoints.py
```diff
@@ -0,0 +1,236 @@
...
+            threshold,
+        )
+
+        job_id = JobClient.run_job_and_wait(self._arrow_client, WCC_ENDPOINT, config)
+        return JobClient.stream_results(self._arrow_client, job_id)
+
+    def write(
+        self,
+        G: Graph,
+        write_property: str,
```
DarthMax, line 146: TODO: respect that property
This requires a change in the gds.arrow.write procedure in the database

graphdatascience/procedure_surface/arrow/wcc_arrow_endpoints.py
```diff
@@ -0,0 +1,236 @@
...
+            computation_result["preProcessingMillis"],
+            computation_result["computeMillis"],
+            write_millis,
+            computation_result["postProcessingMillis"],
+            computation_result["nodePropertiesWritten"],
+            computation_result["configuration"],
+        )
+
+    @staticmethod
+    def _build_configuration(
```
DarthMax: These methods will become repetitive and error prone. Maybe there is a better, more elegant/pythonic way to do this.

graphdatascience/arrow_client/authenticated_arrow_client.py
```diff
@@ -0,0 +1,211 @@
...
+    Ticket,
+)
+from tenacity import retry, retry_any, retry_if_exception_type, stop_after_attempt, stop_after_delay, wait_exponential
+
+from graphdatascience.arrow_client.arrow_authentication import ArrowAuthentication
+from graphdatascience.arrow_client.arrow_info import ArrowInfo
+from graphdatascience.retry_utils.retry_config import RetryConfig
+
+from ..retry_utils.retry_utils import before_log
+from ..version import __version__
```
FlorentinD, line 25: NIT: mixing import styles

graphdatascience/arrow_client/data_mapper.py
```diff
@@ -0,0 +1,55 @@
...
+
+                # Handle nested dataclasses
+                if dataclasses.is_dataclass(field_type) and isinstance(value, dict):
+                    filtered_data[key] = DataMapper.dict_to_dataclass(value, field_type, strict)  # type:ignore
+                else:
+                    filtered_data[key] = value
+            elif strict:
+                raise ValueError(f"Extra field '{key}' not allowed in {cls.__name__}")
+
+        return cls(**filtered_data)  # type: ignore
```
FlorentinD, line 55: for this validation we could think of also using pydantic instead of rolling our custom validation

graphdatascience/graph_data_science.py
```diff
@@ -117,10 +118,16 @@ def __init__(
         self._query_runner.set_show_progress(show_progress)
         super().__init__(self._query_runner, namespace="gds", server_version=self._server_version)
 
+        self._wcc_endpoints = WccCypherEndpoints(self._query_runner)
```
FlorentinD: we could also only expose them on AuraGraphDataScience for now?

as for normal GDS Plugin the v2 endpoints are not available yet

graphdatascience/procedure_surface/api/wcc_endpoints.py
FlorentinD: i also like the typed api more :) 

especially if we also offer a to_pandas utility (using the from_dict utility).
mainly useful for display in jupyter notebooks

graphdatascience/procedure_surface/api/wcc_endpoints.py
```diff
@@ -0,0 +1,276 @@
+from __future__ import annotations
+
+from abc import ABC, abstractmethod
+from dataclasses import dataclass
+from typing import Any, List, Optional
+
+from pandas import DataFrame
+
+from ...graph.graph_object import Graph
+
```
FlorentinD, line 10: NIT: package name we could rather have `endpoints`

graphdatascience/procedure_surface/api/wcc_endpoints.py
```diff
@@ -0,0 +1,276 @@
...
+@dataclass(frozen=True, repr=True)
+class WccMutateResult:
+    component_count: int
+    component_distribution: dict[str, Any]
+    pre_processing_millis: int
+    compute_millis: int
+    post_processing_millis: int
+    mutate_millis: int
+    node_properties_written: int
+    configuration: dict[str, Any]
```
FlorentinD, line 254: would be neat to generate these :)

graphdatascience/procedure_surface/arrow/wcc_arrow_endpoints.py
FlorentinD: in this internal method you could just accept `**kwargs` and build the config from these as it seems to follow snake case -> camel case?

graphdatascience/procedure_surface/cypher/wcc_proc_runner.py
```diff
@@ -0,0 +1,261 @@
...
+        return WccMutateResult(
+            cypher_result["componentCount"],
+            cypher_result["componentDistribution"],
+            cypher_result["preProcessingMillis"],
+            cypher_result["computeMillis"],
+            cypher_result["postProcessingMillis"],
+            cypher_result["mutateMillis"],
+            cypher_result["nodePropertiesWritten"],
+            cypher_result["configuration"],
+        )
```
FlorentinD: i could also see how we just take the cypher_result and rename the dict keys to snake_case -> `WccMutateResult(**cypher_result)`

FlorentinD (commented):
Great start!

left some comments.
i still tend to think we want to start a feature branch for this to work towards a 2.0 version of the client

graphdatascience/graph_data_science.py
DarthMax: Yeah, but this implementation uses Cypher to retrieve the results

graphdatascience/procedure_surface/cypher/wcc_proc_runner.py
DarthMax: That is a good idea, I added that.

graphdatascience/procedure_surface/arrow/wcc_arrow_endpoints.py
DarthMax: I did add that and pulled it into a utility class

graphdatascience/procedure_surface/arrow/wcc_arrow_endpoints.py
Mats-SX: Is this `mutateMillis`?

Mats-SX: Are we not measuring this already server-side?

graphdatascience/procedure_surface/api/wcc_endpoints.py
Mats-SX: I also prefer the typed API. I don't like breaking things, but I do like having a result type that we own and control, and especially if it comes with static fields to help the user.

graphdatascience/procedure_surface/api/wcc_endpoints.py
```diff
@@ -0,0 +1,276 @@
...
+        relationship_types: Optional[List[str]] = None,
+        node_labels: Optional[List[str]] = None,
+        sudo: Optional[bool] = None,
+        log_progress: Optional[bool] = None,
+        username: Optional[str] = None,
+        concurrency: Optional[Any] = None,
+        job_id: Optional[Any] = None,
+        seed_property: Optional[str] = None,
+        consecutive_ids: Optional[bool] = None,
+        relationship_weight_property: Optional[str] = None,
```
Mats-SX, line 32: This is a very different approach from the design principle we used for the 1.x client. Diametrically different. It is somewhat risky. I think we should hand out an early version of this API to our internal stakeholders in the field to gather feedback.

Personally, I like the explicit and pythonic design. I believe it will be easier to use on its own. It will be somewhat more expensive to maintain and we will have to do more releases in general, but those costs will be worth it if our user base agrees with our perception that this is a better API.

graphdatascience/procedure_surface/api/wcc_endpoints.py
```diff
@@ -0,0 +1,276 @@
...
+            Flag to decide whether component identifiers are mapped into a consecutive id space
+        relationship_weight_property : Optional[str], default=None
+            The property name that contains weight
+
+        Returns
+        -------
+        WccMutateResult
+            Algorithm metrics and statistics
+        """
+        pass
```
Mats-SX, line 71: I suppose this is the "extended documentation burden" that I was highlighting in the ad-hoc discussion. It isn't nothing, but it is co-located in code and requires no additional tooling over the sphinx setup that we already have.

While we will still need to do non-trivial changes to the asciidoc parts of the client manual to align its content with the new API design, I do agree that the majority of change is contained to things like this.

Mats-SX: Of course, the co-location of docs and code does mean that a docbug cannot be fixed without a code release. That can create some annoying situations which also will contribute to an increased release frequency.

For example, if we have the wrong description for a parameter here, we could only correct it with a new release of the client itself.

Mats-SX: Well, our online docs we could regenerate and republish, but the inline code IDE help would require the Python package release.

graphdatascience/procedure_surface/api/wcc_endpoints.py
Mats-SX: Do you mean generating the result class? Do you mean generating them statically before release, or generating them at runtime?

Or do you mean generating configuration classes, and using that instead of the `dict[str, Any]`?

graphdatascience/procedure_surface/api/wcc_endpoints.py
```diff
@@ -0,0 +1,276 @@
...
+@dataclass(frozen=True, repr=True)
+class WccWriteResult:
+    component_count: int
+    component_distribution: dict[str, Any]
+    pre_processing_millis: int
+    compute_millis: int
+    write_millis: int
+    post_processing_millis: int
+    node_properties_written: int
+    configuration: dict[str, Any]
```
Mats-SX: We could consider some static code reuse scheme here, like inheritance, for shared fields.
Or we just have good tests to make sure we don't misspell or forget any field.
We used class inheritance for these classes in Java, but while it gave us type safety it does convolute the code when each level just adds one or two fields and one has to navigate to super classes.
The latest approach of using Java `record`s that implement Java `interface`s to make sure that shared field names are the same and are present is one I like quite a lot. I wonder if we can do a similar thing in Python.

Mats-SX: The general API design here does follow a structured scheme though:

write extends stats, writeCommon
mutate extends stats, mutateCommon

stats extends statsCommon

algoStats extends stats, algoCommon
algoWrite extends algoStats, writeCommon
algoMutate extends algoStats, mutateCommon

But this is mostly a code hygiene effort that can be maintained in multiple ways.

graphdatascience/procedure_surface/arrow/wcc_arrow_endpoints.py
Mats-SX: which is the utility class?

graphdatascience/procedure_surface/cypher/wcc_proc_runner.py
```diff
@@ -0,0 +1,261 @@
...
+            result["computeMillis"],
+            result["writeMillis"],
+            result["postProcessingMillis"],
+            result["nodePropertiesWritten"],
+            result["configuration"],
+        )
+
+    @staticmethod
+    def _create_procedure_config(
+        config: dict[str, Any],
```
Mats-SX: Please, can we return a new value instead of modifying the dict in-place here?

Mats-SX: It seems that we create the dict just before we pass it in to this method. We can do any modifications to the dict after we get it returned from the method instead. That would be my preference. I find methods that modify mutable parameters to be difficult to follow through the code.

graphdatascience/procedure_surface/api/wcc_endpoints.py
FlorentinD: i was thinking statically. checked-in code even and a check that they are up-to-date.
Ideally both but will see if its worth it.

graphdatascience/procedure_surface/api/wcc_endpoints.py
```diff
@@ -0,0 +1,276 @@
...
+
+
+@dataclass(frozen=True, repr=True)
+class WccStatsResult:
+    component_count: int
+    component_distribution: dict[str, Any]
+    pre_processing_millis: int
+    compute_millis: int
+    post_processing_millis: int
+    configuration: dict[str, Any]
```
Mats-SX, line 267: I understand what you mean here with being able to accept camel-cased subscripting to access these fields similar to before.

But I think that we shouldn't do that unless we also allow the camel-cased inputs. And I think we shouldn't do either of those things unless we also make sure to make the whole API non-breaking.

Mats-SX (commented):
In general I like this approach and I believe it will lead to a good product.

I'm a little concerned that we are coupling the two, potentially orthogonal, topics of

- support calling operations on Arrow V2 endpoints
- deliver a more Pythonic API experience

when we could deliver only one of them instead. I don't see why these are logically necessary to do both at the same time.

Perhaps the idea is pragmatic? "The cost of doing `X` on its own is `C`, the cost of doing `Y` on its own is `D`, but the cost of doing `X` and `Y` at the same time is `E < C + D`?

When we couple the two, we make the 'cons' of the either affect both -- for example, users cannot get Arrow V2 without accepting the Python API experience. And the Arrow V2 support is not a feature for the user, just an infrastructure change. Maybe with better stability, but otherwise an 'invisible' change.

graphdatascience/procedure_surface/api/wcc_endpoints.py
Mats-SX: So you mean the result class, and not configuration classes then?

graphdatascience/procedure_surface/config_converter.py
```diff
@@ -0,0 +1,33 @@
+from typing import Any, Dict, Optional
+
+
+class ConfigConverter:
```
Mats-SX, line 4: found it

graphdatascience/procedure_surface/arrow/wcc_arrow_endpoints.py
Mats-SX: nvm found it. I was on not-the-latest changes

graphdatascience/procedure_surface/cypher/wcc_proc_runner.py
Mats-SX: nvm

graphdatascience/procedure_surface/api/wcc_endpoints.py
DarthMax: This code was actually AI generated, so the initial burden is somewhat lessened. It is not perfect and needs to be improved. I wonder if we could move that documentation from our docs into the actual java code that defines the parameters. That way we could pull it in more easily when generating the clients

graphdatascience/procedure_surface/api/wcc_endpoints.py
DarthMax: Since these do not change very often any more, if ever. I think i prefer to keep the separate. We could consider some type of static code analysis to make sure things are consistent.

PR #862, opened 2025-04-07
Use JWT instead of credentials for session authentication

soerenreichardt:


graphdatascience/session/dedicated_sessions.py
```diff
@@ -90,8 +90,8 @@ def get_or_create(
 
         session_connection = DbmsConnectionInfo(
             uri=session_details.bolt_connection_url(),
-            username=self._aura_api._credentials[0],
-            password=self._aura_api._credentials[1],
+            username="",
+            password=self._aura_api._auth._auth_token(),
```
FlorentinD, line 94: what if the token expires? Inside the ArrowQueryRunner we wont be able to refresh the Aura API token?

soerenreichardt: You're right, that needs to be handled. I guess it is a little bit more involved than initially thought :)

graphdatascience/query_runner/gds_arrow_client.py
```diff
@@ -48,7 +49,7 @@ class GdsArrowClient:
     @staticmethod
     def create(
         arrow_info: ArrowInfo,
-        auth: Optional[tuple[str, str]] = None,
+        arrow_authentication: Optional[ArrowAuthentication] = None,
```
FlorentinD, line 52: this is a somewhat breaking change as GdsArrowClient is intended to be used by others as well.

Could we have a union of ArrowAuthentication and tuple -> build typed version inside create if user passed the tuple

FlorentinD: also could keep the same name

PR #868, opened 2025-04-16
Retry idempotent queries using builtin mechanism in neo4j driver

FlorentinD:


FlorentinD:
[Retry executing queries in client](https://trello.com/c/ecAHJ3vB/9102-retry-executing-queries-in-client)

graphdatascience/graph_data_science.py
```diff
@@ -218,6 +218,7 @@ def run_cypher(
         if isinstance(self._query_runner, ArrowQueryRunner):
             qr = self._query_runner.fallback_query_runner()
 
+        # not using qr.execute_query as we dont know if it can be retried
```
FlorentinD: this should be 
```suggestion
        # not using qr.run_retryable_cypher as we dont know if it can be retried
```

graphdatascience/query_runner/protocol/write_protocols.py
```diff
@@ -73,9 +73,10 @@ def run_write_back(
             ProtocolVersion.V1.versioned_procedure_name("gds.arrow.write"),
             parameters,
             yields,
-            None,
-            False,
-            False,
+            retryable=True,
```
FlorentinD: essential for write-protocols

graphdatascience/query_runner/protocol/project_protocols.py
```diff
@@ -97,7 +99,9 @@ def run_projection(
         logging: bool = False,
     ) -> DataFrame:
         versioned_endpoint = ProtocolVersion.V2.versioned_procedure_name(endpoint)
-        return query_runner.call_procedure(versioned_endpoint, params, yields, database, logging, False)
+        return query_runner.call_procedure(
+            versioned_endpoint, params, yields, database=database, logging=logging, retryable=True, custom_error=False
```
FlorentinD: unsure if project can be retried in its current form?

graphdatascience/query_runner/neo4j_query_runner.py
```diff
@@ -218,7 +257,11 @@ def call_procedure(
         query = f"CALL {endpoint}({params.placeholder_str()}){yields_clause}"
 
         def run_cypher_query() -> DataFrame:
-            return self.run_cypher(query, params, database, custom_error)
+            if retryable:
+                routing = neo4j.RoutingControl.WRITE if "write" in endpoint else neo4j.RoutingControl.READ
+                return self.run_retryable_cypher(query, params, database, custom_error, routing=routing)
```
DarthMax, line 262: Will this ever be a safe combination. Our write are not always idempotent.

FlorentinD: the main case was `gds.arrow.write.v3`

Maybe should rather have it explicit in for the write protocol case

FlorentinD: okay its already explicit as before the caller needs to have passed `retryable=True` explicitly

DarthMax: Maybe instead of doing the string matching we could pass in the RoutingControl from the caller

PR #882, opened 2025-05-07
Polish docs continued

FlorentinD:
- **Link to algo api ref docs**
- **Map write-back to session types**
- **Fix wording around using cypher for serverless**


FlorentinD:
[[🍓] Python client documentation review](https://trello.com/c/g8Re5Evl/10582-%F0%9F%8D%93-python-client-documentation-review)

nvitucci (approved):
Looks good.

One suggestion: we usually prefer the extended format for admonitions (e.g. `[NOTE]` plus the `====` separators) rather than the one-line, because it's easier to see and edit, and it can include multiple lines.

FlorentinD:
@nvitucci the admonitions are generated from the markdown in the notebooks. Not sure its easy to change?

nvitucci:
> @nvitucci the admonitions are generated from the markdown in the notebooks. Not sure its easy to change?

Ah, right, I should take a look there then. Not an issue though.

PR #887, opened 2025-05-16
Session remote projection fix

DarthMax:
- **Add QueryRunner.clone method**
- **Make sure to connect to the same instance when querying remote projection status**
- **Add changelog**
- **Apply codestyle**

### DIFF
--- a/graphdatascience/query_runner/neo4j_query_runner.py
+++ b/graphdatascience/query_runner/neo4j_query_runner.py
@@ -295,9 +295,8 @@ class Neo4jQueryRunner(QueryRunner):
             self._auto_close,
             self._bookmarks,
             self._show_progress,
-            self._instance_description
+            self._instance_description,
         )
-
 
     @staticmethod
     def handle_driver_exception(session: neo4j.Session, e: Exception) -> None:
--- a/graphdatascience/query_runner/neo4j_query_runner.py
+++ b/graphdatascience/query_runner/neo4j_query_runner.py
@@ -46,6 +46,7 @@ class Neo4jQueryRunner(QueryRunner):
 
             query_runner = Neo4jQueryRunner(
                 driver,
+                Neo4jQueryRunner.parse_protocol(endpoint),
                 auth,
                 auto_close=True,
                 bookmarks=bookmarks,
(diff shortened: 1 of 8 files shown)


graphdatascience/query_runner/neo4j_query_runner.py
```diff
@@ -141,6 +145,8 @@ def run_cypher(
             connectivity_retry_config = Neo4jQueryRunner.ConnectivityRetriesConfig()
         self._verify_connectivity(database=database, retry_config=connectivity_retry_config)
 
+        self._driver.session()
```
FlorentinD, line 148: what is this doing?

graphdatascience/query_runner/protocol/project_protocols.py
```diff
@@ -132,15 +132,21 @@ def is_not_done(result: DataFrame) -> bool:
 
         logger = getLogger()
 
+        member_address = query_runner.call_procedure(
+            ProtocolVersion.V3.versioned_procedure_name(endpoint), params, yields, database, logging, False
+        ).squeeze()["host"]
+
+        projection_query_runner = query_runner.clone(f"neo4j+s://{member_address}")
```
FlorentinD, line 139: might want to do the same for write-back to make sure we dont get hit by a leader switch and start writing back a second time?

graphdatascience/query_runner/neo4j_query_runner.py
DarthMax: Ah this is some leftover code. Will remove it

graphdatascience/query_runner/protocol/project_protocols.py
DarthMax: The write backs are protected by the procedure being a write procedure. We generally do not survive a leaderswitch right now.

graphdatascience/tests/unit/conftest.py
```diff
@@ -126,6 +126,9 @@ def create_graph_constructor(
             self, graph_name, concurrency, undirected_relationship_types, self._server_version
         )
 
+    def clone(self, host: str, port: int) -> "QueryRunner":
```
FlorentinD, line 129:
```suggestion
    def clone(self, host: str, port: int) -> QueryRunner:
```

graphdatascience/query_runner/standalone_session_query_runner.py
```diff
@@ -74,3 +74,6 @@ def last_bookmarks(self) -> Optional[Any]:
 
     def set_server_version(self, _: ServerVersion) -> None:
         super().set_server_version(_)
+
+    def clone(self, host: str, port: int) -> "QueryRunner":
```
FlorentinD, line 78: NIT: if you use `from __future__ import annotations` you can skip the quotes here

FlorentinD (approved):

PR #891, opened 2025-05-21
Set liveness_check_timeout for AuraDB connection + allow custom config values

FlorentinD:
As the user cannot pass the db driver connection directly, we offer they can pass driver options in get_or_create instead.

### DIFF
--- a/graphdatascience/query_runner/neo4j_query_runner.py
+++ b/graphdatascience/query_runner/neo4j_query_runner.py
@@ -35,9 +35,13 @@ class Neo4jQueryRunner(QueryRunner):
         database: Optional[str] = None,
         bookmarks: Optional[Any] = None,
         show_progress: bool = True,
+        config: Optional[dict[str, Any]] = None,
     ) -> Neo4jQueryRunner:
         if isinstance(endpoint, str):
-            config: dict[str, Any] = {"user_agent": f"neo4j-graphdatascience-v{__version__}"}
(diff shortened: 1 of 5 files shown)


graphdatascience/query_runner/neo4j_query_runner.py
```diff
@@ -100,9 +104,13 @@ def create_for_session(
 
     @staticmethod
     def _configure_aura(config: dict[str, Any]) -> None:
-        config["max_connection_lifetime"] = 60 * 8  # 8 minutes
-        config["keep_alive"] = True
-        config["max_connection_pool_size"] = 50
+        # defaults as documented in https://support.neo4j.com/s/article/1500001173021-How-to-handle-Session-Expired-Errors-while-connecting-to-Neo4j-Aura
+        config.setdefault("max_connection_lifetime", 60 * 50)  # 50 minutes
+        config.setdefault("keep_alive", True)
+        config.setdefault("max_connection_pool_size", 50)
```
soerenreichardt, line 110: Should we also set these defaults when talking to aura db and sessions?

soerenreichardt (approved):

graphdatascience/query_runner/neo4j_query_runner.py
FlorentinD: `_configure_aura` is used for the auradb as well. its set in both right now

PR #893, opened 2025-05-23
Add helper functions to init dataclasses from env variables

FlorentinD:
- **source aura/db credentials from .env file using load_dotenv**
- **Update docs**
- **Add helper functions for reading from env variables**
- **Add note to example about helper function**


FlorentinD:
based on #892

FlorentinD:
[Allow creating python objects from env variables](https://trello.com/c/ggyzVZhR/10662-allow-creating-python-objects-from-env-variables)

doc/modules/ROOT/pages/tutorials/graph-analytics-serverless.adoc
```diff
@@ -111,6 +111,7 @@ from datetime import timedelta
 from graphdatascience.session import DbmsConnectionInfo
 
 # Identify the AuraDB instance
+# you can also use DbmsConnectionInfo.from_env() to load credentials from environment variables
```
DarthMax, line 114: Why not use the tooling here and instead call out the required environment variables in prose or as a listing

DarthMax (approved):

doc/modules/ROOT/pages/tutorials/graph-analytics-serverless.adoc
FlorentinD: we were debating with @RafalSkolasinski .

Its more explicit if you type out the parameters in the example.

I was also considering a mix of both worlds

```
try:
   DbmsConnectionInfo.from_env()
except Exception:
     DbmsConnectionInfo(...)
```

arguably thats a bit harder to read but might just work out of the box if you set devenv. 
and you can also easily fallback if you dont use env.

PR #895, opened 2025-06-05
Add query mode to call_procedure and call_function

FlorentinD:


FlorentinD:
based on feedback from #868

FlorentinD:
moving into #896 as it fixes an import issue as well

PR #900, opened 2025-06-12
Allow advanced auth methods for dbms connection info

FlorentinD:


soerenreichardt (approved):
Look good, but I wonder why you didn't also add this to the `GraphDataScience` constructor?

FlorentinD:
> Look good, but I wonder why you didn't also add this to the `GraphDataScience` constructor?

The ArrowAuthentication assumes UsernamePassword. This wont work if the user only provides a token as auth

PR #911, opened 2025-07-04
Expose mode and retryable on user-facing run_cypher

FlorentinD:
seen in our e2e we now need to manually instantiate a neo4j driver to run retryable cypher queries such as for cleanup.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -6,8 +6,12 @@
 
 ## Bug fixes
 
-* Fix reporting error based on http responses from the Aura-API with an invalid JSON body. Earlier the client would report JSONDecodeError instead of showing the actual issue.
+- Fix reporting error based on http responses from the Aura-API with an invalid JSON body. Earlier the client would report JSONDecodeError instead of showing the actual issue.
 
 ## Improvements
+
+- `GraphDataScience::run_query` now supports setting the `mode` of the query to be used for routing. Previously queries would always route the leader of the cluster, assuming write mode.
+- `GraphDataScience::run_query` now support setting `retryable` to enable a retry-mechanism for appropriate errors. This requires `neo4j>=5.5.0`.
+
 
 ## Other changes
--- a/graphdatascience/graph/graph_entity_ops_runner.py
+++ b/graphdatascience/graph/graph_entity_ops_runner.py
@@ -5,6 +5,8 @@ from warnings import filterwarnings
 import pandas as pd
 from pandas import DataFrame, Series
 
+from graphdatascience.query_runner.query_mode import QueryMode
+
 from ..call_parameters import CallParameters
 from ..error.cypher_warning_handler import (
     filter_id_func_deprecation_warning,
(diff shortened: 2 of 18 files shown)


graphdatascience/graph/graph_entity_ops_runner.py
```diff
@@ -164,7 +166,9 @@ def _process_result(
             unique_node_ids = result["nodeId"].drop_duplicates().tolist()
 
             db_properties_df = query_runner.run_retryable_cypher(
-                GraphNodePropertiesRunner._build_query(db_node_properties), params={"ids": unique_node_ids}
+                GraphNodePropertiesRunner._build_query(db_node_properties),
+                params={"ids": unique_node_ids},
+                mode=QueryMode.READ,
```
Mats-SX, line 171: This is a hard-coded change. We will now run with `READ` for a number of operations, instead of `WRITE`. I think we should call that out in the changelog as well.

graphdatascience/query_runner/neo4j_query_runner.py
```diff
@@ -161,20 +161,28 @@ def run_cypher(
...
         database: Optional[str] = None,
+        mode: Optional[QueryMode] = None,
         custom_error: bool = True,
         connectivity_retry_config: Optional[ConnectivityRetriesConfig] = None,
     ) -> DataFrame:
         if params is None:
             params = {}
 
+        if mode is None:
+            mode = QueryMode.WRITE
```
Mats-SX, line 172: just curious: why do the defaulting here instead of in the parameter declaration?

graphdatascience/query_runner/neo4j_query_runner.py
```diff
@@ -205,18 +213,18 @@ def run_retryable_cypher(
         query: str,
         params: Optional[dict[str, Any]] = None,
         database: Optional[str] = None,
-        custom_error: bool = True,
         mode: Optional[QueryMode] = None,
+        custom_error: bool = True,
```
Mats-SX, line 217: this positional change, is that technically breaking? I'm not too concerned for this method but I think it would break code with positional parameters

changelog.md
```diff
@@ -6,8 +6,12 @@
 
 ## Bug fixes
 
-* Fix reporting error based on http responses from the Aura-API with an invalid JSON body. Earlier the client would report JSONDecodeError instead of showing the actual issue.
+- Fix reporting error based on http responses from the Aura-API with an invalid JSON body. Earlier the client would report JSONDecodeError instead of showing the actual issue.
 
 ## Improvements
 
+- `GraphDataScience::run_query` now supports setting the `mode` of the query to be used for routing. Previously queries would always route the leader of the cluster, assuming write mode.
```
Mats-SX, line 13: - The following endpoints will now run in Cypher `READ` mode instead of `WRITE` mode:
    - `gds.graph.cypher.project()`
    - `gds.graph.nodeProperties.stream()`

Mats-SX (approved):
Looks good, I couldn't find any obvious problems. I think we can improve the changelog.

PR #930, opened 2025-08-04
arrow v2/centrality part 2 

FlorentinD:
- **Use built-in tmp_path**
- **Add BaseResult abstraction**
- **Fix password file dir**
- **Use result abstraction**
- **Add degree endpoints**

### DIFF
--- a/graphdatascience/procedure_surface/api/celf_endpoints.py
+++ b/graphdatascience/procedure_surface/api/celf_endpoints.py
@@ -15,12 +15,7 @@ class CelfEndpoints(ABC):
     Abstract base class defining the API for the Cost Effective Lazy Forward (CELF) algorithm.
 
     The CELF algorithm aims to find k nodes that maximize the expected spread of influence
-    in the network. It's an optimization of the greedy algorithm for influence maximization
-    that uses lazy evaluation to improve performance while maintaining the same approximation
-    guarantees.
-
-    The algorithm works by iteratively selecting nodes that provide the maximum marginal
-    influence gain, using a lazy evaluation strategy to avoid redundant computations.
+    in the network.
     """
 
     @abstractmethod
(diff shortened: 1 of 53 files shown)


graphdatascience/procedure_surface/api/celf_endpoints.py
```diff
@@ -0,0 +1,352 @@
...
+        ----------
+        G : Graph
+            The graph to run the algorithm on
+        seed_set_size : int
+            The number of nodes to select as the seed set for influence maximization
+        mutate_property : str
+            The property name to store the influence spread value for each selected node
+        propagation_probability : Optional[float], default=None
+            The probability that influence spreads from one node to another.
+            If not specified, uses the default value from the algorithm configuration.
```
DarthMax: I wonder if we actually have that kind of sentence everywhere or nowhere

DarthMax: I think nowhere is better. It occurs here multiple times

graphdatascience/procedure_surface/api/celf_endpoints.py
```diff
@@ -0,0 +1,352 @@
...
+        node_labels : Optional[List[str]], default=None
+            The node labels used to select nodes for this algorithm run.
+            If not specified, all node labels are considered.
+        sudo : Optional[bool], default=None
+            Override memory estimation limits. Use with caution as this can lead to
+            memory issues if the estimation is significantly wrong.
+        log_progress : Optional[bool], default=None
+            Whether to log progress of the algorithm execution
+        username : Optional[str], default=None
+            The username to attribute the procedure run to for auditing purposes
```
DarthMax: This is probably wrong everywhere but it makes no sense :see_no_evil:

PR #938, opened 2025-08-18
Add graphsage to v2

FlorentinD:
- **Implement fastrp endpoint**
- **Add hashgnn endpoints**
- **Add node2vec**
- **WIP graphsage support**

### DIFF
--- a/graphdatascience/procedure_surface/api/graphsage_train_endpoints.py
+++ b/graphdatascience/procedure_surface/api/graphsage_train_endpoints.py
@@ -20,6 +20,7 @@ class GraphSageTrainEndpoints(ABC):
         G: Graph,
         model_name: str,
         feature_properties: List[str],
+        *,
         activation_function: Optional[Any] = None,
         negative_sample_weight: Optional[int] = None,
         embedding_dimension: Optional[int] = None,
--- a/graphdatascience/procedure_surface/api/model/graphsage_model.py
+++ b/graphdatascience/procedure_surface/api/model/graphsage_model.py
@@ -32,12 +32,12 @@ class GraphSageModelV2(Model):
         self,
         G: Graph,
         write_property: str,
+        *,
         relationship_types: Optional[list[str]] = None,
         node_labels: Optional[list[str]] = None,
         batch_size: Optional[int] = None,
         concurrency: Optional[int] = None,
         write_concurrency: Optional[int] = None,
-        write_to_result_store: Optional[bool] = None,
         log_progress: Optional[bool] = None,
         username: Optional[str] = None,
         sudo: Optional[bool] = None,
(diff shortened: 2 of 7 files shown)


graphdatascience/procedure_surface/api/model/graphsage_model.py
```diff
@@ -0,0 +1,214 @@
...
+
+        Args:
+            G: The graph to generate embeddings for.
+            relationship_types: The relationship types to consider.
+            node_labels: The node labels to consider.
+            batch_size: The batch size for prediction.
+            concurrency: The concurrency for computation.
+            log_progress: Whether to log progress.
+            username: The username for the operation.
+            sudo: Whether to use sudo privileges.
```
DarthMax: This isn't correct. Should be `Whether to ignore memory estimation`

graphdatascience/procedure_surface/api/graphsage_train_endpoints.py
```diff
@@ -0,0 +1,118 @@
...
+from ...graph.graph_object import Graph
+
+
+class GraphSageTrainEndpoints(ABC):
+    """
+    Abstract base class defining the API for the GraphSage algorithm.
+    """
+
+    @abstractmethod
+    def train(
```
DarthMax, line 18: I wonder if this should be a builder pattern? That is a lot of parameters.

graphdatascience/procedure_surface/arrow/graphsage_predict_arrow_endpoints.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def write(self, G: Graph, **config: Any) -> GraphSageWriteResult:
+        config = self._node_property_endpoints.create_base_config(G, **config)
+
+        raw_result = self._node_property_endpoints.run_job_and_write(
+            "v2/embeddings.graphSage",
+            G,
+            config,
+            config.get("writeConcurrency"),
+            config.get("concurrency"),
```
DarthMax: Can we add those as separate parameters to the method instead of having to use the string

graphdatascience/procedure_surface/arrow/model_api_arrow.py
```diff
@@ -0,0 +1,57 @@
...
+        if not result:
+            return False
+
+        return True
+
+    def get(self, model: str) -> ModelDetails:
+        raw_result = self._arrow_client.do_action_with_retry(
+            "v2/model.get", payload=json.dumps({"modelName": model}).encode("utf-8")
+        )
+        result = deserialize(raw_result)
```
DarthMax, line 32: Can you not use deserialize_single here?

graphdatascience/procedure_surface/api/graphsage_train_endpoints.py
DarthMax: After consulting with the team, we decided that the long list is actually fine. 
I learned that you can use a * in the parameter list, to force all following parameters as keyword arguments.

graphdatascience/procedure_surface/arrow/graphsage_predict_arrow_endpoints.py
DarthMax: I ended up specifying all parameters here as well. This will allow us to offer both APIs for prediction, via the model and via gds.graphSage.predict. I think that is consistent

graphdatascience/procedure_surface/arrow/model_api_arrow.py
DarthMax: Ah it can actually return nothing if the model does not exist. 
Needs a `failIfMissing`

PR #942, opened 2025-08-27
AV2 - graph.sample endpoints

DarthMax:
- **Add GraphSamplingEndpoints**
- **Implement GraphSamplingArrowEndpoints**
- **Use snake case names for sampling arguments**
- **Implement cypher sampling endpoints**
- **Expose sampling endpoints in catalog endpoints**

PR #944, opened 2025-08-28
Graph Sampling Doc improvements

DarthMax:
Based on #942


graphdatascience/procedure_surface/api/centrality/articulationpoints_endpoints.py
```diff
@@ -28,7 +23,10 @@ def mutate(
         job_id: Optional[Any] = None,
     ) -> "ArticulationPointsMutateResult":
         """
-        Executes the ArticulationPoints algorithm and writes the results to the in-memory graph as node properties.
+        Runs the Articulation Points algorithm and stores the results in the graph catalog as a new node property.
+
+        Given a graph, an articulation point is a node whose removal increases the number of connected components in the graph.
+        The Neo4j GDS Library provides an efficient linear time sequential algorithm to compute all articulation points in a graph.
```
FlorentinD: do we want to refer here to the Neo4j GDS Library?

IMO, we can skip the last sentence as well.

FlorentinD (approved):
nice cleanup

one small comment

graphdatascience/procedure_surface/api/centrality/articulationpoints_endpoints.py
Mats-SX: I agree, can take this away.

Mats-SX: Thanks for reading through it

Mats-SX: I was thinking when I did this that I didn't want to deviate from the GDS manual because then there'd be two sources of truth. So I have only copied lines, but the amount of copying is varying. I was thinking that having nice one-liners and one-paragraphers to describe the algorithm is a useful thing also in the manual itself, but I didn't want to scope creep too much. So I think just taking some verbatim content from the manual is my current approach

PR #951, opened 2025-09-15
Start exposing v2 endpoints for GDS

FlorentinD:
wcc and graph endpoints


graphdatascience/session/session_v2_endpoints.py
```diff
@@ -0,0 +1,35 @@
...
+    def wcc(self) -> WccEndpoints:
+        return WccArrowEndpoints(self._arrow_client, self._write_back_client)
+
+    # sessions endpoints: clear if arrow or cypher. for plugin switch between cypher and arrow based on boolean (gds constructor)
+
+    # TODO: endpoints sharing a common suffix such as graph catalog and graph sample options need to have a shared interface to return here
+
+    @property
+    def graph(self) -> CatalogEndpoints:
+        return CatalogArrowEndpoints(self._arrow_client, self._db_client)
```
FlorentinD: AFAIK - project is not included here yet?

graphdatascience/session/session_v2_endpoints.py
```diff
@@ -0,0 +1,35 @@
...
+
+        self._write_back_client = WriteBackClient(arrow_client, db_client) if db_client is not None else None
+
+    @property
+    def wcc(self) -> WccEndpoints:
+        return WccArrowEndpoints(self._arrow_client, self._write_back_client)
+
+    # sessions endpoints: clear if arrow or cypher. for plugin switch between cypher and arrow based on boolean (gds constructor)
+
+    # TODO: endpoints sharing a common suffix such as graph catalog and graph sample options need to have a shared interface to return here
```
FlorentinD: seems the case after all.

I would attempt to do a full e2e part here:

* project
* wcc
* nodes stream/write

(and follow up with KNN which is not implemented yet)

graphdatascience/session/session_v2_endpoints.py
FlorentinD: it differs per implementation. but that will make it impossible to include here.

I think for these SessionEndpoints we can always return the ArrowEndpoints .. but these wouldnt have the doc strings?
makes me lean towards adding the remote project endpoint to the general api class 🤔

FlorentinD: signature overload could be an idea for this case? -- however we dont need the non remote one for the Session case. And for plugin we dont need the remote one.

could have different graph catalog interfaces for these two to return to the user

Mats-SX: Yeah I think returning a different type here with different capabilities does make sense. we could still reuse parts of it I guess

PR #953, opened 2025-09-15
AV2 - Implement gds.graph.relationsips.*

DarthMax:
~Based on #952~
ref GDSA-75


graphdatascience/arrow_client/v2/remote_write_back_client.py
```diff
@@ -27,7 +27,7 @@ def write(
         job_id: str,
         concurrency: Optional[int] = None,
         property_overwrites: Optional[dict[str, str]] = None,
-        relationship_type_overwrites: Optional[dict[str, str]] = None,
+        relationship_type_overwrite: Optional[str] = None,
```
Mats-SX, line 30: so this is similar to what we looked at for node properties. it will be filled with `write_relationship_type` from algo calls, I think

graphdatascience/procedure_surface/api/catalog/relationships_endpoints.py
```diff
@@ -0,0 +1,276 @@
...
+        """
+
+    @abstractmethod
+    def to_undirected(
+        self,
+        G: Graph,
+        relationship_type: str,
+        mutate_relationship_type: str,
+        aggregation: Optional[Union[Aggregation, dict[str, Aggregation]]] = None,
+        *,
```
Mats-SX:
```suggestion
        *,
        aggregation: Optional[Union[Aggregation, dict[str, Aggregation]]] = None,
```
no?

graphdatascience/procedure_surface/api/catalog/relationships_endpoints.py
```diff
@@ -0,0 +1,276 @@
...
+        -------
+        RelationshipsInverseIndexResult
+            Execution metrics and statistics
+        """
+
+    pass
+
+
+@dataclass
+class NodePropertySpec:
```
Mats-SX: What is `Node` doing in this `relationship` code?

Mats-SX: I think this should move, or be deleted

graphdatascience/procedure_surface/api/catalog/relationships_endpoints.py
```diff
@@ -0,0 +1,276 @@
...
+    def to_dict(self) -> dict[str, str]:
+        return self._mappings.copy()
+
+
+class RelationshipsWriteResult(BaseResult):
+    graph_name: str
+    relationship_type: str
+    relationship_properties: list[str] = Field(
+        default=[], validation_alias=AliasChoices("relationshipProperty", "relationshipProperties")
+    )
```
Mats-SX, line 216: what is this Field thing?

graphdatascience/procedure_surface/api/catalog/relationships_endpoints.py
```diff
@@ -0,0 +1,276 @@
...
+    relationships_written: int
+    properties_written: int = 0
```
Mats-SX, line 218: why does this one have a default but the others don't? because all others are returned and this only sometimes?

graphdatascience/procedure_surface/api/catalog/relationships_endpoints.py
```diff
@@ -0,0 +1,276 @@
...
+    write_millis: int
+    configuration: dict[str, Any]
+
+    @field_validator("relationship_properties", mode="before")
+    @classmethod
+    def check_alphanumeric(cls, v: str) -> list[str]:
```
Mats-SX: this seems to wrap strings into lists. not sure the name conveys that

graphdatascience/procedure_surface/api/catalog/relationships_endpoints.py
```diff
@@ -0,0 +1,276 @@
...
+    relationship_type: str
+    relationship_properties: list[str] = Field(
+        default=[], validation_alias=AliasChoices("relationshipProperty", "relationshipProperties")
+    )
+    relationships_written: int
+    properties_written: int = 0
+    write_millis: int
+    configuration: dict[str, Any]
+
+    @field_validator("relationship_properties", mode="before")
```
Mats-SX, line 222: is this related to the `Field` thing above?

graphdatascience/procedure_surface/api/catalog/relationships_endpoints.py
```diff
@@ -0,0 +1,276 @@
...
+    relationships_written: int
+
+
+class Aggregation(str, Enum):
+    NONE = "none"
+    SINGLE = "single"
+    SUM = "sum"
+    MIN = "min"
+    MAX = "max"
+    COUNT = "count"
```
Mats-SX: are we not case sensitive in the proc API? because I've seen these only in ALL_CAPS before

graphdatascience/procedure_surface/arrow/catalog/relationship_arrow_endpoints.py
```diff
@@ -0,0 +1,191 @@
...
+    RelationshipsToUndirectedResult,
+    RelationshipsWriteResult,
+)
+from graphdatascience.procedure_surface.utils.config_converter import ConfigConverter
+
+
+class RelationshipArrowEndpoints(RelationshipsEndpoints):
+    def __init__(self, arrow_client: AuthenticatedArrowClient, query_runner: Optional[QueryRunner] = None):
+        self._arrow_client = arrow_client
+        self._query_runner = query_runner
```
Mats-SX: I think we discussed and concluded that we can remove this field and retain only the `write_back_client`, is that right?

Mats-SX: we don't seem to use the field

Mats-SX: and then we could consider taking `Optional[RemoteWriteBackClient]` in the constructor and having the wrapping happen in the call site. not sure which seems better

graphdatascience/procedure_surface/arrow/catalog/relationship_arrow_endpoints.py
```diff
@@ -0,0 +1,191 @@
...
+            "concurrency": concurrency,
+            "sudo": sudo,
+            "log_progress": log_progress,
+            "username": username,
+        }
+
+        endpoint = "v2/graph.relationships.stream"
+        if relationship_properties and relationship_properties != []:
+            config_input["relationship_properties"] = relationship_properties
+            endpoint = "v2/graph.relationshipProperties.stream"
```
Mats-SX, line 49: this went well IMO

graphdatascience/procedure_surface/arrow/catalog/relationship_arrow_endpoints.py
```diff
@@ -0,0 +1,191 @@
...
+
+        write_result = self._write_back_client.write(
+            G.name(),
+            job_id,
+            concurrency=write_concurrency if write_concurrency is not None else concurrency,
+            relationship_type_overwrite=relationship_type,
+        )
+
+        written_relationships = (
+            write_result.written_relationships if hasattr(write_result, "written_relationships") else 0
```
Mats-SX, line 101: when would it not have `written_relationships`?

also, is it not `relationships_written`?

Mats-SX: no, it is `written_relationships`, because this is remote ops, not plugin

graphdatascience/procedure_surface/arrow/catalog/relationship_arrow_endpoints.py
```diff
@@ -0,0 +1,191 @@
...
+        )
+        written_properties = (
+            written_relationships * len(relationship_properties) if relationship_properties is not None else 0
+        )
```
Mats-SX, line 105: ah we don't return it from `arrow.write` ... while this seems safe enough we need to be aware that it could be wrong since we just multiply some counters, rather than actually counting during writing.

graphdatascience/procedure_surface/arrow/catalog/relationship_arrow_endpoints.py
```diff
@@ -0,0 +1,191 @@
...
+    def drop(
+        self,
+        G: Graph,
+        relationship_type: str,
+        *,
+        fail_if_missing: Optional[bool] = None,
+    ) -> RelationshipsDropResult:
+        # TODO enable once we have a propper graph implementation for arrow endpoints
+        # if not relationship_type in G.relationship_types():
+        #     raise ValueError(f"Relationship type '{relationship_type}' does not exist in graph '{G.name()}'")
```
Mats-SX, line 125: we can do this on this PR if we rebase on top of Florentin's, or we do it after both are merged

graphdatascience/procedure_surface/cypher/catalog/relationship_cypher_endpoints.py
```diff
@@ -0,0 +1,226 @@
...
+                    config=config,
+                )
+            else:
+                endpoint = "gds.graph.relationshipProperties.stream"
+                params = CallParameters(
+                    graph_name=G.name(),
+                    relationship_properties=relationship_properties,
+                    relationship_types=effective_rel_types,
+                    config=config,
+                )
```
Mats-SX, line 82: looks good, well done

graphdatascience/tests/integrationV2/procedure_surface/cypher/catalog/test_node_properties_cypher_endpoints.py
```diff
@@ -79,7 +79,6 @@ def test_stream_node_properties_with_labels(
     assert set(result["prop1"].tolist()) == {1, 2, 3}
 
 
-@pytest.mark.db_integration
```
Mats-SX, line 82: ?

Mats-SX: ah this is supposed to only be used on `write` tests

graphdatascience/tests/integrationV2/procedure_surface/cypher/catalog/test_relationship_cypher_endpoints.py
```diff
@@ -0,0 +1,253 @@
...
+    assert result.graph_name == sample_graph.name()
+    assert result.relationship_type == "REL"
+    assert result.relationship_properties == ["weight", "weight2"]
+    assert result.relationships_written == 3
+    assert result.properties_written == 6
+    assert result.write_millis >= 0
+
+
+def test_drop_relationships(relationship_endpoints: RelationshipCypherEndpoints, sample_graph: Graph) -> None:
+    # Drop REL relationship type
```
Mats-SX, line 187: thank you LLM

graphdatascience/procedure_surface/arrow/catalog/relationship_arrow_endpoints.py
DarthMax: Yes, that information is not present

graphdatascience/procedure_surface/api/catalog/relationships_endpoints.py
DarthMax: No it's not case sensitve. But I can make them so

graphdatascience/procedure_surface/api/catalog/relationships_endpoints.py
DarthMax: Answer for this and the related questions below:
Since we call one of the three write procedures, they have different result fields. The difference is that one has `relationshipProperty`, one has `relationshipProperties` and one has nothing. 
So we need to tell pydantic how to map each of these results to this result class. 
This whole configuration makes it so that it can read the value from either fields or use 0 if none exists.

graphdatascience/procedure_surface/api/catalog/relationships_endpoints.py
DarthMax: Because it is optional. The field does not exist when we call gds.graph.relationship.write

graphdatascience/procedure_surface/api/catalog/relationships_endpoints.py
DarthMax: right, copy past issue

graphdatascience/procedure_surface/api/catalog/relationships_endpoints.py
Mats-SX: That's very clever!

graphdatascience/procedure_surface/api/catalog/relationships_endpoints.py
Mats-SX: then it's OK

PR #954, opened 2025-09-16
WIP introduce Arrow impl for Graph object

FlorentinD:
tradeoffs:

* having a new typed vs the old weakly typed -> went for old one. To allow passing v2 created graphs also to old endpoints
* some open question around snake vs camelCase (due to using weakly typed version)


DarthMax (commented):
This looks generally ok. 
What I do not like is the quasi duplication. The Cypher and Arrow implementations are very similar.
The difference in
* getting the graph info
* dropping
* exists

graphdatascience/procedure_surface/api/catalog/graph_api.py
```diff
@@ -0,0 +1,175 @@
...
+from abc import ABC
+from datetime import datetime
+from types import TracebackType
+from typing import Any, Optional, Type
+
+from graphdatascience.procedure_surface.api.catalog.graph_backend import GraphBackend
+from graphdatascience.procedure_surface.api.catalog.graph_info import GraphInfo
+
+
+class Graph(ABC):
```
FlorentinD: might want to name it GraphV2?

graphdatascience/tests/integrationV2/procedure_surface/arrow/catalog/test_graph_object_arrow.py
```diff
@@ -0,0 +1,125 @@
...
+        (b)-[:REL {z: 7.9}]->(c),
+        (b)-[:REL2 {q: 7.9}]->(d)
+        """
+
+    with create_graph(arrow_client, "g", gdl) as G:
+        yield G
+
+
+# def test_graph_database(G: Graph) -> None:
+#     assert G.database() == "foo"
```
FlorentinD: do we want to offer the database method on the new api?

i think in the past we introduced it to allow switching the db on the query_runner but the graph still is aware on which it was created

graphdatascience/procedure_surface/arrow/catalog/graph_ops_arrow.py
```diff
@@ -0,0 +1,28 @@
+from typing import Optional
+
+from graphdatascience.arrow_client.authenticated_flight_client import AuthenticatedArrowClient
+from graphdatascience.arrow_client.v2.data_mapper_utils import deserialize
+from graphdatascience.procedure_surface.api.catalog.graph_info import GraphInfo
+from graphdatascience.procedure_surface.utils.config_converter import ConfigConverter
+
+
+class GraphOpsArrow:
```
FlorentinD, line 9: introduced to avoid a circular dependency: CatalalogEndpointsArrow -> wrap_graph -> GraphBackendArrow -> CatalalogEndpointsArrow

graphdatascience/procedure_surface/api/catalog/graph_api.py
Mats-SX: either that, or have it in a `v2` package are decent options IMO

Mats-SX: if in a package, we could have a convention to use `v2.Graph` as a name

graphdatascience/procedure_surface/api/catalog/graph_api.py
```diff
@@ -0,0 +1,175 @@
...
+
+    def node_properties(self) -> dict[str, list[str]]:
+        """
+        Returns:
+            the node properties per node label
+
+        """
+        labels_to_props = self._backend.graph_info().graph_schema["nodes"]
+
+        return {key: list(val.keys()) for key, val in labels_to_props.items()}
```
Mats-SX, line 85: is `label_to_props` not a dict like this already?

Mats-SX: oh, no, there is a type part that we erase here.

graphdatascience/procedure_surface/api/catalog/graph_api.py
```diff
@@ -0,0 +1,175 @@
...
+
+        return {key: list(val.keys()) for key, val in types_to_props.items()}
+
+    def degree_distribution(self) -> dict[str, float | int]:
+        """
+        Returns:
+            the degree distribution of the graph
+        """
+        distribution = self._backend.graph_info().degree_distribution
+        return distribution if distribution is not None else {}
```
Mats-SX: when is `distribution` None?

graphdatascience/procedure_surface/api/catalog/graph_api.py
```diff
@@ -0,0 +1,175 @@
...
+            the memory usage of the graph
+        """
+        return self._backend.graph_info().memory_usage
+
+    def size_in_bytes(self) -> int:
+        """
+        Returns:
+            the size of the graph in bytes
+        """
+        return self._backend.graph_info().size_in_bytes
```
Mats-SX, line 122: do we need both of these?

graphdatascience/procedure_surface/api/catalog/graph_api.py
```diff
@@ -0,0 +1,175 @@
...
+
+    def exists(self) -> bool:
+        """
+        Returns:
+            whether the graph exists
+        """
+        return self._backend.exists()
+
+    def drop(self, failIfMissing: bool = False) -> Optional[GraphInfo]:
```
Mats-SX: default=False is a change from before, right? I am not sure which default is better

graphdatascience/procedure_surface/api/catalog/graph_backend.py
```diff
@@ -0,0 +1,24 @@
...
+    def exists(self) -> bool:
+        pass
+
+    @abstractmethod
+    def drop(self, fail_if_missing: bool = True) -> Optional[GraphInfo]:
+        pass
+
+    # @abstractmethod
+    # def database(self) -> str:
+    #     pass
```
Mats-SX: should we remove this comment?

graphdatascience/procedure_surface/api/catalog/graph_info.py
```diff
@@ -0,0 +1,35 @@
...
+
+import neo4j
+from pydantic import BaseModel, Field, field_validator
+from pydantic.alias_generators import to_camel
+
+
+class GraphInfo(BaseModel, alias_generator=to_camel):
+    graph_name: str
+    database: str
+    database_location: str
```
Mats-SX, line 16: I wonder if we can remove these two. In a Session, they are constant (`neo4j` and `remote`).

hm, maybe things like filtered or generated graphs aren't remote, but does that matter? I think we use a different mechanism for whether we allow writeback or not, since generated graphs predate remote projections.

graphdatascience/procedure_surface/api/catalog/graph_info.py
```diff
@@ -0,0 +1,35 @@
...
+    configuration: dict[str, Any]
+    memory_usage: str
+    size_in_bytes: int
+    node_count: int
+    relationship_count: int
+    creation_time: datetime
+    modification_time: datetime
+    graph_schema: dict[str, Any] = Field(alias="schema")
+    schema_with_orientation: dict[str, Any]
```
Mats-SX: at least here we could remove one of the schema variations.

graphdatascience/procedure_surface/api/catalog/graph_info.py
```diff
@@ -0,0 +1,35 @@
...
+    degree_distribution: Optional[dict[str, Union[int, float]]] = None
```
Mats-SX: is this really optional? I thought we always had a degree distribution

graphdatascience/procedure_surface/api/catalog_endpoints.py
```diff
@@ -1,22 +1,19 @@
...
+from graphdatascience.procedure_surface.api.catalog.graph_info import GraphInfo
 from graphdatascience.procedure_surface.api.catalog.node_label_endpoints import NodeLabelEndpoints
 from graphdatascience.procedure_surface.api.graph_sampling_endpoints import GraphSamplingEndpoints
+from graphdatascience.procedure_surface.api.graph_with_result import GraphWithResult
 
 
 class CatalogEndpoints(ABC):
     @abstractmethod
-    def list(self, G: Optional[Union[Graph, str]] = None) -> List[GraphListResult]:
+    def list(self, G: Optional[Union[Graph, str]] = None) -> List[GraphInfo]:
```
Mats-SX: Isn't this a breaking change in client v1?

Mats-SX: ah no, I learned now that `procedure_surface` package is for v2

Mats-SX: maybe that package should be renamed (not this PR)

graphdatascience/procedure_surface/arrow/catalog/graph_backend_arrow.py
```diff
@@ -0,0 +1,33 @@
...
+from graphdatascience.procedure_surface.api.catalog.graph_api import Graph
+from graphdatascience.procedure_surface.api.catalog.graph_backend import GraphBackend
+from graphdatascience.procedure_surface.api.catalog.graph_info import GraphInfo
+from graphdatascience.procedure_surface.arrow.catalog.graph_ops_arrow import GraphOpsArrow
+
+
+def wrap_graph(name: str, arrow_client: AuthenticatedArrowClient) -> Graph:
+    backend = ArrowGraphBackend(name, arrow_client)
+
+    return Graph(name, backend)
```
Mats-SX: we could consider having this as a `@classmethod` on `ArrowGraphBackend`

Mats-SX: although it returns Graph ... hm not sure

graphdatascience/procedure_surface/arrow/catalog/graph_backend_arrow.py
```diff
@@ -0,0 +1,33 @@
...
+
+        if not results:
+            raise ValueError(f"There is no projected graph named '{self._name}'")
+
+        return results[0]
+
+    def exists(self) -> bool:
+        return any(self._catalog_endpoints.list(self._name))
+
+    def drop(self, fail_if_missing: bool = True) -> Optional[GraphInfo]:
```
Mats-SX, line 32: here we default=True. I think we shouldn't have conflicting defaults

Mats-SX: I think I would go with `True` as default

graphdatascience/procedure_surface/arrow/catalog/graph_ops_arrow.py
```diff
@@ -0,0 +1,28 @@
...
+    def __init__(self, arrow_client: AuthenticatedArrowClient):
+        self._arrow_client = arrow_client
+
+    def list(self, graph_name: Optional[str] = None) -> list[GraphInfo]:
+        payload = {"graphName": graph_name} if graph_name else {}
```
Mats-SX, line 14: nice camelcase

graphdatascience/procedure_surface/arrow/catalog/graph_ops_arrow.py
```diff
@@ -0,0 +1,28 @@
...
+
+        result = self._arrow_client.do_action_with_retry("v2/graph.list", payload)
+
+        return [GraphInfo(**x) for x in deserialize(result)]
```
Mats-SX: we could call it what it is
```suggestion
        return [GraphInfo(**dict) for dict in deserialize(result)]
```

graphdatascience/procedure_surface/arrow/catalog/graph_ops_arrow.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def drop(self, graph_name: str, fail_if_missing: Optional[bool] = None) -> Optional[GraphInfo]:
```
Mats-SX, line 20: a third variation for the default
```suggestion
    def drop(self, graph_name: str, fail_if_missing: Optional[bool] = True) -> Optional[GraphInfo]:
```

graphdatascience/procedure_surface/arrow/catalog_arrow_endpoints.py
```diff
@@ -87,18 +91,14 @@ def project(
...
 
-        return ProjectionResult(**JobClient.get_summary(self._arrow_client, job_id))
+        job_result = ProjectionResult(**JobClient.get_summary(self._arrow_client, job_id))
 
-    def drop(self, G: Union[Graph, str], fail_if_missing: Optional[bool] = None) -> Optional[GraphListResult]:
-        graph_name = G if isinstance(G, str) else G.name()
-        config = ConfigConverter.convert_to_gds_config(graph_name=graph_name, fail_if_missing=fail_if_missing)
-        result = self._arrow_client.do_action_with_retry("v2/graph.drop", config)
-        deserialized_results = deserialize(result)
+        return GraphWithResult(wrap_graph(graph_name, self._arrow_client), job_result)
```
Mats-SX: nice

graphdatascience/procedure_surface/arrow/catalog_arrow_endpoints.py
```diff
@@ -87,18 +91,14 @@ def project(
...
 
-        if len(deserialized_results) == 1:
-            return GraphListResult(**deserialized_results[0])
-        else:
-            return None
+    def drop(self, G: Union[Graph, str], fail_if_missing: Optional[bool] = None) -> Optional[GraphInfo]:
```
Mats-SX:
```suggestion
    def drop(self, G: Union[Graph, str], fail_if_missing: Optional[bool] = True) -> Optional[GraphInfo]:
```

Mats-SX: or maybe I am wrong and it should be `False`. But I think having it as `None` makes the reader (me) confused as to what the default actually is

graphdatascience/procedure_surface/cypher/catalog/graph_backend_cypher.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def graph_info(self) -> GraphInfo:
+        info = self._query_runner.call_procedure(
+            endpoint="gds.graph.list",
+            params=CallParameters(graph_name=self._name),
+            custom_error=False,
+        )
+
+        if len(info) == 0:
+            raise ValueError(f"There is no projected graph named '{self._name}'")
```
Mats-SX, line 32: having this fail in the client is new, right? we always had the Bolt Proc fail on this condition, and forwarded that error.
But this is different -- they are just calling e.g. `node_count()` and the graph is suddenly not in the catalog anymore.

We could have some more client errors, that doesn't seem like a bad idea IMO.

graphdatascience/procedure_surface/cypher/catalog/graph_backend_cypher.py
```diff
@@ -0,0 +1,57 @@
...
+        if len(info) > 1:
+            # for multiple dbs we can have the same graph name. But db + graph name is unique
+            info = info[info["database"] == self._db]
```
Mats-SX, line 35: but on sessions this will never happen since the `db` is hard-coded

graphdatascience/procedure_surface/cypher/catalog/graph_backend_cypher.py
```diff
@@ -0,0 +1,57 @@
...
+from graphdatascience.query_runner.query_runner import QueryRunner
+
+
+def wrap_graph(name: str, query_runner: QueryRunner) -> Graph:
+    backend = CypherGraphBackend(name, query_runner)
+
+    return Graph(name, backend)
+
+
+class CypherGraphBackend(GraphBackend):
```
Mats-SX, line 18: I'm thinking about when this class would be used ... is this for plugin? I don't see this class be useful for Sessions -- there's always an Arrow Server and it serves all the catalog information (via `ArrowBackend`).

graphdatascience/procedure_surface/cypher/catalog_cypher_endpoints.py
```diff
@@ -76,7 +79,7 @@ def project(
 
         result = self._query_runner.call_procedure(endpoint="gds.graph.project", params=params).squeeze()
         list_result = GraphProjectResult(**result.to_dict())
-        return Graph(list_result.graph_name, self._query_runner), list_result
+        return wrap_graph(list_result.graph_name, self._query_runner), list_result
```
Mats-SX: so this is where we return the Cypher graph backend. We couldn't return the Arrow backend, because we don't have an Arrow client.

But functionally, we could be using the Arrow server for this use case.

Except no -- because all of the surface here is for plugin. This is the v2 API for plugin. And then we cannot rely on Arrow with V2 endpoints (yet, or ever), but we can rely on a Bolt server because plugin.

graphdatascience/procedure_surface/cypher/catalog/graph_backend_cypher.py
Mats-SX: I think I understood now where this is used. it is for plugin.

graphdatascience/tests/integrationV2/procedure_surface/arrow/catalog/test_graph_object_arrow.py
Mats-SX: I think that if we can avoid it, we should.
And I thought that we could, but I didn't consider that the `Graph` type would be used both for plugin and Sessions. So then perhaps we do need it.

The use case is to make sure we target writeback correctly, I think.

graphdatascience/tests/integrationV2/procedure_surface/arrow/catalog/test_graph_object_arrow.py
```diff
@@ -0,0 +1,125 @@
...
+    assert not G.exists()
+
+    with pytest.raises(ValueError):
+        G.node_count()
+
+    # Should not raise error
+    G.drop(failIfMissing=False)
+
+    with pytest.raises(Exception):
+        G.drop(failIfMissing=True)
```
Mats-SX: perhaps an error message assertion would be good

graphdatascience/tests/integrationV2/procedure_surface/arrow/graph_creation_helper.py
```diff
@@ -1,25 +1,19 @@
...
 from graphdatascience.procedure_surface.arrow.catalog_arrow_endpoints import CatalogArrowEndpoints
 
 
-class TestArrowGraph(Graph):
-    def __init__(self, name: str):
-        self._name = name
-
-    def name(self) -> str:
-        return self._name
-
```
Mats-SX: yay

graphdatascience/tests/integrationV2/procedure_surface/arrow/test_catalog_arrow_endpoints.py
```diff
@@ -77,7 +78,7 @@ def test_drop_nonexistent(catalog_endpoints: CatalogArrowEndpoints) -> None:
 def test_projection(arrow_client: AuthenticatedArrowClient, query_runner: QueryRunner) -> None:
     try:
         endpoints = CatalogArrowEndpoints(arrow_client, query_runner)
-        result = endpoints.project(
+        G, result = endpoints.project(
```
Mats-SX, line 85: I think we should have a couple of assertions, or at least one, on the returned graph object

graphdatascience/tests/integrationV2/procedure_surface/arrow/test_catalog_arrow_endpoints.py
```diff
@@ -94,7 +95,7 @@ def test_projection(arrow_client: AuthenticatedArrowClient, query_runner: QueryR
 
 def test_graph_filter(catalog_endpoints: CatalogArrowEndpoints, sample_graph: Graph) -> None:
     try:
-        result = catalog_endpoints.filter(
+        G, result = catalog_endpoints.filter(
```
Mats-SX, line 103: same here and in some below tests

graphdatascience/tests/integrationV2/procedure_surface/arrow/test_graph_sampling_arrow_endpoints.py
```diff
@@ -67,7 +67,7 @@ def test_rwr_with_weights(graph_sampling_endpoints: GraphSamplingArrowEndpoints,
 
 
 def test_rwr_minimal_config(graph_sampling_endpoints: GraphSamplingArrowEndpoints, sample_graph: Graph) -> None:
-    result = graph_sampling_endpoints.rwr(G=sample_graph, graph_name="sampled")
+    G, result = graph_sampling_endpoints.rwr(G=sample_graph, graph_name="sampled")
```
Mats-SX, line 70: same

graphdatascience/tests/integrationV2/procedure_surface/cypher/test_catalog_cypher_endpoints.py
```diff
@@ -152,7 +154,7 @@ def test_projection(catalog_endpoints: CatalogCypherEndpoints, sample_graph: Gra
 
 
 def test_graph_generate(catalog_endpoints: CatalogCypherEndpoints) -> None:
-    result = catalog_endpoints.generate(
+    G, result = catalog_endpoints.generate(
```
Mats-SX, line 162: assertion on G

graphdatascience/tests/integrationV2/procedure_surface/cypher/test_graph_sampling_cypher_endpoints.py
```diff
@@ -89,7 +90,7 @@ def test_rwr_minimal_config(graph_sampling_endpoints: GraphSamplingCypherEndpoin
 
 def test_cnarw_basic(graph_sampling_endpoints: GraphSamplingCypherEndpoints, sample_graph: Graph) -> None:
     """Test CNARW sampling with basic configuration."""
-    result = graph_sampling_endpoints.cnarw(
+    G, result = graph_sampling_endpoints.cnarw(
```
Mats-SX, line 93: assertion on G?

graphdatascience/tests/unit/procedure_surface/cypher/test_unit_eigenvector_cypher_endpoints.py
```diff
@@ -12,21 +12,11 @@
 from graphdatascience.tests.unit.procedure_surface.cypher.conftests import estimate_mock_result
 
 
-@pytest.fixture
-def query_runner() -> CollectingQueryRunner:
-    return CollectingQueryRunner(DEFAULT_SERVER_VERSION, {})
-
-
```
Mats-SX: nice deduplication

Mats-SX (changes requested):
I think we need to address some details, but overall this seems very solid

graphdatascience/procedure_surface/api/catalog/graph_api.py
DarthMax: That is why I argue that maybe we want the `graph_schema` be a dedicated typed object instead of just a dict[str, Any]

graphdatascience/procedure_surface/api/catalog/graph_api.py
DarthMax: For some reason it is optional in the GraphList Proc output

graphdatascience/procedure_surface/arrow/catalog/graph_backend_arrow.py
```diff
@@ -0,0 +1,33 @@
+from typing import Optional
+
+from graphdatascience.arrow_client.authenticated_flight_client import AuthenticatedArrowClient
+from graphdatascience.procedure_surface.api.catalog.graph_api import Graph
+from graphdatascience.procedure_surface.api.catalog.graph_backend import GraphBackend
+from graphdatascience.procedure_surface.api.catalog.graph_info import GraphInfo
+from graphdatascience.procedure_surface.arrow.catalog.graph_ops_arrow import GraphOpsArrow
+
+
+def wrap_graph(name: str, arrow_client: AuthenticatedArrowClient) -> Graph:
```
DarthMax: I don't really like the method name. 
I think it looks a bit odd to have the wrap_graph call everywhere we create a graph.
Why not call it `create_graph` or `get_graph`. The graph itself is not wrapped.

graphdatascience/tests/integrationV2/procedure_surface/arrow/catalog/test_graph_object_arrow.py
FlorentinD: graph_info also returns the database and database_location. 

so might be fine to fall back to this one. not sure if there is a caveat to it?

graphdatascience/procedure_surface/api/catalog/graph_api.py
FlorentinD: went with GraphV2

graphdatascience/tests/integrationV2/procedure_surface/arrow/catalog/test_graph_object_arrow.py
FlorentinD: the database field was only used in tests right now.

so it looks like we dont check on writeback that its the same db? (unsure)

graphdatascience/procedure_surface/api/catalog/graph_api.py
FlorentinD: will have a typed schema in a follow up

graphdatascience/procedure_surface/api/catalog/graph_api.py
FlorentinD: its the same as on the original graph_object.
our cypher-api -- https://neo4j.com/docs/graph-data-science/current/management-ops/graph-drop/ has a drop if missing to True by default.

so its a bit of a split here

graphdatascience/procedure_surface/api/catalog/graph_api.py
FlorentinD: i think both are useful.
could think of putting them under one endpoint as a tuple as well.

graphdatascience/procedure_surface/api/catalog/graph_info.py
FlorentinD: depends on if we want to reuse this for the plugin as well

graphdatascience/procedure_surface/api/catalog/graph_info.py
FlorentinD: drop doesnt return it. only list does. i will adjust the returns

graphdatascience/procedure_surface/cypher/catalog/graph_backend_cypher.py
FlorentinD: actually existing behaviour of _graph_info on the graph object

graphdatascience/procedure_surface/cypher/catalog/graph_backend_cypher.py
FlorentinD: yes. that actually explains why we had this exposed also on the graph object.
only relevant for cypher based impl for now

graphdatascience/procedure_surface/arrow/catalog/graph_backend_arrow.py
FlorentinD: i like get_graph.

started with create_graph but its not actually creating a graph

graphdatascience/tests/integrationV2/procedure_surface/arrow/catalog/test_graph_object_arrow.py
Mats-SX: in our discussion, we concluded that we are not protecting against this, and that we don't think we need to.

Mats-SX: We think we should retain this method after all, since in the plugin case users may want to know which database (name) their graph was projected from

graphdatascience/procedure_surface/api/catalog/graph_api.py
Mats-SX: ok let's retain them

graphdatascience/procedure_surface/api/catalog/graph_api.py
Mats-SX: yeah hm, I think in this new class we either go with `False` everywhere or `True` everywhere. I am leaning to `True` but perhaps we should do `False` based on the feedback we've got

graphdatascience/procedure_surface/api/catalog/graph_info.py
Mats-SX: I think we do want to use the same GraphV2 for plugin and sessions, so these need to stay

graphdatascience/procedure_surface/api/catalog/graph_info.py
Mats-SX: aha, good find

graphdatascience/procedure_surface/cypher/catalog/graph_backend_cypher.py
Mats-SX: unless it is also useful in the API for plugin users, as a feature

PR #955, opened 2025-09-16
AV2 - Graph Object backends

DarthMax:
* allow different backends
* implement Cypher and Arrow backends

### DIFF
--- a/graphdatascience/graph/cypher_graph_backend.py
+++ b/graphdatascience/graph/cypher_graph_backend.py
@@ -1,0 +1,49 @@
+from __future__ import annotations
+
+from typing import Any
+
+import neo4j
+
+from graphdatascience.call_parameters import CallParameters
+from graphdatascience.graph.graph_backend import GraphBackend, GraphInfo
+
+
+class CypherGraphBackend(GraphBackend):
+    def __init__(self, name: str, query_runner: neo4j.QueryRunner) -> None:
+        self._name = name
+        self._query_runner = query_runner
+        self._db = self._query_runner.database()
+
+    def graph_info(self) -> GraphInfo:
+        info = self._query_runner.call_procedure(
+            endpoint="gds.graph.list",
+            params=CallParameters(graph_name=self._name),
+            custom_error=False,
+        )
+
+        if len(info) == 0:
+            raise ValueError(f"There is no projected graph named '{self._name}'")
+        if len(info) > 1:
+            # for multiple dbs we can have the same graph name. But db + graph name is unique
+            info = info[info["database"] == self._db]
+
+        return GraphInfo(**info.squeeze())
+
+    def exists(self) -> bool:
+        result = self._query_runner.call_procedure(
+            endpoint="gds.graph.exists",
+            params=CallParameters(graph_name=self._name),
+            custom_error=False,
+        )
+        return result.squeeze()["exists"]  # type: ignore
+
+    def drop(self, fail_if_missing: bool = True) -> dict[str, Any]:
+        return (
+            self._query_runner.call_procedure(
+                endpoint="gds.graph.drop",
+                params=CallParameters(graph_name=self._name, failIfMissing=fail_if_missing),
+                custom_error=False,
+            )
+            .squeeze()
+            .to_dict()
+        )  # type: ignore
--- a/graphdatascience/graph/graph_backend.py
+++ b/graphdatascience/graph/graph_backend.py
@@ -1,0 +1,50 @@
+from __future__ import annotations
+
+import re
+from abc import ABC, abstractmethod
+from datetime import datetime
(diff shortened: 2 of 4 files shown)


graphdatascience/graph/cypher_graph_backend.py
```diff
@@ -0,0 +1,49 @@
...
+from graphdatascience.graph.graph_backend import GraphBackend, GraphInfo
+
+
+class CypherGraphBackend(GraphBackend):
+    def __init__(self, name: str, query_runner: neo4j.QueryRunner) -> None:
+        self._name = name
+        self._query_runner = query_runner
+        self._db = self._query_runner.database()
+
+    def graph_info(self) -> GraphInfo:
```
FlorentinD, line 17: we are dropping the yield optimization here, which we had before. but might be worth it.

shouldnt have a big impact

graphdatascience/graph/cypher_graph_backend.py
```diff
@@ -0,0 +1,49 @@
...
+            custom_error=False,
+        )
+
+        if len(info) == 0:
+            raise ValueError(f"There is no projected graph named '{self._name}'")
+        if len(info) > 1:
+            # for multiple dbs we can have the same graph name. But db + graph name is unique
+            info = info[info["database"] == self._db]
+
+        return GraphInfo(**info.squeeze())
```
FlorentinD, line 30: the graph list api on the cypher api was rather stable so this should be fine to return internally as typed

graphdatascience/graph/graph_object.py
```diff
@@ -113,75 +110,70 @@ def node_properties(self, label: Optional[str] = None) -> Union[Series[str], lis
...
-        result = self._query_runner.call_procedure(
-            endpoint="gds.graph.exists",
-            params=CallParameters(graph_name=self._name),
-            custom_error=False,
-        )
-        return result.squeeze()["exists"]  # type: ignore
+        return self._graph_backend.exists()
 
-    def drop(self, failIfMissing: bool = False) -> "Series[str]":
+    def drop(self, failIfMissing: bool = False) -> dict[str, Any]:
```
FlorentinD, line 176: this changes the previous signature and will be a breaking change. i would avoid it

graphdatascience/graph/graph_object.py
```diff
@@ -39,72 +36,72 @@ def name(self) -> str:
...
     def database(self) -> str:
         """
         Returns:
             the name of the database the graph is stored in
         """
-        return self._graph_info(["database"])  # type: ignore
+        return self._graph_backend.graph_info().database
 
-    def configuration(self) -> Series[Any]:
+    def configuration(self) -> dict[str, Any]:
```
FlorentinD, line 68: signature change i would like to do but i think we need to wait for a 2.0 version for this

graphdatascience/graph/graph_object.py
```diff
@@ -39,72 +36,72 @@ def name(self) -> str:
...
     def relationship_types(self) -> list[str]:
         """
         Returns:
             the relationship types in the graph
         """
-        return list(self._graph_info(["schema"])["relationships"].keys())
+        return list(self._graph_backend.graph_info().schema_with_orientation["relationships"].keys())
 
-    def node_properties(self, label: Optional[str] = None) -> Union[Series[str], list[str]]:
+    def node_properties(self, label: Optional[str] = None) -> Union[dict[str, list[str]], list[str]]:
```
FlorentinD, line 104: i would have even liked to always return the same type here, but as above i would wait for a 2.0 version to do this

graphdatascience/graph/graph_object.py
```diff
@@ -220,13 +206,4 @@ def __str__(self) -> str:
...
-            "graphName",
-            "nodeCount",
-            "relationshipCount",
-            "database",
-            "configuration",
-            "schema",
-            "memoryUsage",
-        ]
-        return f"{self.__class__.__name__}({self._graph_info(yields=yield_fields).to_dict()})"
+        return f"{self.__class__.__name__}({self._graph_backend.graph_info()})"
```
FlorentinD, line 209: pydantic allows for a parameter to specify which fields to include

graphdatascience/graph/graph_backend.py
```diff
@@ -0,0 +1,50 @@
...
+    @abstractmethod
+    def graph_info(self) -> GraphInfo:
+        pass
+
+    @abstractmethod
+    def exists(self) -> bool:
+        pass
+
+    @abstractmethod
+    def drop(self, fail_if_missing: bool = True) -> dict[str, Any]:
```
FlorentinD, line 23: why dict here? wouldnt it be also GraphInfo?

DarthMax:
replaced by #954

PR #959, opened 2025-09-18
Introduce aura_instance_id parameter

FlorentinD:
allow to explicitly pass it.
This is important if the original Aura instance id not part of the uri. Such as for multi-db support

ref GDSA-283


changelog.md
```diff
@@ -17,6 +16,6 @@
 
 - `GraphDataScience::run_query` now supports setting the `mode` of the query to be used for routing. Previously queries would always route the leader of the cluster, assuming write mode.
 - `GraphDataScience::run_query` now support setting `retryable` to enable a retry-mechanism for appropriate errors. This requires `neo4j>=5.5.0`.
-
+- `GDsSessions.get_or_create` now allows to specify the `aura_instance_id`. This is required if the instance id could not be derived from the provided database connection URI.
```
FlorentinD: make sure its part of the 1.18 release

FlorentinD:
this override, is only important temporarily while the feature is still in EA.
in the future the instance id will be part of the uri again.

having this as a fallback option still seems like a good idea to me. thinking of the case where the customer has the original uri hidden. could be also a very rare use-case though. especially, would the user know instance_id?

graphdatascience/session/dedicated_sessions.py
```diff
@@ -74,12 +75,17 @@ def get_or_create(
         else:
             db_runner = self._create_db_runner(db_connection, neo4j_driver_options)
 
-            dbid = AuraApi.extract_id(db_connection.uri)
-            aura_db_instance = self._aura_api.list_instance(dbid)
+            aura_instance_id = AuraApi.extract_id(db_connection.uri) if not aura_instance_id else aura_instance_id
+            aura_db_instance = self._aura_api.list_instance(aura_instance_id)
 
             if aura_db_instance is None:
```
FlorentinD: could always have a deprecation warning if the aura_instance_id was not provided explicitly.
helping the user to migrate from the implicit behavior -- being future proof.

src/graphdatascience/query_runner/db_environment_resolver.py
```diff
@@ -0,0 +1,17 @@
...
+            db_runner.run_retryable_cypher("""
+        CALL dbms.components() YIELD name, versions
+        WHERE name = "Neo4j Kernel"
+        UNWIND versions as v
+        WITH name, v
+        WHERE v ENDS WITH "aura"
+        RETURN count(*) <> 0
+        """).squeeze()
+            is True
+        )
```
DarthMax, line 17: The formatting here is very strange

src/graphdatascience/session/dbms_connection_info.py
```diff
@@ -19,6 +19,7 @@ class DbmsConnectionInfo:
     database: str | None = None
     # Optional: typed authentication, used instead of username/password. Supports for example a token. See https://neo4j.com/docs/python-manual/current/connect-advanced/#authentication-methods
     auth: Auth | None = None
+    aura_instance_id: str | None = None
```
DarthMax, line 25: Could it be a either or thing between this and the uri? If the instance id is set we can try to guess the URI?

graphdatascience/session/dedicated_sessions.py
```diff
@@ -74,9 +75,24 @@ def get_or_create(
...
 
+            aura_db_instance = self._aura_api.list_instance(aura_instance_id)
+
+            if not aura_db_instance and DbEnvironmentResolver.hosted_in_aura(db_runner):
+                if db_connection.aura_instance_id:
+                    raise ValueError(
+                        f"Aura instance with id `{db_connection.aura_instance_id}` could not be found. Please verify that the instance id is correct and that you have access to the Aura instance."
+                    )
+                else:
+                    raise ValueError(
```
DarthMax: This is a strange place for this error. I think it would be more clear if we checked if the aura_instance_id variable is None

DarthMax (approved):

src/graphdatascience/session/dbms_connection_info.py
FlorentinD: we can derive the uri from the aura-api endpoint ⭐

doc/modules/ROOT/pages/tutorials/graph-analytics-serverless-spark.adoc
```diff
@@ -0,0 +1,294 @@
+// DO NOT EDIT - AsciiDoc file generated automatically
+
```
FlorentinD, line 2: looks like this was forgotten on main

PR #967, opened 2025-10-06
Do not close not owned neo4j driver

FlorentinD:


graphdatascience/query_runner/neo4j_query_runner.py
```diff
@@ -344,7 +344,8 @@ def set_bookmarks(self, bookmarks: Optional[Any]) -> None:
         self._bookmarks = bookmarks
 
     def close(self) -> None:
-        self._driver.close()
+        if self._auto_close:
+            self._driver.close()
```
Mats-SX, line 348: well found.

Mats-SX: we do this correctly in `__del__`

PR #974, opened 2025-10-07
Allow interruption during wait for job

FlorentinD:


graphdatascience/arrow_client/v2/job_client.py
```diff
@@ -32,28 +34,41 @@ def run_job(client: AuthenticatedArrowClient, endpoint: str, config: dict[str, A
...
-                        progress_bar = TqdmProgressBar(
-                            task_name=base_task,
-                            relative_progress=job_status.progress_percent(),
-                            bar_options=self._progress_bar_options,
-                        )
-                if progress_bar:
-                    progress_bar.update(job_status.status, job_status.progress_percent(), job_status.sub_tasks())
+
+        if termination_flag is None:
+            termination_flag = TerminationFlag.create()
```
Mats-SX, line 47: Do we want to put `TerminationFlag.create()` into the signature as the default value instead?

Mats-SX: no because same object used for all invocations

PR #1091, opened 2026-04-17
Fixes issues in sphinx docs after dark mode added to docs

recrwplay:
Resolves some css conflicts after dark mode was added to docs. Removes almost all colors from this css, relying on defaults from the main site.css. The exception is `class="sphinxsidebar"` which does not exist outside of sphinx docs and has no value already defined.

Where colors are used, they use vards defined in needle design system. These vars are included in the main docs css that the sphinx theme imports.


nvitucci (approved):
LGTM

PR #1097, opened 2026-04-29
Remove model catalog limitations and add section about it (GDSA-1004)

soerenreichardt:


doc/modules/ROOT/pages/aura-graph-analytics.adoc
```diff
@@ -564,10 +564,6 @@ For a full list of the available algorithms, see the https://neo4j.com/docs/grap
    After the Session is deleted, all trained models will be lost.
```
mahdiqb: there's a paragraph here that also needs to be removed.

doc/modules/ROOT/pages/aura-graph-analytics.adoc
```diff
@@ -681,3 +677,7 @@ NOTE: If you want to use Cypher to operate Aura Graph Analytics use the https://
 ----
 gds.run_cypher("MATCH (n:User) RETURN n.name, n.embedding")
 ----
+
+
+== Model Catalog
+Storing models in GDS Sessions is identical to other deployment modes. There are no code or parameter changes necessary.
```
mahdiqb: we can also link to the GDS page with more details on how the model catalog works for Aura Graph Analytics: https://neo4j.com/docs/graph-data-science/current/aura-graph-analytics/machine-learning/

(since we're introducing new concepts like project-level scoping)

mahdiqb (changes requested):

PR #1102, opened 2026-05-08
Expose Store Projection

DarthMax:
Based on #1108


src/graphdatascience/graph/graph_remote_proc_runner.py
```diff
@@ -1,8 +1,55 @@
...
-    @property
-    def project(self) -> GraphProjectRemoteRunner:
-        return GraphProjectRemoteRunner(self._query_runner, self._namespace, self._server_version)
+    def __init__(
+        self, query_runner: QueryRunner, arrow_client: GdsArrowClient, namespace: str, server_version: ServerVersion
+    ):
+        super().__init__(query_runner, namespace, server_version)
+        self._arrow_client = arrow_client
+
+    # The parameters need to be in camelCase to not break the api
```
FlorentinD: ? the parameters are in snake case now

src/graphdatascience/procedure_surface/arrow/catalog/catalog_arrow_endpoints.py
```diff
@@ -108,37 +111,96 @@ def project(
...
-            CatalogArrowEndpoints.GDS_REMOTE_PROJECTION_PROC_NAME,
-            project_params,
-            TerminationFlag.create(),
-            None,
-            None,
-            self._show_progress and logging,
+    def project_native(
+        self,
+        graph_name: str,
+        node_label_filter: builtins.list[str],
```
FlorentinD: NIT - you can also use `List` from typing

src/graphdatascience/procedure_surface/arrow/catalog/catalog_arrow_endpoints.py
```diff
@@ -108,37 +111,96 @@ def project(
...
-
-        self._project_protocol.run_projection(
-            self._query_runner,
-            CatalogArrowEndpoints.GDS_REMOTE_PROJECTION_PROC_NAME,
-            project_params,
-            TerminationFlag.create(),
-            None,
-            None,
-            self._show_progress and logging,
+    def project_native(
```
FlorentinD, line 131: i was also thinking if this shouldnt also be under `project` and inside the impl we call into cypher or native.
but might make it more difficult to call correctly

src/graphdatascience/procedure_surface/arrow/catalog/catalog_arrow_endpoints.py
```diff
@@ -347,12 +409,21 @@ class ProjectionResult(BaseResult):
...
+    project_millis: int
+
+
 class GraphWithProjectResult(NamedTuple):
     """Result object for graph projection jobs, containing the projected graph and the projection result.
     Can be used as a context manager to ensure the projected graph is dropped after use."""
 
     graph: GraphV2
-    result: ProjectionResult
+    result: ProjectionResult | StoreProjectionResult
```
FlorentinD, line 426: could also use inheritance to have the base fields in both cases

src/graphdatascience/procedure_surface/arrow/catalog/catalog_arrow_endpoints.py
```diff
@@ -347,12 +409,21 @@ class ProjectionResult(BaseResult):
     query: str
 
 
+class StoreProjectionResult(BaseResult):
+    """Result object for graph projection jobs."""
```
FlorentinD: thats reads as its for both cypher and native

FlorentinD (approved):

src/graphdatascience/procedure_surface/arrow/catalog/catalog_arrow_endpoints.py
DarthMax: I just used what we use in other places, but I will have a look.

src/graphdatascience/procedure_surface/arrow/catalog/catalog_arrow_endpoints.py
DarthMax: Yeah that was my initial idea, but I did not like mixing the parameters, because then it is not clear what is supported where.

src/graphdatascience/procedure_surface/arrow/catalog/catalog_arrow_endpoints.py
FlorentinD: both works :)

src/graphdatascience/procedure_surface/arrow/catalog/catalog_arrow_endpoints.py
DarthMax: I kinda like doing this via the type system. It is good enough to figure out that there are common methods between the two

src/graphdatascience/graph/graph_remote_proc_runner.py
DarthMax: That comment should have been removed

src/graphdatascience/procedure_surface/arrow/catalog/catalog_arrow_endpoints.py
DarthMax: I have replaced it with `typing.List`

PR #1107, opened 2026-05-12
V4 write endpoints

DarthMax:


tests/integrationV2/procedure_surface/arrow/test_write_protocols.py
```diff
@@ -0,0 +1,100 @@
...
+@pytest.fixture
+def projected_graph(
+    arrow_client: AuthenticatedArrowClient, query_runner: QueryRunner
+) -> Generator[GraphV2, None, None]:
+    graph_data = """
+        CREATE
+        (a {prop: 42}),
+        (b {prop: 43}),
+        (c {prop: 44})
+    """
```
soerenreichardt, line 29: Isn't this the data we insert into the database? Because later we seem to assert on exactly these values, which is always true. We should probably rename the written property values.

soerenreichardt (approved):

tests/integrationV2/procedure_surface/arrow/test_write_protocols.py
DarthMax: You are right, will do in a follow up

PR #1108, opened 2026-05-12
Fix variouse issues with integrationV2 tests

DarthMax:


tests/integrationV2/procedure_surface/conftest.py
```diff
@@ -85,7 +86,7 @@ def start_database(logs_dir: Path, network: Network) -> Generator[DbmsConnection
     neo4j_image = os.getenv("NEO4J_DATABASE_IMAGE", default_neo4j_image)
     if neo4j_image is None:
         raise ValueError("NEO4J_DATABASE_IMAGE environment variable is not set")
-    db_logs_dir = logs_dir / "arrow_surface" / "db_logs"
+    db_logs_dir = logs_dir / str(uuid4()) / "db_logs"
```
FlorentinD: how about we also put the package name in here for clarity? we can use the request fixture that they provide (https://docs.pytest.org/en/latest/reference/reference.html#request)

FlorentinD (approved):

FlorentinD:
there seems to be an issue with the session container setup

PR #1110, opened 2026-05-13
align catalog endpoints with spec

FlorentinD:
- **Fix gds.graph.drop signature**
- **Fix gds.graph.filter.signature**
- **Fix gds.graph.list and gds.graph.generate**
- **Fix sampling and property stream signatures**
- **Fix node label signature**


DarthMax:
Done in another PR, sorry I did not see this one

PR #1112, opened 2026-05-18
fix endpoint signatures for catalog and model endpoints

DarthMax:


soerenreichardt (approved):
👍  there seems to be 1 more failing test though

PR #1116, opened 2026-05-22
Introduce JobHandle

DarthMax:
- **Introduce JobHandle**
- **Introduce WriteJobHandle**
- **Adapt arrow endpoints to use WriteJobHandle**
- **Fix tests**


src/graphdatascience/procedure_surface/api/write_job_handle.py
```diff
@@ -0,0 +1,146 @@
...
+            return self._terminal_status
+        status = self._write_protocol.get_status(self._job_id)
+        if status.done:
+            self._terminal_status = status
+        return status
+
+    def done(self) -> bool:
+        return self.status().done
+
+    def wait(self, log_progress: bool = True) -> None:
```
soerenreichardt, line 72: nit: Rename to `await` to use the default async terminology

tests/integrationV2/procedure_surface/arrow/test_job_handle.py
```diff
@@ -0,0 +1,176 @@
...
+    assert handle.done() is True
+
+
+def test_status_returns_terminal_after_wait(arrow_client: AuthenticatedArrowClient, sample_graph: GraphV2) -> None:
+    handle = _make_handle(arrow_client, sample_graph)
+    handle.wait()
+
+    status = handle.status()
+
+    assert status.succeeded() or status.aborted()
```
soerenreichardt, line 96: Why can both happen?

soerenreichardt (approved):
🎉

src/graphdatascience/procedure_surface/api/write_job_handle.py
DarthMax: Turns out `await` is a keyword and cannot be used.

tests/integrationV2/procedure_surface/arrow/test_job_handle.py
DarthMax: It should not, I will address this in the follow up

PR #1123, opened 2026-06-02
Async Projection methods

DarthMax:
- **Refactor ProjectionProtocol and add ProjectionRunner**
- **Introduce ProjectionJobHandle**
- **Add async projection methods and adapt GraphRemoteProcRunner to ProjectionRunner**
- **Smaller code improvements**

PR #1124, opened 2026-06-02
Return appropriate job handles from gds.jobs.get

DarthMax:
based on #1123

### DIFF
--- a/src/graphdatascience/procedure_surface/arrow/jobs_arrow_endpoints.py
+++ b/src/graphdatascience/procedure_surface/arrow/jobs_arrow_endpoints.py
@@ -59,7 +59,7 @@ class JobsArrowEndpoints:
 
         job = matching_jobs[0]
 
-        if job.job_name == G.name():
+        if self._is_projection(job.job_name, G.name()):
             return ProjectionJobHandle(self._arrow_client, G.name(), job_id, TerminationFlag.create())
         elif self._write_protocol is not None:
             # currently there is no good way of checking if a job is a write job.
@@ -91,3 +91,15 @@ class JobsArrowEndpoints:
 
         rows = deserialize(self._arrow_client.do_action_with_retry(self.LIST_ENDPOINT, {}))
         return [JobInfo(**row) for row in rows]
+
+    def _is_projection(self, job_name: str, graph_name: str) -> bool:
+        projection_endpoints = [
+            "v2/graph.project.fromTables",
+            "v2/graph.project.fromTriplets",
+            "v2/graph.project.filter",
+            "v2/graph.sample.rwr",
+            "v2/graph.sample.cnarw",
+            "v2/graph.generate",
+        ]
+
+        return job_name in projection_endpoints or job_name == graph_name
--- a/src/graphdatascience/procedure_surface/arrow/jobs_arrow_endpoints.py
+++ b/src/graphdatascience/procedure_surface/arrow/jobs_arrow_endpoints.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import time
 import typing
 
 from graphdatascience.arrow_client.authenticated_flight_client import AuthenticatedArrowClient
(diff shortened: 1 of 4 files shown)


src/graphdatascience/procedure_surface/arrow/jobs_arrow_endpoints.py
```diff
@@ -56,13 +59,24 @@ def get(self, G: GraphV2, job_id: str) -> JobHandle | WriteJobHandle:
 
         job = matching_jobs[0]
 
+        if job.job_name == G.name():
+            return ProjectionJobHandle(self._arrow_client, G.name(), job_id, TerminationFlag.create())
+        elif self._write_protocol is not None:
+            # currently there is no good way of checking if a job is a write job.
+            # We can only try to check the job status and fall back to a regular job handle if it fails.
+            try:
+                self._write_protocol.get_status(job.job_id)
```
FlorentinD, line 68: this will work well for protocol > 4 right?

src/graphdatascience/procedure_surface/arrow/jobs_arrow_endpoints.py
```diff
@@ -56,13 +59,24 @@ def get(self, G: GraphV2, job_id: str) -> JobHandle | WriteJobHandle:
 
         job = matching_jobs[0]
 
+        if job.job_name == G.name():
```
FlorentinD: this would break if the user has some graph named the same as an algorithm right?
(unlikely, but imo it would be cleaner to have a job category returned by the jobs.list)

tests/integrationV2/procedure_surface/arrow/test_jobs_arrow_endpoints.py
```diff
@@ -0,0 +1,134 @@
...
+) -> None:
+    # Run a pagerank algorithm so we have a result to write back.
+    algo_job_id = _start_pagerank(arrow_client, db_graph)
+    JobClient().wait_for_job(arrow_client, algo_job_id, show_progress=False)
+
+    write_protocol = WriteProtocol.select(arrow_client, query_runner)
+    write_protocol.start_job(
+        graph_name=db_graph.name(),
+        job_id=algo_job_id,
+        property_overwrites={"pagerank": "pagerank"},
```
FlorentinD, line 125: separate idea: for the property-overwrites to be usable, the results.summary should contain the property names. otherwise the keys here are a bit difficult to specify.

also surprised that page_rank stores a `pagerank` property. would have assumed its called `score`.
likely this write_back wont succeed?

DarthMax: I just realized that the key is actually ignored. I am not quite sure why this is even the API. I will remove the map, so that it is only a string.

FlorentinD: the key is ignored if its only a single property to write-back.
for computation with multiple properties, the keys are mandatory

DarthMax: Right, We do use it in a single algorithm `NodeClassification`. I think I will leave it as is for now. I have no idea what a good API would be. In the end I think the user will simply have to know what to overwrite.

src/graphdatascience/procedure_surface/arrow/jobs_arrow_endpoints.py
DarthMax: Yeah I want to change it on the session side so that the name is the endpoint as well. Maybe then we can even rename it to endpoint instead of name.

src/graphdatascience/procedure_surface/arrow/jobs_arrow_endpoints.py
DarthMax: Yes, that will really only work for V4 which will be 99% of use cases going forward.

tests/integrationV2/procedure_surface/arrow/test_jobs_arrow_endpoints.py
FlorentinD: i think we could offer a way to inspect the shape of the result. maybe even part of the job_summary?
but we can do this in a follow up card

FlorentinD: such as a `results.summary` endpoint

src/graphdatascience/procedure_surface/arrow/jobs_arrow_endpoints.py
DarthMax: I will add a check for the endpoint names of projections as well to make this future proof.

PR #1126, opened 2026-06-03
Refresh flight client on certain failures

FlorentinD:
still seeing FlightUnavailable failures from time to time in our testing. best hunch is to refresh the flight client connection as a whole


DarthMax (approved):
Hard to tell if this really helps but probably worth a try. Do we have any informal metric to check the effectiveness?

FlorentinD:
no its mainly my best idea as we see such failures occur right in the middle of a test after multiple previous action call succeeded.

PR #1129, opened 2026-06-05
Remove v1 endpoints

DarthMax:


tests/integration/test_edge_embedding_model.py
```diff
@@ -1,116 +0,0 @@
-from typing import Generator
```
FlorentinD, line 1: i think edge_embedding endpoints are not even mapped yet in v2.
do we have this tracked?

FlorentinD (approved):
looking forward on seeing the result after the cleanup :)

PR #1130, opened 2026-06-09
Make V2 endpoints default for Plugin

DarthMax:


doc/tests/test_docs.rb
```diff
@@ -11,7 +11,7 @@
 import pandas
 
 from graphdatascience import GraphDataScience
-from graphdatascience.server_version.server_version import ServerVersion
+from graphdatascience.graph.v2.graph_api import ServerVersion
```
FlorentinD, line 14: thats an odd import for ServerVersion

FlorentinD (approved):
Nice simplification.

only wonder about the one import in our docs testing

doc/tests/test_docs.rb
DarthMax: I will have a look at that in a follow up

PR #1131, opened 2026-06-09
Make V2 endpoints default for Sessions

DarthMax:
Based on #1130 
fixes GDSA-1150


src/graphdatascience/error/standalone_session_error.py
```diff
@@ -0,0 +1,3 @@
+class NotAvailableInSessions(Exception):
```
FlorentinD: maybe rather NotAvailableInStandAloneSessions ?
this reads at first as a general unavailability

src/graphdatascience/session/dedicated_sessions.py
```diff
@@ -135,20 +135,18 @@ def get_or_create(
 
         self._await_session_running(session_details, timeout)
 
-        session_bolt_connection_info = DbmsConnectionInfo(
-            uri=session_details.bolt_connection_url(),
-            username=self._aura_api._credentials[0],
-            password=self._aura_api._credentials[1],
-        )
+        session_host = session_details.host.split(":")[0]
+        session_port = 8491  # TODO we should get this from the API
```
FlorentinD: i do wonder about this.

reminds me. we can also move the dedicated_sessions impl into gds_sessions.
this is a remainder from when AuraDS backed sessions were a thing

src/graphdatascience/session/session_lifecycle_manager.py
```diff
@@ -1,7 +1,20 @@
...
 
-class SessionLifecycleManager:
+class LifecycleManager(ABC):
+    @abstractmethod
+    def delete(self) -> bool:
+        pass
+
+    @abstractmethod
+    def verify_health(self) -> None:
+        # Raises a SessionStatusError if the season is in an unhealthy state
```
FlorentinD, line 13: remove this comment from the interface

FlorentinD (approved):

src/graphdatascience/session/dedicated_sessions.py
DarthMax: For now I will hard code as we know it is hard coded in Aura, but I think for the future we might want to change that.

tests/unit/session/test_gds_sessions.py
```diff
@@ -251,6 +256,51 @@ def estimate_size(
         return self._size_estimation
 
 
+class FakeGdsSessions(GdsSessions):
```
FlorentinD, line 259: definitely easier to read 👍

FlorentinD: similar to the collecting query runner

PR #1133, opened 2026-06-10
fix: reshape multi-property relationship stream result in v2 Cypher endpoint

FlorentinD:
The v2 relationship stream endpoint returned long-format results
(relationshipProperty/propertyValue columns) when called via Cypher
(gds without arrow) with multiple relationship properties, instead of
the wide format (one column per property) produced by the Arrow path.
Pivot the Cypher result to match the documented contract.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>


FlorentinD:
@DarthMax there seems to be an issue with the api-ref docs (GraphSage model) at the moment

DarthMax (approved):

PR #1134, opened 2026-06-10
Refactor graph construction

DarthMax:
- **Move graph construction into its own package**
- **Add ArrowV2 graph constructor**
- **Remove graph constructor from query runner**
- **Add graph construction integration tests**


src/graphdatascience/graph_construction/arrow_v1_graph_constructor.py
```diff
@@ -105,14 +108,14 @@ def progress_callback(num_rows: int) -> None:
                     self._client.upload_nodes(
                         self._graph_name,
                         node_data=df,
-                        batch_size=self._min_batch_size,
+                        batch_size=self._min_partition_size,
```
FlorentinD: NIT: batch_size seemed more aligned. why call it partition here?

src/graphdatascience/graph_construction/arrow_v2_graph_constructor.py
```diff
@@ -0,0 +1,106 @@
...
+        batch_size: int = 100_000,
+        show_progress: bool = True,
+    ):
+        self._arrow_client = authenticated_arrow_client
+        self._graph_name = graph_name
+        self._concurrency = concurrency
+        self._client = authenticated_arrow_client
+        self._undirected_relationship_types = (
+            [] if undirected_relationship_types is None else undirected_relationship_types
+        )
```
FlorentinD: here we use the exact is None check? a bit inconsistent with the V1 version

src/graphdatascience/graph_construction/cypher_graph_constructor.py
```diff
@@ -76,33 +78,13 @@ def run(self, node_dfs: list[DataFrame], relationship_dfs: list[DataFrame]) -> N
                 "edition graph construction (slower)"
             )
 
-        # New Cypher propjection supports concurrency since 2.3.0
-        if self._server_version >= ServerVersion(2, 3, 0):
```
FlorentinD, line 80: nice to get rid of this 🚀

tests/integrationV2/graph_construction/test_cypher_graph_constructor.py
```diff
@@ -0,0 +1,172 @@
+from typing import Generator
+
+import pytest
+from pandas import DataFrame
+
+from graphdatascience.graph_construction.cypher_graph_constructor import CypherGraphConstructor
+from graphdatascience.procedure_surface.cypher.catalog.catalog_cypher_endpoints import CatalogCypherEndpoints
+from graphdatascience.query_runner.neo4j_query_runner import Neo4jQueryRunner
+
+
```
FlorentinD, line 12: for these different graph constructors it does feel a bit like we could also have a base test suite or something which gets executed over the different constructors as all should produce the same result

FlorentinD (approved):
good to get rid of the legacy version. 
only minor comments

src/graphdatascience/graph_construction/arrow_v1_graph_constructor.py
DarthMax: Yeah that was incorrect. Batch Size is the size per batch, partition_size is the partition per thread. Fixed it

PR #1135, opened 2026-06-10
GraphV2 to Graph

DarthMax:
Based on #1134 
ref GDSA-1170


src/graphdatascience/graph/__init__.py
```diff
@@ -0,0 +1,7 @@
+from graphdatascience.graph.graph_api import Graph
+from graphdatascience.graph.graph_backend import GraphBackend
+
+__all__ = [
+    "Graph",
+    "GraphBackend",
```
FlorentinD: wonder if we want to expose GraphBackend here.
its an internal class only. could not make it easy to import for users

DarthMax: Right! I removed it

PR #1136, opened 2026-06-10
Tests for nx and ogb loader

DarthMax:
Fixes GDSA-1148

### DIFF
--- a/tests/integrationV2/conftest.py
+++ b/tests/integrationV2/conftest.py
@@ -25,6 +25,12 @@ def pytest_collection_modifyitems(config: Any, items: Any) -> None:
             if "integrationV2" in str(item.fspath):
                 item.add_marker(skip_v2)
 
+    if not config.getoption("--include-ogb"):
+        skip_ogb = pytest.mark.skip(reason="need --include-ogb option to run")
+        for item in items:
+            if "ogb" in item.keywords:
+                item.add_marker(skip_ogb)
+
 
 # best used with pytest --basetemp=tmp/pytest for easy access to logs
 @pytest.fixture(scope="session")
--- a/tests/integrationV2/datasets/conftest.py
+++ b/tests/integrationV2/datasets/conftest.py
@@ -1,0 +1,22 @@
+from pathlib import Path
+from typing import Generator
+
+import pytest
+
+from graphdatascience.arrow_client.authenticated_flight_client import AuthenticatedArrowClient
+from tests.integrationV2.conftest import GdsSessionConnectionInfo, create_arrow_client, start_session
+
+
+@pytest.fixture(scope="package")
+def session_connection(
+    network: object,
+    tmp_path_factory: pytest.TempPathFactory,
+    logs_dir: Path,
+    request: pytest.FixtureRequest,
+) -> Generator[GdsSessionConnectionInfo, None, None]:
+    yield from start_session(logs_dir, tmp_path_factory, network, request)  # type: ignore[arg-type]
+
+
+@pytest.fixture(scope="package")
+def arrow_client(session_connection: GdsSessionConnectionInfo) -> AuthenticatedArrowClient:
+    return create_arrow_client(session_connection)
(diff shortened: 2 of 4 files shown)


tests/integrationV2/datasets/conftest.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+@pytest.fixture(scope="package")
+def session_connection(
+    network: object,
+    tmp_path_factory: pytest.TempPathFactory,
+    logs_dir: Path,
+    request: pytest.FixtureRequest,
+) -> Generator[GdsSessionConnectionInfo, None, None]:
+    yield from start_session(logs_dir, tmp_path_factory, network, request)  # type: ignore[arg-type]
```
FlorentinD, line 18: this is not an issue for mypy is it?
i see pycharm warns about it as its it using ty by default

tests/integrationV2/datasets/conftest.py
```diff
@@ -0,0 +1,22 @@
...
+
+import pytest
+
+from graphdatascience.arrow_client.authenticated_flight_client import AuthenticatedArrowClient
+from tests.integrationV2.conftest import GdsSessionConnectionInfo, create_arrow_client, start_session
+
+
+@pytest.fixture(scope="package")
+def session_connection(
+    network: object,
```
FlorentinD: why type it as object? you can use `from testcontainers.core.network import Network`

tests/integrationV2/datasets/test_nx_loader.py
```diff
@@ -0,0 +1,73 @@
+from typing import Generator
+
+import pytest
+
+pytest.importorskip("networkx")
```
FlorentinD, line 5: importskip is a nice option. something i also read up on recently 👍

tests/integrationV2/datasets/test_nx_loader.py
```diff
@@ -0,0 +1,73 @@
...
+
+import networkx as nx  # noqa: E402
```
FlorentinD: why the `# noqa: E402` on the imports?

tests/integrationV2/conftest.py
```diff
@@ -25,6 +25,12 @@ def pytest_collection_modifyitems(config: Any, items: Any) -> None:
             if "integrationV2" in str(item.fspath):
                 item.add_marker(skip_v2)
 
+    if not config.getoption("--include-ogb"):
+        skip_ogb = pytest.mark.skip(reason="need --include-ogb option to run")
+        for item in items:
+            if "ogb" in item.keywords:
+                item.add_marker(skip_ogb)
```
FlorentinD, line 32: we dont have an option for ogb but none for networkx?

FlorentinD (commented):
found some inconsistencies around the handling of ogb vs networkx.
i was also thinking we could allow to only include the networkx/ogb tests.

in the past we mainly had all + networkx/ogb but this does not make a lot of sense in CI where we have jobs like: default, networkx&ogb.
No need to test the others multiple times

tests/integrationV2/conftest.py
DarthMax: Yeah because OGB needs another dependency and the created graph is fairly large

tests/integrationV2/datasets/test_nx_loader.py
DarthMax: That was already there, this was build from the remnants of an old test

tests/integrationV2/datasets/test_nx_loader.py
DarthMax: it is to suppress warnings about imports not being on the top of the file, but we seem to not check that. Removed it.

tests/integrationV2/datasets/conftest.py
DarthMax: No complaints here.

tests/integrationV2/datasets/test_nx_loader.py
FlorentinD: i also noticed that Claude will add this by default 👍

PR #1137, opened 2026-06-15
Restructure QueryRunner classes etc

DarthMax:


tests/integrationV2/session/dbms/test_db_environment_resolver.py
```diff
@@ -4,11 +4,6 @@
 from graphdatascience.session.dbms.db_environment_resolver import DbEnvironmentResolver
 
 
-@pytest.mark.only_on_aura
-def test_hosted_in_aura_aura_dbms(db_query_runner: Neo4jQueryRunner) -> None:
-    assert DbEnvironmentResolver.hosted_in_aura(db_query_runner)
```
FlorentinD, line 9: you just removed the test-case?

DarthMax: Yeah I did for now. It was the only test left that used this `only_on_aura`.

My goal for now is to run all the integration tests in docker containers without any special configuration. 
We can create an E2E test that actually uses aura but I think that should be  defined separately.

PR #1138, opened 2026-06-15
Remove old integration tests

DarthMax:
Based on #1137 

- **Remove old integration tests and make v2 integration tests standard**
- **update just file**
- **Simplify fixture creation and reduce code duplication in conftest.py**


tox.ini
```diff
@@ -64,13 +43,8 @@ deps =
 	networkx3: networkx >= 3.0, < 4.0
 commands =
 ;	standard: bash -ec 'pytest tests --include-enterprise --include-model-store-location && ruby ./doc/tests/test_docs.rb python3 -n test_enterprise' && pytest ./doc/tests/test_client_only_endpoints.py'
-	standard: bash -ec 'pytest tests --include-enterprise --include-model-store-location && ruby ./doc/tests/test_docs.rb python3 -n test_enterprise'
-	encrypted: pytest tests --encrypted-only
```
FlorentinD, line 68: encrypted is captured through a testcontainer setup?

DarthMax: There was a single test that used the encrypted_only mark on the 1.22 branch and it was a graph construction test. I will add a card to design new tests for encryption.

PR #1142, opened 2026-06-17
Rework docs

nvitucci:


nvitucci:
Note: the examples in the new Getting started page need to be updated to use V2.

FlorentinD:
will merge it and do improvements on follow up PRs

PR #1149, opened 2026-06-22
Support topological link prediction functions in Cypher endpoints

DarthMax:


FlorentinD (commented):
for now its plugin only and in the future they should be part of the Cypher-API in the AGA case i assume?

would be great for the api assertion to have them part of the spec.
will look into it

PR #1152, opened 2026-06-23
Test against latest released by default

FlorentinD:
ref GDSA-1210


FlorentinD:
test_delta_stepping_write failing with unknown flightruntime exception. 
its a flaky test

PR #1172, opened 2026-07-14
AGA projection improvements

DarthMax:
- **Replace gds.graph.project with gds.graph.project.remote in query rewrite**
- **For arrow endpoints introduce project.cypher and project.native**

### DIFF
--- a/doc/modules/ROOT/pages/tutorials/graph-analytics-serverless-self-managed.adoc
+++ b/doc/modules/ROOT/pages/tutorials/graph-analytics-serverless-self-managed.adoc
@@ -11,7 +11,7 @@ This Jupyter notebook is hosted
 https://github.com/neo4j/graph-data-science-client/blob/main/examples/graph-analytics-serverless-self-managed.ipynb[here]
 in the Neo4j Graph Data Science Client Github repository.
 
-The notebook shows how to use the `+graphdatascience+` Python library to
+The notebook shows how to use the `graphdatascience` Python library to
 create, manage, and use a GDS Session.
 
 We consider a graph of people and fruits, which we're using as a simple
@@ -30,8 +30,8 @@ Aura Graph Analytics
 https://neo4j.com/docs/aura/graph-analytics/#aura-gds-serverless[feature]
 is enabled for your Neo4j Aura project.
 
-You also need to have the `+graphdatascience+` Python library installed,
-version `+2.0+` or later.
+You also need to have the `graphdatascience` Python library installed,
+version `2.0` or later.
 
 [source, python, role=no-test]
 ----
@@ -49,7 +49,7 @@ load_dotenv(".env")
 
 == Aura API credentials
 
-The entry point for managing GDS Sessions is the `+GdsSessions+` object,
+The entry point for managing GDS Sessions is the `GdsSessions` object,
 which requires creating
 https://neo4j.com/docs/aura/api/authentication[Aura API credentials].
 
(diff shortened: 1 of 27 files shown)


src/graphdatascience/session/remote_ops/projection_runner.py
```diff
@@ -38,6 +38,17 @@ def run_cypher_projection(
         batch_size: int | None = None,
         show_progress: bool = True,
     ) -> dict[str, Any]:
+        if "gds.graph.project" in query and "gds.graph.project.remote" not in query:
```
FlorentinD, line 41: this is something we might want to change log now.

tests/integration/procedure_surface/arrow/catalog/test_project_arrow_endpoints.py
```diff
@@ -0,0 +1,189 @@
...
+
+@pytest.mark.db_integration
+def test_projection_with_query_rewrite(
+    arrow_client: AuthenticatedArrowClient, query_runner: QueryRunner, endpoints: CatalogArrowEndpoints
+) -> None:
+    try:
+        G, result = endpoints.project.cypher(
+            graph_name="g",
+            query="UNWIND range(1, 10) AS x WITH gds.graph.project(x, null) as g RETURN g",
+        )
```
FlorentinD: shouldnt this produce a warning now which we need to capture?

tests/integration/procedure_surface/arrow/catalog/test_project_arrow_endpoints.py
```diff
@@ -0,0 +1,189 @@
...
+        assert G.name() == "g"
+        assert result.graph_name == "g"
+        assert result.node_count == 10
+        assert result.relationship_count == 0
+        assert result.project_millis >= 0
+
+        assert len(endpoints.list("g")) == 1
+    finally:
+        endpoints.drop("g", fail_if_missing=False)
+
```
FlorentinD, line 82: what about a test to warn on the presence of the undirectedRelTypes in the params?

doc/modules/ROOT/pages/tutorials/graph-analytics-serverless-self-managed.adoc
```diff
@@ -72,12 +72,12 @@ sessions = GdsSessions(api_credentials=api_credentials)
 
 == Creating a new session
 
-A new session is created by calling `sessions.get++_++or++_++create()`
```
FlorentinD, line 75: i think you might have a newer/older version of pandoc installed locally (i have 3.10)

src/graphdatascience/procedure_surface/arrow/catalog/projection_arrow_endpoints.py
```diff
@@ -0,0 +1,298 @@
...
+        """
+        Projects a graph from the Neo4j database into the GDS graph catalog using Cypher projection.
+
+        Parameters
+        ----------
+        graph_name
+            Name of the graph to be created in the catalog.
+        query
+            Cypher query to select nodes and relationships for the graph projection.
+            Must contain `gds.graph.project.remote`. Example: `MATCH (n)-->(m) RETURN gds.graph.project.remote(n, m)`
```
soerenreichardt, line 62: Outdated
```suggestion
```

FlorentinD: wonder if this is still not good to document as a best practice after all? or do you want to update all examples to use `gds.graph.project`?
at least `gds.graph.project` should be part of it

soerenreichardt: Why would the `.remote` option be best practice? I'd actually say the opposite and we should try to not mention that syntax anymore. We could document both options

DarthMax: In the current implementation the `.remote` variant has a different syntax so technically correct would be to use that. I tried in the past to use the same aggregation function for both Cypher API and the Python client but that was unsuccessful. So we will have to keep the two aggregation functions for the time being.

PR #1184, opened 2026-07-16
Align getting started docs with new API

FlorentinD:
also merge AuraDS & plugin sections after initial setup.

ref GDSA-593


doc/modules/ROOT/pages/migration-from-1x.adoc
```diff
@@ -0,0 +1,181 @@
...
+
+
+== Migrating from the `gds.v2` preview prefix
+
+If your code already uses the V2 endpoints via the `gds.v2` prefix, migration only requires dropping the prefix:
+
+[source, python, role=no-test]
+----
+# 1.x preview
+G, result = gds.v2.graph.project.native("g", "*", "*")
```
FlorentinD: actually in the preview it was still `gds.graph.project` the native suffix just came with 2.0 now

doc/modules/ROOT/pages/migration-from-1x.adoc
```diff
@@ -0,0 +1,181 @@
...
+    """
+)
+----
+|===
+
+
+=== Machine learning pipelines and models
+
+Pipelines are now created via the `gds.pipeline` namespace, and the pipeline and model objects expose snake_case methods:
+
```
FlorentinD, line 159: gds.model.get --> use model-specific endpoint such as gds.graph_sage.get()

doc/modules/ROOT/pages/migration-from-1x.adoc
FlorentinD: might remove the projection example for here as there it is AGA vs plugin difference

PR #1195, opened 2026-07-24
Add show_progress config plumbing and FastPath compute endpoint

RafalSkolasinski:
Two small, independent additions to the core library (split out from unrelated CLI work).

## Progress-bar configuration
- `ProgressBar.set_default_options(...)`: process-wide tqdm options merged into every bar, with per-call `bar_options` winning on conflict. Lets a caller set e.g. `{"leave": False}` once for all bars.
- `GdsSessions.get_or_create` / `connect` accept `show_progress` (default `True`).
- `WriteJobHandle` stores a default `log_progress`; `wait()` falls back to it when not given one explicitly.
- Catalog / projection arrow endpoints now default their `logging` to the client's `show_progress` instead of a hardcoded `True`.

## FastPath compute endpoint
- `FastPathArrowEndpoints.compute()` starts the algorithm and returns a `JobHandle`, so the caller decides how to materialize the result (`mutate` / `write` / `stream`). Mirrors `mutate()` but omits the property name.

Unit tests added for both.

### DIFF
--- a/src/graphdatascience/procedure_surface/arrow/catalog/projection_arrow_endpoints.py
+++ b/src/graphdatascience/procedure_surface/arrow/catalog/projection_arrow_endpoints.py
@@ -48,7 +48,7 @@ class ProjectArrowEndpoints:
         undirected_relationship_types: typing.List[str] | None = None,
         inverse_indexed_relationship_types: typing.List[str] | None = None,
         batch_size: int | None = None,
-        logging: bool = True,
+        logging: bool | None = None,
     ) -> GraphWithProjectResult:
         """
         Projects a graph from the Neo4j database into the GDS graph catalog using Cypher projection.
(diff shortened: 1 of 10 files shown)


src/graphdatascience/procedure_surface/arrow/catalog/projection_arrow_endpoints.py
```diff
@@ -72,8 +72,9 @@ def cypher(
             List of relationship types to index in both directions.
         batch_size : int | None, default=None
             Number of rows to process in each batch when projecting the graph.
-        logging : bool, default=True
-            Whether to log progress during graph projection.
+        logging : bool | None, default=None
+            Whether to log progress during graph projection. Defaults to the
+            client's `show_progress` setting.
```
FlorentinD: this is a bit off an indirection to read. 
afterall the default client setting is True. so might as well have True here as a default (like in any other endpoint we have in the client).

i would like to align with other endpoints we have here

FlorentinD (approved):
good fix and interesting see the custom options for the progress bar.

main issue is see is about the log_progress default value in the method signature.
i see you point of None but i think we should align it

src/graphdatascience/procedure_surface/arrow/catalog/projection_arrow_endpoints.py
RafalSkolasinski: sure

PR #1207, opened 2026-07-30
Allow algorithms parameter for session estimations

soerenreichardt:
Part of GDSA-578

PR #1208, opened 2026-07-30
Make sure spec coverage mappings are all used

FlorentinD:
remove unused entries to avoid confusion


FlorentinD:
could be used by #1207 were this is elevated

PR #1210, opened 2026-08-03
gdsa 1424 fix expired token issue on gdsdelete

FlorentinD:
- **Improve handling of expired AuraAPI tokens**
- **Improve Aura API request handling**

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -14,8 +14,8 @@
 
 ## Bug fixes
 
-* Requests to the Aura API that are rejected as unauthorized are now retried once with a newly minted OAuth token, instead of failing permanently while the expired token stays cached. This affected operations such as `GdsSessions.delete`, which could leave a session running.
-* Requests to the Aura API now use a connect and read timeout, so that a stalled connection can no longer block a call indefinitely. The number of retries was also reduced, so that a request is less likely to be retried for longer than its OAuth token stays valid.
+* Requests to the Aura API that are rejected as unauthorized are now retried once with a new OAuth token, instead of failing permanently while the expired token stays cached
+* Requests to the Aura API now use a connect and read timeout, so that a stalled connection can no longer block a call indefinitely.
 * `GdsSessions.delete` now returns `False` if no session was deleted when called with a `session_id`. It previously always returned `True`.
 * `AuraApiError` and `SessionStatusError` no longer include a repetition of the exception object in their message.
 
--- a/changelog.md
+++ b/changelog.md
@@ -14,6 +14,8 @@
 
 ## Bug fixes
 
+* Requests to the Aura API that are rejected as unauthorized are now retried once with a newly minted OAuth token, instead of failing permanently while the expired token stays cached. This affected operations such as `GdsSessions.delete`, which could leave a session running.
+
 
 ## Improvements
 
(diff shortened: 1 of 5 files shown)


tests/unit/test_aura_api.py
```diff
@@ -32,6 +33,33 @@ def mock_auth_token(requests_mock: Mocker) -> None:
...
 
 
+def test_error_message_is_not_wrapped() -> None:
+    error = AuraApiError("Request for https://api.neo4j.io/ failed with status code 401", 401)
+
+    assert str(error) == "Request for https://api.neo4j.io/ failed with status code 401"
+    assert error.args == ("Request for https://api.neo4j.io/ failed with status code 401",)
+
+
+def test_session_status_error_message_is_not_wrapped() -> None:
```
soerenreichardt: I don't understand the `is_not_wrapped` part here. Your assertion clearly expects a string that is not part of the original message. So it is wrapped?

tests/unit/test_aura_api.py
```diff
@@ -32,6 +33,33 @@ def mock_auth_token(requests_mock: Mocker) -> None:
...
+    api = AuraApi(client_id="", client_secret="", project_id="some-tenant")
+    mock_auth_token(requests_mock)
+    requests_mock.delete("https://api.neo4j.io/v1/graph-analytics/sessions/id0", status_code=202)
+
+    api.delete_session("id0")
+
+    # Every request needs a bound, so that a stalled connection cannot block indefinitely.
+    # This covers the oauth token request as well as the api call itself.
+    timeouts: list[Any] = [r.timeout for r in requests_mock.request_history]
+    assert len(timeouts) == 2
```
soerenreichardt: Is it 2 because of the retry? That would be worth a comment

changelog.md
```diff
@@ -14,6 +14,11 @@
 
 ## Bug fixes
 
+* Requests to the Aura API that are rejected as unauthorized are now retried once with a newly minted OAuth token, instead of failing permanently while the expired token stays cached. This affected operations such as `GdsSessions.delete`, which could leave a session running.
```
soerenreichardt:
```suggestion
* Requests to the Aura API that are rejected as unauthorized are now retried once with a new OAuth token, instead of failing permanently while the expired token stays cached.
```
Let's remove some of those common LLM words :D Also we usually don't specify the failure scenarios in the bug fix announcements

soerenreichardt (approved):
Looks good but I have a few smaller comments

tests/unit/test_aura_api.py
FlorentinD: we passed a reference of itself before in the AuraApiError case.
the test is not very helpful though just reading at it.

tests/unit/test_aura_api.py
FlorentinD: clarified. its token + get request

src/graphdatascience/session/aura_api.py
```diff
@@ -416,6 +421,38 @@ def _check_endpoint_deprecation(self, resp: requests.Response) -> None:
...
+            response = super().request(method, url, *args, **kwargs)
+
+            if response.status_code != HTTPStatus.UNAUTHORIZED.value:
+                return response
+
+            self._logger.debug("Request was unauthorized, retrying it with a new oauth token")
+            # A 401 means the request was rejected before being acted on, so resending it is
+            # safe even for the methods we otherwise refuse to retry.
+            self._auth.invalidate_token()
+            response.close()
```
DarthMax, line 452: nit: I guess we should close this immediately.

DarthMax (approved):

PR #1217, opened 2026-08-06
Apply dependabot dependency updates

FlorentinD:
Batches the applicable open dependabot bumps into one commit so that
`pyproject.toml` and `uv.lock` stay consistent:

- ruff 0.15.20 -> 0.16.1 (#1203 proposed 0.16.0; 0.16.1 is the latest and adds only bug fixes and preview-rule changes on top of it)
- types-setuptools 83.0.0.20260706 -> 83.0.0.20260724 (#1202)
- types-python-dateutil 2.9.0.20260518 -> 2.9.0.20260716 (#1201)
- types-tqdm 4.68.0.20260608 -> 4.70.0.20260805 (#1199)
- neo4j-viz 1.6.0 -> 1.7.0 (#1200, lock only; the `>=1.6.0` floor stays put so the constraint remains permissive for library consumers)
- aiohttp 3.14.1 -> 3.14.3 (#1214)
- cryptography 49.0.0 -> 50.0.0 (#1216)

Also re-resolves the whole lockfile with `uv lock --upgrade` rather than only the
packages dependabot filed PRs for, which moves 48 packages in total. The ones worth
noting beyond the list above: **pyarrow 24.0.0 -> 25.0.0**, tqdm 4.68.4 -> 4.70.0 and
multimethod 2.0.2 -> 2.1 are runtime dependencies; pandas goes 3.0.3 -> 3.0.5; twine
6.2.0 -> 7.0.0 affects the `publish` group only; and sphinx-autodoc-typehints 2.3.0 ->
3.0.1 plus alabaster 0.7.16 -> 1.0.0 are major bumps but transitive only, since neither
is in the sphinx extension list and the theme is the vendored `neo4j` one.

### Why mypy.ini changes

pyarrow 25 withdraws the `py.typed` marker and inline annotations that 24 shipped
([apache/arrow#50168](https://github.com/apache/arrow/pull/50168)), because those
annotations were incomplete and broke type checkers on submodules such as
`pyarrow.compute` ([apache/arrow#49831](https://github.com/apache/arrow/issues/49831)).
Without `py.typed`, strict mypy reports 26 `import-untyped` errors across the arrow
client. Restoring `ignore_missing_imports` for pyarrow reverts exactly what #1164
removed when it adopted pyarrow 24, so this is the pre-24 behaviour rather than a new
relaxation. `pyarrow-stubs` is not a usable alternative: it targets the pyarrow 20 API
and yields 35 errors in 7 files. The stanza can go once upstream ships complete
annotations ([apache/arrow#32609](https://github.com/apache/arrow/issues/32609)).

### Verified

- ruff 0.16.1 reformats nothing and reports no new findings. The 0.16 default-rule
  expansion (59 -> 413 rules) does not reach us because `tool.ruff.lint.select` is
  explicit, and the new markdown code-block formatting is excluded by `include`.
- mypy clean; 632 unit tests pass.
- The sphinx API docs still build with `-W`, and the rendered HTML is **byte-identical**
  to a pre-upgrade build, so the sphinx transitive majors change no output.
- nbconvert still produces byte-identical asciidoc for the tutorial notebooks despite
  the mistune 3.3.2 -> 3.3.4 bump.

### Not included

The torch 2.12.1 -> 2.13.0 bump (#1186, #1192) is still left out, for the same reason as
in #1196 and re-checked today: the PyG wheel index for `torch-2.13.0+cpu` ships only
`pyg_lib`, no `torch-scatter` or `torch-sparse` wheels, so `notebook-ci` would fall back
to building those from source — which `pyproject.toml` documents as failing on macOS
arm64.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


neo4j-docops-agent:
Thanks for the API reference updates.

The preview API reference has now been torn down - reopening this PR will republish it.

PR #1190, opened 2026-07-21
DOCOPS-168 Add standard docs workflows (doc/ subfolder)

recrwplay:
## What

Adds the standard Neo4j docs workflow set to this repo (docs live in the `doc/` subfolder):

- **`docs-pr-checks.yml`** – build + verify docs on PRs, and post a preview-changes comment.
- **`docs-deploy-surge.yml`** – deploy the PR preview to surge (with the hardened artifact validation from docs-template: validate-changelog, `process.env.RUN_ID`, zipinfo path checks).
- **`docs-teardown.yml`** – tear the preview down when the PR closes.
- **`docs-generate-html.yml`** – build + publish on push to a published branch.

## Adapted for this repo

- `docs-dir: 'doc'` and path filters on `doc/**` (this repo uses `doc/`, not `docs/`).
- Trigger branches follow `doc/publish.yml`:
  - pr-checks / teardown: `main` + published versions `1.19`, `1.20`, `1.21`, `1.22`.
  - generate-html (publish): the published versions `1.19`–`1.22` (`main`/2.0 is not published yet).

## Needs maintainer review

- **generate-html publish config:** the standard single `DEV_BRANCH`/`PROD_BRANCH` model doesn't map cleanly onto this repo's 4-version publish (`doc/publish.yml`). Confirm the staging/prod branch scheme before relying on the publish step (marked with a NOTE in the file).
- These workflow files are added on `main` only. A `pull_request`/`push` workflow only runs when the file is present on the **base branch**, so to get docs CI/publish on the `1.19`–`1.22` branches the files would also need adding there — deliberately **not** done here, pending the developers' decision on which branches should carry them.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1028, opened 2025-12-10
Docs workflows

recrwplay:
Adds docs GHA workflows to make sure docs always build cleanly before being merged.

- _docs-pr-checks.yml_ - runs when a PR contains changes in _doc/**_
- _docs-deploy-surge.yml_ - runs when _docs-pr-checks.yml_ finishes. Deploys the generated HTML to a surge.sh URL for preview
- _docs-teardown.yml_ - removes the deployed preview HTML when a PR is closed.

The docs _package.json_ file has been updated to use the latest versions of docs packages.

The playbooks (_preview.yml_ and _publish.yml_ have been updated with the latest extensions).

This PR also fixes an issue with the formatting of footnotes in the _pipelines.adoc_ file.

PR #1219, opened 2026-08-06
fix docs build issues

recrwplay:
Note: This PR supercedes #1028, which was intended to resolve the same issues, but also added some docs project changes that are now redundant.

Fixes warnings in the docs build for the 1.22 branch.

Warnings come mostly from some footnotes. Additionally there is the `Conclusion` anchor link that was resolved in `main` as part of #1190. The fix here was to remove the link rather than resolving it, however).


nvitucci (approved):

neo4j-docops-agent:
Thanks for the API reference updates.

The preview API reference has now been torn down - reopening this PR will republish it.

Issue #254, opened 2023-01-18
Add more examples


FlorentinD:
**Is your feature request related to a problem? Please describe.**

In our [examples](https://github.com/neo4j/graph-data-science-client/tree/main/examples) we have several jupyter notebooks to show how to use GDS.
However, there are several areas left uncovered as we mostly focused on the ML parts of our library.
Adding a new example would help to learn the best practices of the GDS client as well as other users.

Potential areas to cover (see https://neo4j.com/docs/graph-data-science/current/algorithms/ for all of our algorithms):

- [ ] Path algorithms such as Steiner trees
- [ ] Similarity algorithms
- [ ] Community detection algorithms
- [ ] Centrality algorithms
- [ ] ...

You can also come up with your own idea of course.

Please drop a comment if you like to work on an example :)

kedarghule:
Hey, I'm a complete novice at Neo4j but I have been working on an example [here](https://github.com/kedarghule/Community-Detection-using-Neo4j/blob/main/community-detection.ipynb) for community detection. Let me know your thoughts :)

kedarghule:
Now working on a centrality algorithms notebook. Will open a PR when it's done. It's on the repo you shared with me @FlorentinD :)

adamnsch:
> Now working on a centrality algorithms notebook. Will open a PR when it's done. It's on the repo you shared with me @FlorentinD :)

very cool! :tada:

TreavVasu:
Hi @FlorentinD  
Hope there's still requirement for this notebook :)

PR #1225, opened 2026-08-09
Added Similarity example notebook

TreavVasu:
## Graph Similarity Algorithms in Neo4j GDS

Hi Team,

With ref to #254, This notebook demonstrates how to calculate and compare graph similarity metrics using the  [SNAP Act-MOOC dataset](https://snap.stanford.edu/data/act-mooc.html) and the Neo4j Graph Data Science (GDS) library.

The dataset is:

* Directed
* Temporal
* Attributed

## Tasks Covered

* Data ingestion into Neo4j Aura
* In-memory graph projections
* Structural node similarity using:
  * Jaccard similarity
  * Overlap similarity
* Feature-based similarity using the K-Nearest Neighbors (KNN) algorithm
* Algorithm execution in Stream, Stats, and Mutate modes
* Comparison of Jaccard, Overlap, and Cosine similarity metrics

## Learning Objectives
* **Data Ingestion:** Load large-scale interaction data into Neo4j Aura.
* **Graph Projections:** Create in-memory graph projections for analysis.
* **Similarity Algorithms:** Execute Node Similarity algorithms in Stream, Stats, and Mutate modes.
* **Metric Comparison:** Understand the differences between Jaccard, Overlap, and Cosine similarity.

## Contribution Checklist
* [x] You signed the [[Neo4j Contributor License Agreement (CLA)](https://neo4j.com/developer/cla/#sign-cla)](https://neo4j.com/developer/cla/#sign-cla), allowing Neo4j to ship your code with the library.
* [x] Your contribution is covered by tests.


TreavVasu:
Hi @FlorentinD  
Lmk if this example books are needeed, I can work on others as well if this would be approved (other left overs).

FlorentinD:
@TreavVasu thanks for picking this up!
i will take a look on Friday, I think its definitely useful to have these examples. 

we are right now working on a 2.0 release which brings a couple of breaking changes. i would likely adjust this notebook to the new 2.0 syntax (main branch).

TreavVasu:
Sure 
Let me know if there's an updated documentation on this section I can ref to

brs96 (changes requested):
Thanks for the contribution, generally it looks good.
A few things to clean up before merging:
1. In the opening paragraph we can probably describe in one or two sentences what the MOOC dataset is about and what the node similarity analysis tells us. I know full summary is at the bottom already.
2. Under `### Run Cypher Query on GDS Session` and `### Database Stats: Labels and Relationships`, I see some Cypher queries inspecting the DB. However, the actual CSV->Neo4j data loading happens later in `# Step 1: Tutorial Setup & Constants`? I believe we want to have the data loading first and then the EDA steps. Also, `Constants`->`Constraints?
3. Section 1 has 3 steps, perhaps name the steps as Step 1.1, 1.2, 1.3 to be clearer?
4. We should clear the outputs of all the cells similar to the other notebooks.

TreavVasu:
Hi @brs96 

Thanks for the lookout 

I didn't notice the other notebooks didn't include output lol (will clean that up) 


Thanks for other points as well, I'll check on those and update it

FlorentinD:
@TreavVasu for the output cleaning there is a script you can use in the repository (scripts/makestyle).

for the 2.0 api, you can take a look at https://neo4j.com/docs/graph-data-science-client/2.0-preview/migration-from-1x/

TreavVasu:
Sure 
I'll check on these thanks 🤜🤛

TreavVasu:
Hi @FlorentinD  

Thanks for the tip on that cleaning script made the job easier. I'm checking if I can migrate to v2 in this but getting few installation and sync issues with AGA Version 

[Colab ref](https://colab.research.google.com/drive/15sVOR5qNsaPaU9C-BRm-omB_LQRFEpXd?usp=sharing) 

Hopefully this if fine for 1.X 

Thanks @brs96  for the reviews, lmk if there are any additional things I need to take care of

FlorentinD (commented):
Overall I like the notebook. Good structure to start with node-similarity and continue with pair-wise similarity and end with knn

Going through your notebook, i have some comments.

- Great to have both the local GDS setup and the AGA setup next to each other. But the remote projection query will only work for the AGA case. in the local case the projection looks slightly different. I would suggest to only keep the AGA setup on this PR.
- Instead of `my_session` lets call it `gds` for consistency with other examples
- The node label count query is written quite safe. I think we can drop the try+catch and also just display the counts_df without checking for empty. Reads a bit simpler
- I noticed the REMOTE_PROJECT_CYPHER projects feature0-4 as relationship properties, but never uses them? Using features on relationships is not supported in GDS. Most have algos give relationship properties a weight meaning. not sure if this applies here
- on your `Drop if exists`. you can make use of the `failIfMissing=False` parameter on graph.drop. this allows becoming a single line
- on `nodesimilarity` you can skip the stats mode if you call mutate anyway. the mutate mode does contain statistics in its result
- after mutate you can use `relationshipProperties.write` instead of computing the whole algo again. lets make use of the mutate result.
- how about moving the data viz before writing back? (maybe directly after the stream). makes it easier to reason if the result is good enough to write back)
- feature-based similarity: can you explain why you average the features per userId? I think it makes sense, but a comment would be good in the notebook
- you can drop the `WITH collect({id: userId, vec: vector}) as users` if you unwrap the user anyway afterwards.

on the 2.x syntax. we can take care of it

TreavVasu:
Thanks @FlorentinD 
For the thorough review 

I have started it will be updating once done

TreavVasu:
There's an issue with stanford edu in my region or globlally it appears, did most of the changes will update leftovers once the site is back up  

image

TreavVasu:
Hi @FlorentinD  

So far looks well from my end, tried to include everything as per review, lmk if something I missed. 

Looking forward to it

TreavVasu:
FYI: 

Below are the changes summary:

* **Local vs. AGA setup:** Removed 
* **Session naming (`my_session` → `gds`):** Changed. Updated session calls to `gds`.
* **Node count query simplification:**  The `if counts_df.empty:` check only remains now removed try catch.
* **Remove relationship features in projection:** Changed. Unused relationship properties removed from `REMOTE_PROJECT_CYPHER`.
* **Single-line graph drop (`failIfMissing=False`):** Changed. Simplified drops to single-line calls.
* **Skip stats mode before mutate:** Changed. Removed `stats` call and used `mutate_results` output directly.
* **Use `relationshipProperties.write`:** Changed. Used `relationshipProperties.write` on mutated results instead of re-running the algorithm.
* **Move visualization before write:** Changed. Placed plot directly after the `stream` step. Not on VIZ right now, would take a bit of time if required here, LMK 
* **Add explanation for averaging features:** Changed. Added comments explaining the 4D behavioral profile vector.
* **Simplify Cypher query (remove `collect`/unwrap):** Changed. Replaced `collect` with `max(CASE WHEN ...)` conditional aggregation.

brs96 (approved):

FlorentinD:
@TreavVasu thank you. I will merge this PR and open up a follow up PR with smaller changes to make it align with our CI.

TreavVasu:
Thanks @FlorentinD 

Appreciate the review and feedback for this PR. 

Hoping for few more next 🚀🚀

PR #1226, opened 2026-08-10
Add overwrite parameter to graph creation endpoints

FlorentinD:


src/graphdatascience/procedure_surface/api/catalog/catalog_endpoints.py
```diff
@@ -25,6 +25,20 @@
 from graphdatascience.procedure_surface.api.catalog.relationships_endpoints import RelationshipsEndpoints
 
 
+def validate_distinct_from_source(graph_name: str, source_graph: Graph) -> None:
```
DarthMax: nit: feels a bit off to me for this method to just live here public.

DarthMax: Generally this seems like a validation we should do in GDS not here.

src/graphdatascience/procedure_surface/cypher/catalog/utils.py
```diff
@@ -30,3 +32,26 @@ def require_database(query_runner: QueryRunner) -> str:
         )
 
     return database
+
+
+def drop_graph_if_exists(cypher_runner: QueryRunner, graph_name: str, username: str | None = None) -> None:
```
DarthMax: why is this implemented differently for Cypher?

DarthMax (approved):
Looks good, some smaller remarks

src/graphdatascience/procedure_surface/api/catalog/catalog_endpoints.py
FlorentinD: for filtering and sampling i didnt want to drop base graph if they tried to do the filter in-place.
i would see this check as needed if we want to introduce an overwrite on the client side

src/graphdatascience/procedure_surface/cypher/catalog/utils.py
FlorentinD: good point. will check if we can just use the cypher-based graphbackend as well

PR #1228, opened 2026-08-11
Expose embedding-api as preview

AlfredClemedtson:


src/graphdatascience/procedure_surface/api/gml/embed_endpoints.py
```diff
@@ -0,0 +1,95 @@
...
+from abc import ABC, abstractmethod
+
+from pandas import DataFrame
+
+from graphdatascience.graph.graph_api import Graph
+from graphdatascience.procedure_surface.api.gml.config import FastRPConfig, IdentityConfig
+from graphdatascience.procedure_surface.api.gml.results import EmbedMutateResult, EmbedWriteResult
+
+
+class GMLEmbedEntrypoints(ABC):
```
FlorentinD: can we type out gml? always confused what it stands for. or at least have this in the endpoint str

scripts/generate_endpoints.sh
```diff
@@ -0,0 +1,13 @@
...
+set -o nounset
+set -o pipefail
+
+cd "$(dirname "$0")"
+
+uv run gml_codegen/generate.py
+uvx ty check --fix ../src/graphdatascience/procedure_surface/api/gml
+uvx ty check --fix ../src/graphdatascience/procedure_surface/arrow/gml
+uvx ruff check --fix ../src/graphdatascience/procedure_surface/
+uv run ruff format ../src/graphdatascience/procedure_surface/api/gml ../src/graphdatascience/procedure_surface/arrow/gml
```
FlorentinD: nit, this could have been a just target, but i dont think its worth to keep

scripts/endpoint_codegen/code_generation.py
```diff
@@ -0,0 +1,468 @@
+import ast
```
FlorentinD, line 1: the code generation is quite sophisticated. as its only used by gml, i am not sure i want to keep it. also we dont regenerate these often.

in other places we went with asserting against the gds-api-spec for correctness

src/graphdatascience/procedure_surface/api/gml/embed_endpoints.py
```diff
@@ -0,0 +1,95 @@
...
+        node_ids: list[int] | None = None,
+        random_seed: int | None = None,
+        mutate_property: str,
+    ) -> EmbedMutateResult:
+        """
+        Parameters
+        ----------
+        G
+            The graph to run the procedure on.
+        graph_encoder
```
FlorentinD: how would you train an encoder upfront? is that also part of gml or just the usual train endpoint called upfront?

src/graphdatascience/procedure_surface/api/gml/embed_endpoints.py
```diff
@@ -0,0 +1,95 @@
...
+            Encoder used to produce node embeddings: either the name of a previously trained encoder model, or an inline configuration for a non-trainable encoder (e.g. FastRP or Identity).
+        node_ids
+            Node IDs to restrict the computation to. Defaults to all nodes in the graph if not specified.
+        random_seed
+            Seed for random number generation to ensure reproducible results.
+
+        Returns
+        -------
+        DataFrame
```
FlorentinD: can we specify it returns embeddings and also what the expected columns are?

src/graphdatascience/procedure_surface/api/gml/embed_endpoints.py
```diff
@@ -0,0 +1,95 @@
...
+        node_ids
+            Node IDs to restrict the computation to. Defaults to all nodes in the graph if not specified.
+        random_seed
+            Seed for random number generation to ensure reproducible results.
+        write_property
+            No description provided.
+
+        Returns
+        -------
+        EmbedWriteResult
```
FlorentinD: NIT: lets write out Embedding instead of Embed

src/graphdatascience/procedure_surface/api/gml/predict_endpoints.py
```diff
@@ -0,0 +1,103 @@
...
+        node_ids: list[int] | None = None,
+        random_seed: int | None = None,
+    ) -> DataFrame:
+        """
+        Parameters
+        ----------
+        G
+            The graph to run the procedure on.
+        graph_encoder
+            Name of the previously trained graph encoder model to use for prediction.
```
FlorentinD: is it helpful to call graph_encoder opposed to lets say `model`?

scripts/endpoint_codegen/embedding/train.py
```diff
@@ -0,0 +1,73 @@
...
+    graph_encoder: str | Annotated[NoTrainGraphEncoderConfig, Field(discriminator="graph_encoder_type")] = Field(
+        description="Encoder used to produce node embeddings: either the name of a previously trained encoder "
+        "model, or an inline configuration for a non-trainable encoder (e.g. FastRP or Identity)."
+    )
+    node_ids: list[int] | None = Field(default=None, description=NODE_IDS_DESCRIPTION)
+    random_seed: int = Field(default_factory=lambda: random.randint(0, 2**32 - 1), description=RANDOM_SEED_DESCRIPTION)
+
+
+class TrainConfig(BaseModel):
+    task_name: Literal["gml_train"] = Field("gml_train", validation_alias="taskName", description=TASK_NAME_DESCRIPTION)
```
FlorentinD, line 19: not quite sure i like the task concept to be shown to the user

scripts/endpoint_codegen/embedding/train.py
```diff
@@ -0,0 +1,73 @@
...
+    graph_name: str = Field(description=GRAPH_NAME_DESCRIPTION)
+    graph_encoder: Annotated[GraphEncoderConfig, Field(discriminator="graph_encoder_type")] = Field(
```
FlorentinD, line 20: whats the train part for fastrp?

scripts/endpoint_codegen/embedding/train.py
```diff
@@ -0,0 +1,73 @@
...
+        description="Configuration for the graph encoder to train (e.g. FastRP, GraphSAGE, or Identity)."
+    )
+    decoder: Annotated[DecoderConfig, Field(discriminator="decoder_type")] = Field(
+        description="Configuration for the decoder to train on top of the graph encoder's embeddings."
+    )
+    graph_encoder_save_name: str = Field(
+        description="Name under which the trained graph encoder model is saved."
+    )  # todo: make optional and generate random name
+    decoder_save_name: str = Field(
+        description="Name under which the trained decoder model is saved."
```
FlorentinD: why not just decoder_model_name?

scripts/endpoint_codegen/embedding/train.py
```diff
@@ -0,0 +1,73 @@
...
+    )  # todo: make optional and generate random name
+    train_data: list[tuple[int, int]] = Field(
+        description="Training examples, given as (node ID, class) pairs."
+    )  # todo: assert has consecutive classes or accept non-consecutive and non-numerical
+    val_data: list[tuple[int, int]] | None = Field(
```
FlorentinD: do we expect people to send these long list of tuples on this call?
sound like something modeled in the GDS graph via rel types?

scripts/endpoint_codegen/embedding/train.py
AlfredClemedtson: I don't think it is? Should be ignored in the endpoint signature.

src/graphdatascience/procedure_surface/api/gml/embed_endpoints.py
AlfredClemedtson: that's part of gml

scripts/endpoint_codegen/embedding/train.py
AlfredClemedtson: Same as for the existing pipelines. Trains only the downstream decoder (classifier), either mlp or gradient boosting, on the supervision data. For the mode for generating embeddings no training is needed for fastRP.

scripts/endpoint_codegen/embedding/train.py
AlfredClemedtson: The names are not set in stone, why it's convenient to change in only one place

src/graphdatascience/procedure_surface/arrow/embedding/encode_arrow_endpoints.py
```diff
@@ -0,0 +1,174 @@
...
+        config = self._build_config(
+            G,
+            graph_encoder=graph_encoder,
+            random_seed=random_seed,
+            job_id=job_id,
+            node_labels=node_labels,
+            relationship_types=relationship_types,
+            feature_properties=feature_properties,
+        )
+        return self._endpoints_helper.run_job(G, "v2/embedding.encode", config)
```
FlorentinD: we should have preview warnings here

src/graphdatascience/procedure_surface/api/node_embedding/config.py
```diff
@@ -0,0 +1,57 @@
...
+from annotated_types import Len
+from pydantic import BaseModel, Field, PositiveInt, model_validator
+
+NodeType = str
+
+Dropout = Annotated[float, Field(ge=0.0, lt=1.0)]
+
+
+class GraphSAGEConfig(BaseModel):
+    graph_encoder_type: Literal["graphsage"] = "graphsage"
```
FlorentinD, line 12: interesting that these are still lower-case

src/graphdatascience/procedure_surface/api/node_embedding/config.py
```diff
@@ -0,0 +1,57 @@
...
+    @model_validator(mode="after")
+    def depth_validator(self) -> "GraphSAGEConfig":
+        if self.hidden_dims is not None:
+            if self.num_neighbors is not None:
+                if len(self.hidden_dims) + 1 != len(self.num_neighbors):
+                    raise ValueError(
+                        "The depth of the GNN is inferred from the hidden dimensions per layer (+1) and the number of neighbors per layer. These do not align."
+                    )
+        return self
+
```
FlorentinD, line 28: usually we keep such validation on the session endpoint side and not check on the client

src/graphdatascience/procedure_surface/api/node_embedding/encode_endpoints.py
```diff
@@ -0,0 +1,226 @@
...
+        -------
+        EncodeWriteResult
+        """
+
+
+class EncodeConfig(BaseModel):
+    task_name: Literal["gml_encode"] = Field(
+        "gml_encode", validation_alias="taskName", description=TASK_NAME_DESCRIPTION
+    )
+    graph_encoder: str | Annotated[NoTrainGraphEncoderConfig, Field(discriminator="graph_encoder_type")] = Field(
```
FlorentinD, line 211: NoTrainGraphEncoderConfig reads a bit odd. how about rather stating what it is than what not?

src/graphdatascience/procedure_surface/api/node_embedding/encode_endpoints.py
```diff
@@ -0,0 +1,226 @@
...
+        *,
+        graph_encoder: str | (FastRPConfig | IdentityConfig),
+        random_seed: int | None = None,
+        mutate_property: str,
+        job_id: str | None = None,
+        node_labels: list[str] = ["*"],
+        relationship_types: list[str] = ["*"],
+        feature_properties: list[str],
+    ) -> EncodeMutateResults:
+        """
```
FlorentinD, line 105: nit - a small one sentence description would be helpful to understand this endpoint

src/graphdatascience/procedure_surface/api/descriptions.py
```diff
@@ -0,0 +1,3 @@
+TASK_NAME_DESCRIPTION = "Discriminator identifying which task the backend should run; not settable by the client."
```
FlorentinD, line 1: interesting idea.
maybe also useable inside the docstrings itself

src/graphdatascience/procedure_surface/api/job_config.py
```diff
@@ -0,0 +1,41 @@
...
+    node_labels: list[str] = Field(
+        default=["*"],
+        description="Filter the graph using the given node labels. Nodes with any of the given labels will be included.",
+    )
+    relationship_types: list[str] = Field(
+        default=["*"],
+        description="Filter the graph using the given relationship types. Relationships with any of the given types will be included.",
+    )
+    feature_properties: list[str] = Field(description="List of features to include in the job configuration.")
+
```
FlorentinD: not sure i like this base class.
it does not apply for to FastPath

src/graphdatascience/procedure_surface/arrow/embedding/predict_arrow_endpoints.py
```diff
@@ -0,0 +1,179 @@
...
+        node_labels: list[str] = ["*"],
+        relationship_types: list[str] = ["*"],
+        feature_properties: list[str],
+    ) -> dict[str, Any]:
+        extra_kwargs: dict[str, Any] = {}
+        if random_seed is not None:
+            extra_kwargs["random_seed"] = random_seed
+        config = PredictConfig(model_name=model_name, **extra_kwargs).model_dump(
+            exclude_none=True, exclude={"task_name"}
+        )
```
FlorentinD: as you exclude None, why is the handling for randomSeed needed?

tests/integration/procedure_surface/arrow/embedding/conftest.py
```diff
@@ -0,0 +1,54 @@
+from pathlib import Path
```
FlorentinD: why but it under `embeddings` instead of `node_embeddings`? this file is a pure duplicate of the `node_embeddings` one

tests/integration/procedure_surface/arrow/node_embedding/test_embeddings_predict_arrow_endpoints.py
```diff
@@ -0,0 +1,158 @@
...
+        yield g
+
+
+@pytest.fixture
+def predict_endpoints(arrow_client_runtime: AuthenticatedArrowClient) -> Generator[PredictArrowEndpoints, None, None]:
+    yield PredictArrowEndpoints(arrow_client_runtime)
+
+
+@pytest.fixture
+def model_name(arrow_client_runtime: AuthenticatedArrowClient, sample_graph: Graph) -> Generator[str, None, None]:
```
FlorentinD, line 69: i think ideally we would return the model_object here (result from train arrow endpoint) instead of the raw str

FlorentinD (commented):
found some smaller things which we can talk about

src/graphdatascience/procedure_surface/api/node_embedding/config.py
AlfredClemedtson: it's supposed to be the exact copy of the pydantic class from python runtime. is there a reason not to have it here? thought early validation would be useful due to server start up time

src/graphdatascience/procedure_surface/api/node_embedding/encode_endpoints.py
AlfredClemedtson: A graph encoder that does not need training 🤷. you have a suggestion?

src/graphdatascience/procedure_surface/api/node_embedding/config.py
FlorentinD: i think the question will be how they stay in sync in the future. we can keep this one for now

src/graphdatascience/procedure_surface/api/job_config.py
AlfredClemedtson: needed for me to be able to reuse python runtime pydantic models. Gets really messy if they start deviating.

AlfredClemedtson: could put it somewhere else if you prefer that

src/graphdatascience/procedure_surface/api/node_embedding/encode_endpoints.py
FlorentinD: just GraphEncoderConfig? or was this taken already? as this field is on the toplevel EncodeConfig, i dont see why we need the negation

src/graphdatascience/procedure_surface/arrow/embedding/predict_arrow_endpoints.py
AlfredClemedtson: I don't think the exclude_none should be there. I'll fix

tests/integration/procedure_surface/arrow/node_embedding/test_embeddings_predict_arrow_endpoints.py
AlfredClemedtson: there is no model object at this point. I think we need to think if we should have one and how it will behave.

tests/integration/procedure_surface/arrow/embedding/conftest.py
AlfredClemedtson: mimics the structure of `src`. There it is its own directory because I thought I was distinct enough from the existing `node_embeddings`

src/graphdatascience/procedure_surface/api/node_embedding/config.py
AlfredClemedtson: Copy pasting as we have to do with the rest of the class. Copying all is easier than specific parts i'd say.

src/graphdatascience/procedure_surface/api/node_embedding/encode_endpoints.py
AlfredClemedtson: GraphEncoderConfig is both trainable/learnable and non-trainable

src/graphdatascience/procedure_surface/api/job_config.py
FlorentinD: maybe we can clarify its not for every job but rather for EmbeddingJobs?

src/graphdatascience/procedure_surface/arrow/embedding/predict_arrow_endpoints.py
FlorentinD: if you have exclude_none, you could inline the random_seed no?

tests/integration/procedure_surface/arrow/embedding/conftest.py
FlorentinD: i would think we can inline with node_embeddings. for the user its node_embeddings right?

tests/integration/procedure_surface/arrow/node_embedding/test_embeddings_predict_arrow_endpoints.py
FlorentinD: true. okay i will let you card it and we can get this PR in after all its in preview

tests/integration/procedure_surface/arrow/node_embedding/test_embeddings_train_arrow_endpoints.py
```diff
@@ -0,0 +1,55 @@
...
+    result = train_endpoints(
+        G=sample_graph,
+        feature_properties=["x"],
+        graph_encoder=FastRPConfig(),
+        decoder=GBClassifierConfig(),
+        target_label="A",
+        target_property="y",
+        model_save_name="my_model456",
+    )
+
```
FlorentinD: this model is leaking. we need to drop it at the end of the test

PR #1246, opened 2026-08-25
Add per-call timeout and wait_for_available to Arrow Flight client

FlorentinD:
Fixes hangs in GDS release checks where the Arrow gRPC server accepts a
TCP connection but never responds, causing xdist workers to block
indefinitely on socket reads (20min pytest timeout → worker crash).

ref GDSA-1548

### DIFF
--- a/src/graphdatascience/arrow_client/authenticated_flight_client.py
+++ b/src/graphdatascience/arrow_client/authenticated_flight_client.py
@@ -13,6 +13,7 @@ from pyarrow import __version__ as arrow_version
 from pyarrow.flight import (
     Action,
     ActionType,
+    FlightCallOptions,
     FlightInternalError,
     FlightStreamReader,
     FlightTimedOutError,
@@ -58,7 +59,9 @@ class AuthenticatedArrowClient:
         encrypted
             A flag that indicates whether the connection should be encrypted (default is False)
         arrow_client_options
-            Additional options to be passed to the Arrow Flight client.
+            Additional options for the Arrow Flight client. The key ``call_timeout`` sets the
+            per-call RPC timeout in seconds (default 30s); all other keys are forwarded to
+            ``pyarrow.flight.FlightClient``.
         user_agent
             The user agent string to use for the connection. (default is `neo4j-graphdatascience-v[VERSION] pyarrow-v[PYARROW_VERSION]`)
         retry_config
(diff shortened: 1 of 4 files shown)


src/graphdatascience/session/aura_graph_data_science.py
```diff
@@ -916,6 +916,7 @@ def _verify_session_connectivity(self) -> None:
 
         :raises Exception: If the Aura Graph Analytics Session is unreachable
         """
+        self._authenticated_arrow_client.wait_for_available(timeout=30)
```
FlorentinD: this fails a notebook right now -- could increase it from 30s to something higher?

src/graphdatascience/arrow_client/authenticated_flight_client.py
```diff
@@ -87,17 +90,23 @@ def __init__(
                 wait_config=ExponentialWaitConfig(multiplier=1, min=1, max=10),
             )
 
+        # Extract our own call_timeout before forwarding the rest to FlightClient
```
DarthMax: Let's remove the chatty LLM explanations

src/graphdatascience/arrow_client/authenticated_flight_client.py
```diff
@@ -157,26 +166,45 @@ def auth_with_retry() -> None:
...
+    def do_action_with_retry(
+        self, endpoint: str, payload: bytes | dict[str, Any], options: FlightCallOptions | None = None
+    ) -> list[Result]:
         @self._retry_config.decorator(operation_name="Send action", logger=self._logger)
         def run_with_retry() -> list[Result]:
             try:
                 # the Flight response error code is only checked on iterator consumption
                 # we eagerly collect iterator here to trigger retry in case of an error
-                return list(self.do_action(endpoint, payload))
+                return list(self.do_action(endpoint, payload, options))
```
DarthMax: why do we ignoe self.call_options here?

src/graphdatascience/arrow_client/authenticated_flight_client.py
```diff
@@ -218,6 +246,9 @@ def close(self) -> None:
         if self._flight_client:
             self._flight_client.close()
 
+    def _build_call_options(self) -> FlightCallOptions | None:
```
DarthMax, line 240: call_timeout is not used anywhere else, is it? since this is called in the constructor, we can pass it in here and skip the self._call_timeout

src/graphdatascience/arrow_client/authenticated_flight_client.py
```diff
@@ -157,26 +166,45 @@ def auth_with_retry() -> None:
     def get_stream(self, ticket: Ticket) -> FlightStreamReader:
         return self._flight_client.do_get(ticket)
 
-    def do_action(self, endpoint: str, payload: bytes | dict[str, Any]) -> Iterator[Result]:
+    def do_action(
```
DarthMax: Why do we need to ability to specify override options? It is not used anywhere

DarthMax (changes requested):
Call timeout looks good but there are a couple of other changes that can be improved

src/graphdatascience/arrow_client/authenticated_flight_client.py
FlorentinD: it is used when reconstructing the client from  __setstate__  without self._call_timeout we would lose the timeout on a pickle roundtrip

src/graphdatascience/arrow_client/authenticated_flight_client.py
FlorentinD: fair point. i thought of allowing users to specify. but they cant override the timeout on init now. so having it here as well seems redundant to start with

src/graphdatascience/arrow_client/authenticated_flight_client.py
FlorentinD: would have been considered inside the self.do_action

DarthMax (approved):

PR #1251, opened 2026-08-27
Support HITS for AGA

DarthMax:


src/graphdatascience/procedure_surface/api/centrality/hits_endpoints.py
FlorentinD: this indicates a prefix, but in the arrow impl you applied it as a suffix?

FlorentinD (changes requested):
tests are failing due to comment inconsistency. in this case likely good to add exceptions in the expected parameter docstring resource.

mostly confused about prefix vs suffix on the mutate/write property part

src/graphdatascience/procedure_surface/api/centrality/hits_endpoints.py
DarthMax: Yeah this is a bit confusing. However this is how it is handled in the plugin
```
    private String property(String property){
        return property.concat(mutateProperty);
    }
```

FlorentinD: so you need to fix the docs?

DarthMax: Yes, fixed it for now only in the python docs

FlorentinD (approved):

FlorentinD:
fails on test_hits_stats now

PR #1252, opened 2026-08-27
Retryable write-job and start-job

FlorentinD:
- **Retry start of remote write**
- **Document why project cannot be retried**
- **Fix retry logic for start job**

### DIFF
--- a/src/graphdatascience/arrow_client/v2/job_client.py
+++ b/src/graphdatascience/arrow_client/v2/job_client.py
@@ -1,16 +1,22 @@
 import json
+import logging
+import uuid
 from typing import Any
 
 from pandas import ArrowDtype, DataFrame
+from pyarrow import ArrowKeyError
 from pyarrow.flight import Ticket
-from tenacity import Retrying, retry_if_result
+from tenacity import Retrying, retry_if_exception, retry_if_result, stop_after_attempt, wait_fixed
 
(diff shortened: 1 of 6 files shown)


src/graphdatascience/arrow_client/v2/job_client.py
```diff
@@ -31,10 +37,30 @@ def run_job_and_wait(
...
+                    res = client.do_action_with_retry(endpoint, config)
+                    return JobIdConfig(**deserialize_single(res)).job_id
+                except FLIGHT_TRANSIENT_EXCEPTIONS as e:
+                    try:
+                        JobClient.get_job_status(client, job_id)
+                        logger.debug(f"Job '{job_id}' already started, recovering job_id.")
+                        return job_id
+                    except ArrowKeyError:
+                        logger.debug(f"Job '{job_id}' not found on server, will retry run_job.")
+                        raise e
```
DarthMax, line 61: maybe move this into a function and reuse it for the write path

DarthMax (approved):

src/graphdatascience/arrow_client/v2/job_client.py
FlorentinD: on the write-path we interact via bolt and here with arrow.
the error handling does differ a bit. so i opt to not try to share a common function

PR #1264, opened 2026-09-09
Support dropping multiple graphs

DarthMax:


src/graphdatascience/procedure_surface/arrow/catalog/catalog_arrow_endpoints.py
```diff
@@ -110,24 +111,32 @@ def construct(
         constructor.run(nodes, relationships)
         return get_graph(graph_name, self._arrow_client)
 
-    def drop(self, G: Graph | str, fail_if_missing: bool = True) -> GraphInfo | None:
-        """Drop a graph from the graph catalog.
+    def drop(
+        self, G: Graph | str | list[Graph | str], fail_if_missing: bool = True
+    ) -> GraphInfo | list[GraphInfo] | None:
```
FlorentinD: that could be a bit annoying for the user as they need to check the output type for mypy not to complain

FlorentinD (commented):
change itself is straight forward. but wonder about the user implications on typing.

but i cant come up with a better way immediately

src/graphdatascience/procedure_surface/arrow/catalog/catalog_arrow_endpoints.py
DarthMax: Yes it is not ideal but would otherwise be an even more breaking change. We can also not do this change and keep it as is

FlorentinD: IMO exposing this option is more harmful for the API than helpful.

soerenreichardt: > Yes it is not ideal but would otherwise be an even more breaking change

What is the more breakling change alternative? Is the 2.0 release already finalized, because otherwise that would be the perfect timing for breaking changes

FlorentinD: i would agree 2.0 release is the best point in time to do this change.

i see the advantage for the cypher-api but with python i think a simple for loop is easy enough here.

DarthMax: :+1:  made the breaking change complete. I had forgotten that 2.0 was not yet released :see_no_evil:

PR #1265, opened 2026-09-09
Expose topological LinkPrediction

FlorentinD:
closes GDSA-1515


src/graphdatascience/session/aura_graph_data_science.py
```diff
@@ -252,6 +254,7 @@ def __init__(
             self._write_protocol = WriteProtocol.select(authenticated_arrow_client, db_query_runner)
         self._session_lifecycle_manager = session_lifecycle_manager
         self._show_progress = show_progress
+        self._db_in_aura: bool | None = None
```
soerenreichardt: Why do we need this bool flag in addition to the `DbEnvironmentResolver`?

soerenreichardt (approved):

src/graphdatascience/session/aura_graph_data_science.py
FlorentinD: otherwise we would resolve the database enviroment on every call to an AuraDB-only endpoint.
this is essentially a cache.
makes me wonder if we couldnt use an annotation in the method for caching

soerenreichardt: You are caching the value anyways by having this line:
```
if self._db_in_aura is None:
            self._db_in_aura = DbEnvironmentResolver.hosted_in_aura(self._db_query_runner)     
```
So I still think it is not necessary here

FlorentinD: yes thats setting the field on the aura_graph_data_science object.
your comment is refering to the init of the same field no?

FlorentinD: i reworked the logic a bit to avoid this boolean flag on the AuraGDS object but rather move it closer to the Neo4jQueryRunner

PR #1271, opened 2026-09-16
Fix expected projection endpoint names

FlorentinD:


src/graphdatascience/procedure_surface/arrow/jobs_arrow_endpoints.py
```diff
@@ -100,6 +100,7 @@ def _is_projection(self, job_name: str, graph_name: str) -> bool:
             "v2/graph.sample.rwr",
             "v2/graph.sample.cnarw",
             "v2/graph.generate",
+            "v2/graph.create.fromTriplets",
```
FlorentinD, line 103: @DarthMax i saw you changed this on the server side.
was this by accident or do you prefer `"v2/graph.create.fromTriplets"` instead of `"v2/graph.project.fromTriplets"`

FlorentinD: also should be `v2/graph.create.fromTables` now

FlorentinD: will be resolved on server side

PR #1275, opened 2026-09-17
Scale default Arrow retry budget with the per-call timeout

FlorentinD:
before we could end up with only a single attempt as the 10s stop_after_delay was already reached after the first timeout.


DarthMax (approved):
Looks good, I think it is a bit strange that we have to do all these calculations ourselves.

FlorentinD:
i agree and we might want to consider allowing a better retry config api we expose to the end user. 
alternatively we could drop the stop_after_delay

PR #1280, opened 2026-09-18
Update GHA workflows to include 2.0

nvitucci:
@recrwplay Asking for a review to double-check this is still the way we want to do this after 2.0 has been released. The intention is to keep publishing the content on `main` (2.1 now) as a preview.


neo4j-docops-agent:
Looks like you've updated the API reference!

Check out your changes at https://neo4j-graph-data-science-client-1280-api.surge.sh

If this PR also previews the manual, its API reference links point here.

neo4j-docops-agent:
This PR includes documentation updates
View the updated docs at https://neo4j-graph-data-science-client-1280.surge.sh

PR #1282, opened 2026-09-18
Derive default session project across all organizations

RafalSkolasinski:
When no `project_id` is given, the client now first checks the account's organizations via the Aura API v2beta1 `organizations` endpoint. Accounts with access to multiple organizations must specify `project_id` explicitly. Single-organization accounts keep the previous derivation based on `/v1/tenants`.

Previously `/v1/tenants` alone was used, which is not organization-aware and could make accounts with access to multiple organizations appear to have a single project, silently creating sessions in the wrong one.


src/graphdatascience/session/aura_api.py
```diff
@@ -434,6 +435,15 @@ def estimate_size(
...
     def _get_project_id(self) -> str:
+        # `v1/tenants` is not organization-aware and can silently default to a single
+        # project, so multi-organization accounts are rejected first.
+        organizations = self._list_organizations()
+
+        if len(organizations) > 1:
+            raise RuntimeError(
+                "This account has access to multiple organizations. Please specify the `project_id` to use."
+            )
+
```
FlorentinD, line 446: can we have a follow up ticket somewhere to allow users to specify the organization inside the AuraAPI credentials data class?

this would already help here and giving the option sooner than later makes sense to me.

src/graphdatascience/session/aura_api_responses.py
```diff
@@ -247,6 +247,19 @@ def from_connection_url(cls, connection_url: str) -> WaitResult:
...
+class OrganizationDetails:
+    id: str
+    name: str
+
+    @classmethod
+    def from_json(cls, json: dict[str, Any]) -> OrganizationDetails:
+        return cls(
+            id=json["id"],
+            name=json["name"],
+        )
```
FlorentinD, line 260: matches style in other responses.
at some point we might want to switch to pydantic classes

changelog.md
```diff
@@ -8,6 +8,7 @@
 
 ## Bug fixes
 
+* Deriving the default project for GDS Sessions now first checks the account's organizations via the Aura API `v2beta1` endpoint. Accounts with access to multiple organizations must specify `project_id` explicitly, since the non-organization-aware `v1/tenants` endpoint could make them appear to have a single project and silently create sessions in the wrong one.
```
FlorentinD:
```suggestion
* Deriving the default project for GDS Sessions now first checks the users organizations. Users with access to multiple organizations must specify `project_id` explicitly to have a single project and avoid silently create sessions in the wrong one.
```

FlorentinD (approved):
good fix.

some smaller comments and follow up ideas, but nothing blocking

PR #1283, opened 2026-09-22
Infer arrow tls options from bolt uri scheme

DarthMax:


src/graphdatascience/graph_data_science.py
```diff
@@ -905,6 +908,23 @@ def _derive_aura_ds(endpoint: str, auth: neo4j.Auth | None, database: str | None
...
+    @staticmethod
+    def _derive_arrow_client_options(
+        endpoint: str | Driver | QueryRunner, arrow_client_options: dict[str, Any] | None
+    ) -> dict[str, Any] | None:
+        if not isinstance(endpoint, str):
+            return arrow_client_options
+
+        scheme = endpoint.split("://")[0].lower()
+        if not scheme.endswith("+ssc"):
+            return arrow_client_options
```
FlorentinD, line 920: if you like you can also use 
```
from urllib.parse import urlsplit
scheme = urlsplit("bolt+scc://foo.bar:7474").scheme
if scheme and not scheme.endswith("+ssc")"
```

this will allow you to access `scheme` on the result directly

src/graphdatascience/graph_data_science.py
```diff
@@ -905,6 +908,23 @@ def _derive_aura_ds(endpoint: str, auth: neo4j.Auth | None, database: str | None
...
+
+        options = dict(arrow_client_options) if arrow_client_options else {}
+        if "disable_server_verification" in options:
+            return options
+
```
FlorentinD, line 925: nit you could move this check up before splitting

changelog.md
```diff
@@ -5,6 +5,7 @@
 
 ## New features
 
+* `GraphDataScience` now disables Arrow server certificate verification when the Bolt URI uses a `+ssc` scheme (`bolt+ssc`, `neo4j+ssc`), unless `disable_server_verification` is set explicitly in `arrow_client_options`. This restores the 1.x behaviour for self-signed certificates, where `arrow_disable_server_verification` defaulted to `True`.
```
FlorentinD, line 8: did we really default to True in 1.x .. thats seems like a crazy default to me

FlorentinD (approved):
making the users live easier 👍 

have some smaller improvement ideas

src/graphdatascience/graph_data_science.py
DarthMax: I suggested that as well, but I think the driver should actually parse that URL and throw the specific error.

changelog.md
DarthMax: Yip https://github.com/neo4j/graph-data-science-client/blob/7b1d2c47234aa53e50b5574d18d8a8f1ab960b06/src/graphdatascience/graph_data_science.py#L45 bad bug oO

PR #1284, opened 2026-09-22
Allow primitive types in API instead of forcing enums

FlorentinD:
- **Allow plain string aliases for enum parameters**
- **Simplify str aliases: drop AggregationLike/DirectionLike, docstring cleanup**


scripts/ci/aura_api_ci.py
```diff
@@ -104,10 +104,17 @@ def create_instance(self, name: str, memory: str, type: str) -> dict[str, Any]:
             if should_retry:
                 logging.debug(f"Error code: {response.status_code} - Retrying in {wait_time} s")
 
-        response_json = response.json()
+        try:
+            response_json = response.json()
+        except requests.JSONDecodeError:
+            # e.g. an HTML error page from a gateway 502 -- fail with a readable error
```
DarthMax, line 110: niot remove

src/graphdatascience/procedure_surface/api/catalog/relationships_endpoints.py
```diff
@@ -170,7 +170,7 @@ def to_undirected(
         relationship_type: str,
         mutate_relationship_type: str,
         *,
-        aggregation: Aggregation | dict[str, Aggregation] | None = None,
+        aggregation: Aggregation | str | dict[str, Aggregation | str] | None = None,
```
DarthMax, line 173: I honestly do not love this change

DarthMax (commented):
I have to be very honest, this change goes against all my developer instincts. I think we should discuss this a bit.

vtaskow (approved):
Looks good

FlorentinD:
> I have to be very honest, this change goes against all my developer instincts. I think we should discuss this a bit.

lets do so tomorrow at adhoc. i do see your point. IMO its more pythonic to allow the primitive types.
(hearing Adam in my head to allow for dict[str, Any]) but it also see how it goes a bit against other changes we made

as an example on the python driver take https://neo4j.com/docs/api/python-driver/current/api.html#neo4j.RoutingControl. 
maybe actually using StrEnum might do the trick once we opt to not support python 3.10

FlorentinD:
a point for this change are other places in our api such as https://neo4j.com/docs/graph-data-science-client/2.0/api/sessions/gds_sessions/#graphdatascience.session.gds_sessions.GdsSessions.get_or_create where we also allow str or SessionMemory in the signature

PR #1288, opened 2026-09-23
Add a hint about DNS failures w.r.t. VPNs

DarthMax:
* This will happen to many Neo4j employees on Windows + Cloudflare WARP client


src/graphdatascience/arrow_client/authenticated_flight_client.py
```diff
@@ -237,11 +238,42 @@ def _diagnose_connection_failure(self, operation: Callable[[], T]) -> T:
...
-        except FLIGHT_TRANSIENT_EXCEPTIONS:
+        except FLIGHT_TRANSIENT_EXCEPTIONS as e:
             if self._health_check:
                 self._health_check.raise_if_unhealthy()
+            self._raise_if_grpc_dns_resolver_failed(e)
             raise
 
+    def _raise_if_grpc_dns_resolver_failed(self, error: Exception) -> None:
+        """
+        gRPC ships its own DNS resolver (c-ares) that talks directly to the configured nameserver.
```
FlorentinD:
```suggestion
        gRPC uses the c-ares DNS resolver that talks directly to the configured nameserver.
```
a bit better wording that gRPC did not develop this

src/graphdatascience/arrow_client/authenticated_flight_client.py
```diff
@@ -237,11 +238,42 @@ def _diagnose_connection_failure(self, operation: Callable[[], T]) -> T:
...
         try:
             return operation()
-        except FLIGHT_TRANSIENT_EXCEPTIONS:
+        except FLIGHT_TRANSIENT_EXCEPTIONS as e:
             if self._health_check:
                 self._health_check.raise_if_unhealthy()
+            self._raise_if_grpc_dns_resolver_failed(e)
             raise
 
+    def _raise_if_grpc_dns_resolver_failed(self, error: Exception) -> None:
```
FlorentinD, line 247: i would consider to move this error rewriting into its own file inside the arrow-client folder

src/graphdatascience/arrow_client/authenticated_flight_client.py
```diff
@@ -237,11 +238,42 @@ def _diagnose_connection_failure(self, operation: Callable[[], T]) -> T:
...
+        while the OS resolver works fine. If gRPC failed to resolve the host but the OS resolver can,
+        point the user at the `GRPC_DNS_RESOLVER=native` workaround.
+        """
+        if not isinstance(error, FlightUnavailableError):
+            return
+        message = str(error)
+        if "address lookup failed" not in message and "errors resolving" not in message:
+            return
+        if not self._os_resolver_can_resolve_host():
+            return
```
FlorentinD, line 260: NIT: might just want to return error the others here as well. that makes it possible to just raise this one in the usage site instead of calling this one or raise on its own.

FlorentinD (approved):
makes sense to have it for now.

only some code style comments

src/graphdatascience/arrow_client/authenticated_flight_client.py
DarthMax: but this is very specific to this client. Wouldn't tat only make sense if we wanted to reuse it?

src/graphdatascience/arrow_client/authenticated_flight_client.py
DarthMax: I think the current flow is cleaner from the call side perspective. It makes it clear that we will definitely raise an error.

src/graphdatascience/arrow_client/authenticated_flight_client.py
FlorentinD: that would be another reason. for me it would be also to keep this class a bit slimmer and use "composition"

but fine with keeping as is

